Issue #7, opened 2018-09-19
[feature-request] provide customizable storageclass


mmisztal1980:
It would be pretty awesome if the chart would support providing a storage class name via the values.yaml file. Is there a chance of having that implemented?

mitchellh:
Yep definitely, feel free to open a PR but we'll also try to get to it soon!

mmisztal1980:
What is the default value, you guys would prefer?

mitchellh:
I would just set it only if the value is set, so we retain the cluster default for now.

mmisztal1980:
I’ll open up a PR in a few hrs

mitchellh:
Closed by #7. Thanks!

PR #8, opened 2018-09-20
#7 added an optional storageClass to the server-statefuset

mmisztal1980:
This PR adds an optional storageClass support to the HELM template as specified in #7

When defined, the provided `server.storageClass` shall be rendered into the `server-statefulset.yaml`.

@mitchellh Is that satisfactory? Please note that I'm not a GO expert, so if there's anything wrong, please let me know.

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -160,4 +160,7 @@ spec:
         resources:
           requests:
             storage: {{ .Values.server.storage }}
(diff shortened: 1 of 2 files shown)


mitchellh:
Yep this looks pretty good. I haven't merged yet because I'm going to add some unit tests to this and I want to pair it with docs updates (which are unfortunately in another repo, Consul).

Issue #13, opened 2018-09-29
Unable to run v0.2.0 in minikube. "pod has unbound PersistentVolumeClaims"


kohv:
Running `helm install --name consul --namespace=pkr -f dev-consul.yaml ./consul-helm` with these custom values:
```
syncCatalog:
  enabled: true
server:
  storage: "8Gi"
```
It completes successfully, but when running `kubectl get pods -n pkr`, I see that `consul-server-1` & `consul-server-2` are Pending.
```
NAME                                   READY     STATUS    RESTARTS   AGE
consul-4vstb                           0/1       Running   0          15m
consul-server-0                        0/1       Running   0          15m
consul-server-1                        0/1       Pending   0          15m
consul-server-2                        0/1       Pending   0          15m
consul-sync-catalog-587b6859f6-dpj5v   1/1       Running   0          15m
```
Closer inspection on `pods consul-server-0`:
```  
kubectl describe pods consul-server-0 -n pkr
Name:           consul-server-0
Namespace:      pkr
Node:           minikube/10.0.2.15
Start Time:     Sat, 29 Sep 2018 22:39:09 +0300
Labels:         app=consul
                chart=consul-0.1.0
                component=server
                controller-revision-hash=consul-server-66479c5df5
                hasDNS=true
                release=consul
                statefulset.kubernetes.io/pod-name=consul-server-0
Annotations:    consul.hashicorp.com/connect-inject=false
Status:         Running
IP:             172.17.0.9
Controlled By:  StatefulSet/consul-server
Containers:
  consul:
    Container ID:  docker://7de44c6027bdb78ba4b7bc73643701aa9e0bbb55abce8ce2c7b8e12e2adf82b0
    Image:         consul:1.2.3
    Image ID:      docker-pullable://consul@sha256:ea66d17d8c8c1f1afb2138528d62a917093fcd2e3b3a7b216a52c253189ea980
    Ports:         8500/TCP, 8301/TCP, 8302/TCP, 8300/TCP, 8600/TCP, 8600/UDP
    Host Ports:    0/TCP, 0/TCP, 0/TCP, 0/TCP, 0/TCP, 0/UDP
    Command:
      /bin/sh
      -ec
      CONSUL_FULLNAME="consul"

exec /bin/consul agent \
  -advertise="${POD_IP}" \
  -bind=0.0.0.0 \
  -bootstrap-expect=3 \
  -client=0.0.0.0 \
  -config-dir=/consul/config \
[...]

PR #18, opened 2018-10-04
Configure affinity to be a variable to allow overriding

s3than:
AS per issue https://github.com/hashicorp/consul-helm/issues/13

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -37,15 +37,10 @@ spec:
       annotations:
         "consul.hashicorp.com/connect-inject": "false"
     spec:
+    {{- if .Values.server.affinity }}
       affinity:
-        podAntiAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
-            - labelSelector:
-                matchLabels:
-                  app: {{ template "consul.name" . }}
-                  release: "{{ .Release.Name }}"
-                  component: server
-              topologyKey: kubernetes.io/hostname
+{{ tpl .Values.server.affinity . | indent 8 }}
+{{- end }}
       terminationGracePeriodSeconds: 10
       securityContext:
         fsGroup: 1000
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -39,7 +39,7 @@ spec:
     spec:
     {{- if .Values.server.affinity }}
       affinity:
-      {{ tpl .Values.server.affinity . | indent 8 }}
+        {{ tpl .Values.server.affinity . }}
     {{- end }}
       terminationGracePeriodSeconds: 10
       securityContext:
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -39,8 +39,8 @@ spec:
     spec:
     {{- if .Values.server.affinity }}
       affinity:
-{{ tpl .Values.server.affinity . | indent 8 }}
-{{- end }}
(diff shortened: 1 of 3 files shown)


values.yaml
```diff
@@ -79,6 +79,20 @@ server:
     #   name: my-secret
     #   load: false # if true, will add to `-config-dir` to load by Consul
 
+  # Affinity Settings
+  # Overriding this will allow deployment to single node services such as
```
adilyse: What does an override look like? Can a user set the value to `null` or would they need to comment this option out? It would be good to add these details into the comment.

s3than:
Thanks I'll get on this in the next couple of days.

templates/server-statefulset.yaml
```diff
@@ -37,15 +37,10 @@ spec:
...
       affinity:
-        podAntiAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
-            - labelSelector:
-                matchLabels:
-                  app: {{ template "consul.name" . }}
-                  release: "{{ .Release.Name }}"
-                  component: server
-              topologyKey: kubernetes.io/hostname
+{{ tpl .Values.server.affinity . | indent 8 }}
```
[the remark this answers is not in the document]
s3than: Fixed, templating and removed indent values.

values.yaml
s3than: Updated comment and provide details on the override.

s3than:
@adilyse 
Updated details and added tests.

test/unit/server-statefulset.bats
```diff
@@ -11,6 +11,27 @@ load _helpers
...
 }
 
+
+@test "server/StatefulSet: affinity not set with server.affinity" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/server-statefulset.yaml  \
+      --set 'server.affinity=' \
+      . | tee /dev/stderr |
+      yq 'length > 0' | tee /dev/stderr)
```
adilyse: The `yq 'length > 0'` just checks that there _is_ output when we run `helm template` on the `server-statefulset.yaml`, so it doesn't give us much information about the `affinity` variable that we're trying to check.

In the case where `affinity=null`, we want to make sure that the `affinity:` section doesn't show up in the yaml file. Try this instead:
```suggestion
      yq '.spec.template.spec | any(contains("affinity"))' | tee /dev/stderr)
```

The condition needs to be updated to `false` with this change (I've created another suggestion for that).

test/unit/server-statefulset.bats
```diff
@@ -11,6 +11,27 @@ load _helpers
...
+  [ "${actual}" = "true" ]
```
adilyse: To match the changed condition above:
```suggestion
  [ "${actual}" = "false" ]
```

test/unit/server-statefulset.bats
```diff
@@ -11,6 +11,27 @@ load _helpers
...
+}
+
+@test "server/StatefulSet: affinity set with server.affinity as empty" {
```
adilyse: Since the other test checks what happens when `affinity` is null, it would be good to check the default case instead here. I'm happy to provide specific suggestions if you'd like.

s3than:
@adilyse Thanks for the feedback, I'll go over #55 and your comments and update.

test/unit/server-statefulset.bats
adilyse: Okay, glad you found something that worked!

PR #24, opened 2018-10-06
Adding StorageClass object

smugcloud:
When running the example on a stock Amazon EKS cluster, the server pods do not come up because the PersistentVolumeClaims indicate `no persistent volumes available for this claim and no storage class is set`.

I believe a StorageClass is needed for the [dynamic volume provisioner](https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/#enabling-dynamic-provisioning) to function properly in the Cloud Providers, by creating the PersistentVolumes (related to .  This change offers the option to create a StorageClass, and enable it as the default provisioner.  I commented out the AWS values, but provided examples to reference.

This is related to #8.

### DIFF
--- a/templates/server-storageclass.yaml
+++ b/templates/server-storageclass.yaml
@@ -1,0 +1,13 @@
+{{- if eq .Values.server.provisionerEnabled "-"}}
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: {{ .Values.server.storageClass }}
+  annotations:
+    "storageclass.kubernetes.io/is-default-class": "true"
+provisioner: {{ .Values.server.provisioner }}
+parameters:
+  type: {{ .Values.server.type }}
+reclaimPolicy: Delete
+volumeBindingMode: Immediate
+{{- end }}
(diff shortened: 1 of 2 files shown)


adilyse:
@smugcloud Thank you for putting in time on this issue. Unfortunately, there are several issues with this as a solution:

The creation and management of a Kubernetes cluster's StorageClasses is an operational concern specific to each cluster that is best left to cluster operators. We would be making assumptions about what the appropriate storage is and applying it to every situation. It would come as a big surprise for a cluster operator for us to silently switch the default StorageClass and might cause hard-to-find problems in clusters where StorageClasses are already defined. 

Additionally, this doesn't seem to be as broad a problem as it might at first appear. Checking this with both GKE and AKS, both of these managed solutions define a default StorageClass already, making this unnecessary. 

That said, your point about this being an unexpected error for folks starting from scratch in a cluster without a provided default StorageClass is a good one. To address that, I've added information into the documentation to clarify this [here](https://www.consul.io/docs/platform/k8s/helm.html#v-server-storageclass).

PR #29, opened 2018-10-09
feat: add rbac support

### LINKED ISSUE #20: Add support for RBAC for consul-k8s in chart
Currently when running the Helm chart against a Kubernetes cluster that has RBAC fully enabled, no RBAC policies exist for consul-k8s which prevents the functionality from being able to communicate with the Kubernetes API.

The required policy to enable consul-k8s to function also does not exist anywhere in the documentation.

Note: this will also be needed when the connect auto-injection is complete as well

tomwganem:
closes #20

### DIFF
--- a/templates/sync-cluster-role.yaml
+++ b/templates/sync-cluster-role.yaml
@@ -1,0 +1,25 @@
+{{- $rbacEnabled := (or (and (ne (.Values.rbac.enabled | toString) "-") .Values.rbac.enabled) (and (eq (.Values.rbac.enabled | toString) "-") .Values.rbac.enabled)) }}
+{{- $syncEnabled := (or (and (ne (.Values.syncCatalog.enabled | toString) "-") .Values.syncCatalog.enabled) (and (eq (.Values.syncCatalog.enabled | toString) "-") .Values.global.enabled)) }}
+{{- if (and $rbacEnabled $syncEnabled) }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: consul:sync
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+rules:
+  - apiGroups: [""]
+    resources:
+      - services
+      - endpoints
+    verbs:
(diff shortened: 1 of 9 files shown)


values.yaml
```diff
@@ -186,3 +186,8 @@ connectInject:
     # defaults but can be customized if necessary.
     certName: tls.crt
     keyName: tls.key
+
+# Enabling rbac will create cluster roles and cluster role bindings to allow the
+# syncCatalog service to communicate with the kubernetes api to get/create services
+rbac:
+  enabled: false
```
mitchellh: I think this should probably go under the `syncCatalog` section. I can see the easiest way would be:

```yaml
syncCatalog:
  rbac:
    enabled: false/true
```

tomwganem:
I've been deploying this helm chart to a dedicated namespace, hence why I didn't think to add a service account. I agree with that approach however, and will integrate it with this PR.

tomwganem:
PR updated to include the creation of a service account

jipperinbham:
I tested this out and the use of `:` in the API object names will not work. Once I changed it to `-`, the helm install worked as expected.

templates/sync-cluster-role.yaml
```diff
@@ -0,0 +1,25 @@
...
+    resources:
+      - services
+      - endpoints
+    verbs:
+      - get
+      - list
+      - watch
+      - update
+      - patch
+      - delete
```
jipperinbham:
```suggestion
      - delete
      - create
```

I was getting ` error="services is forbidden: User "system:serviceaccount:platform:consul-sync-catalog" cannot create services in the namespace "platform""` without this. Once added, the `consul` service was synced to k8s.

mitchellh: Nice. Yep.

It seems possible that we should create two roles here. One for k8s => Consul and one for Consul => k8s, since either one-way can be disabled/enabled and k8s => Consul only requires read-only privileges which is a nice to have.

mitchellh: Actually decided to just add `-create` for now and keep the simple thing, we can split it out later if its useful.

mitchellh:
Thank you very much @tomwganem! I just ran it and it worked great. I added the change @jipperinbham recommended as well. Merged!

PR #30, opened 2018-10-09
Provide a mechanism to run a Prometheus monitoring sidecar

pcfens:
Since Prometheus and Kubernetes go together, and there's an official Consul exporter maintained by the authors of Prometheus, I thought some local changes for testing might be worth sharing.

Things I'm still uncertain about:
- Is `server.prometheusSidecar` the right place to put the feature flag (and should it be called that)?
- Should the service always be created if the sidecar is enabled? (we use Prometheus operator, so services are a must)
- Assuming the disabled by default is correct, should a sidecar be an exception to the `global.enabled` flag?

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -151,6 +151,15 @@ spec:
             periodSeconds: 3
             successThreshold: 1
             timeoutSeconds: 5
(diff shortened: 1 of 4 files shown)


mitchellh:
Hey thanks! I'm not sure if we're ready _right now_ to fully support an official-ish Prometheus sidecar like this. Instead, what I would be open to, is another config to add _any_ extra container to the server statefulset. Then you can run whatever sidecar(s) you want. 

That would let you run Prometheus now, and now preclude us from just adding a `prometheus: true` type value later.

How does that sound?

If that's okay, going to close this. I appreciate it though!

PR #33, opened 2018-10-11
Add nodeSelector and tolerations support

### LINKED ISSUE #19: Add possiblity to add tolerations to client and server agents
We are running the consul chart in a cluster that taints the controller nodes. We would like to run the consul client on the controller nodes as well. In order to do this we need to be able to specify tolerations to those taints.

https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/

joakimhellum-in:
Fixes #19

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -142,4 +142,8 @@ spec:
                   grep -E '".+"'
           resources:
 {{ toYaml .Values.client.resources | indent 12 }}
+    {{- if .Values.client.tolerations }}
+      tolerations:
+{{ toYaml .Values.client.tolerations | indent 8 }}
+    {{- end }}
 {{- end }}
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -142,6 +142,10 @@ spec:
                   grep -E '".+"'
           resources:
 {{ toYaml .Values.client.resources | indent 12 }}
+    {{- if .Values.client.nodeSelector }}
+      nodeSelector:
+{{ toYaml .Values.client.nodeSelector | indent 8 }}
+    {{- end }}
     {{- if .Values.client.tolerations }}
       tolerations:
 {{ toYaml .Values.client.tolerations | indent 8 }}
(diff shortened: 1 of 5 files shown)


sc7565:
hi @adilyse and @anubhavmishra, can you guys please take a look at this PR and help in merging it

adilyse:
@joakimhellum-in Could you tell me a bit more about your choice of `nodeSelector` support over `affinity`? Is there a reason that `affinity` doesn't work for your use case?

ttyS0:
FWIW, I use `nodeSelector` instead of `affinity` because I have a mixed architecture home lab, and I like to be able to direct `consul-helm` to deploy to `beta.kubernetes.io/arch: arm64` for example.

labria:
Why was this never merged?

joakimhellum-in:
@adilyse I'm very sorry for the late reply. Honestly after asking around I don't have any good justification for this, but i'm not an expert and didn't have time to look too much into this. Here is my justification:
- Some documentation still recommends to use `nodeSelector`, for example we have both Windows and Linux nodes in some cluster, and the [Kubernetes Windows documentation](https://kubernetes.io/docs/getting-started-guides/windows/#scheduling-pods-on-windows) says to use `nodeSelector`. I know it's not a very good example in this context.
- Some tools still recommends to use `nodeSelector` and support it in the CLI, for example Helm.
- I understand the  [Kubernetes docs](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity) says `nodeSelector` will be eventually be deprecated, as [node affinity](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#node-affinity-beta-feature) can express everything, but I do not understand if it really replaces `nodeSelector` currently. Maybe some experts can answer this.
I will make some updates to this PR. Sorry again for late response.

AlexShemeshWix:
We use taints to assign app servers to proper instance types. 
We really need to be able to change tollerations to create consul on all nodes.

adilyse:
NodeSelector support from this PR was merged as part of #135, tolerations are now supported with the merge of #117.

Thanks for contributing!

PR #37, opened 2018-10-15
Add a Service account for the Consul Connect Injector to generate certificates automatically

anubhavmishra:
This adds `ClusterRole`, `ClusterRoleBinding`, and most importantly a `ServiceAccount` for the connect webhook injector to automatically generate certificates to access the Kubernetes API. This helps give a nice UX for a user using the Helm chart without providing the `connectInject.certs.secretName`. 

The `ClusterRole` has permissions to `get`, `list`, `watch` and `patch` on resources of kind `MutatingWebhookConfiguration`.


templates/connect-inject-deployment.yaml
```diff
@@ -50,7 +53,7 @@ spec:
                 -tls-cert-file=/etc/connect-injector/certs/{{ .Values.connectInject.certs.certName }}
                 -tls-key-file=/etc/connect-injector/certs/{{ .Values.connectInject.certs.keyName }}
 {{- else }}
-                -tls-auto=consul-connect-injector-cfg \
+                -tls-auto={{ template "consul.fullname" . }}-connect-injector-cfg \
```
anubhavmishra: The mutating webhook configuration name is actually set to `{{ template "consul.fullname" . }}-connect-injector-cfg` so I fixed it as part of this PR.

anubhavmishra:
Yea, totally agree on the tests. I will add them soon. :)

templates/connect-inject-deployment.yaml
adilyse: It would be good to rebase on master to pick up the change to ${CONSUL_FULLNAME} from your other [PR](https://github.com/hashicorp/consul-helm/pull/39) so we don't overwrite that fix.

mitchellh:
That doesn't look like a rebase @anubhavmishra ;) But works fine in this case.

PR #39, opened 2018-10-16
Fix connect inject configuration name

### LINKED ISSUE #36: name for auto generated key doesn't always match
Dynamic:
https://github.com/hashicorp/consul-helm/blob/master/templates/connect-inject-mutatingwebhook.yaml#L6
Static:
https://github.com/hashicorp/consul-helm/blob/master/templates/connect-inject-deployment.yaml#L53

I'd PR but I'm not sure how you'd want to handle that.. if it's variable or static in both locations just let me know and I can submit one.

anubhavmishra:
Doing this PR separately to fix #36 since it is independent of my [PR](https://github.com/hashicorp/consul-helm/pull/37) for RBAC for the injector.

/cc @mitchellh @adilyse

### DIFF
--- a/templates/connect-inject-deployment.yaml
+++ b/templates/connect-inject-deployment.yaml
@@ -50,7 +50,7 @@ spec:
                 -tls-cert-file=/etc/connect-injector/certs/{{ .Values.connectInject.certs.certName }}
                 -tls-key-file=/etc/connect-injector/certs/{{ .Values.connectInject.certs.keyName }}
 {{- else }}
-                -tls-auto={{ template "consul.fullname" . }}-connect-injector-cfg \
+                -tls-auto=${CONSUL_FULLNAME}-connect-injector-cfg \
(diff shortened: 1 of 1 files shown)


templates/connect-inject-deployment.yaml
```diff
@@ -50,7 +50,7 @@ spec:
                 -tls-cert-file=/etc/connect-injector/certs/{{ .Values.connectInject.certs.certName }}
                 -tls-key-file=/etc/connect-injector/certs/{{ .Values.connectInject.certs.keyName }}
 {{- else }}
-                -tls-auto=consul-connect-injector-cfg \
+                -tls-auto={{ template "consul.fullname" . }}-connect-injector-cfg \
```
adilyse: It looks like you can use the already defined `CONSUL_FULLNAME`, like it's being used in the `-tls-auto-hosts` flag below:

${CONSUL_FULLNAME}-connect-injector-cfg

anubhavmishra:
nice catch @adilyse. I have updated the PR with the changes you requested. :)

PR #41, opened 2018-10-17
namespace is now set using helm release

### LINKED ISSUE #38: connect-inject doesn't work if consul not placed in "default" namespace
In connect-inject-mutatingwebhook.yaml, namespace is hardcoded "default".
 
https://github.com/hashicorp/consul-helm/blob/b0176f6b27bef2efb4ded4f0c9ed8d8e4d761d99/templates/connect-inject-mutatingwebhook.yaml#L17

anubhavmishra:
This fixes #38.

### DIFF
--- a/templates/connect-inject-mutatingwebhook.yaml
+++ b/templates/connect-inject-mutatingwebhook.yaml
@@ -14,7 +14,7 @@ webhooks:
     clientConfig:
       service:
         name: {{ template "consul.fullname" . }}-connect-injector-svc
-        namespace: default
+        namespace: {{ .Release.Namespace }}
         path: "/mutate"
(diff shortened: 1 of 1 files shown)


mitchellh:
Can we add a unit test for this one too? `helm template` supports `--namespace` so we can easily test we do the right thing here. 😄

anubhavmishra:
@mitchellh I can add unit test for making sure namespace is set correctly. The reason I didn't add a unit test was that we will be testing whether helm is setting its pre-defined variables correctly. So, if we are testing pre-defined variables then we should test things like `{{ .Release.Name }}` that can be supplied via `--name` etc. What do you think?

Issue #43, opened 2018-10-17
Talking to agent with downward API fails when hosts have multiple IPs (ex. EKS w/ aws-cniI)


popopanda:
Hello,

I have an EKS cluster with 3 workers, and deployed Consul and Rabbitmq via Helm. Both services are running, and I am able to access the Consul UI.

Each eks-worker in AWS has a private IP, as well as a few secondary IPs, which are used for Pods.

I am attempting to get syncCatalog working, but I am getting connection refused.

```
2018-10-17T23:52:23.572Z [WARN ] to-consul/sink: error registering service: node-name=ip-10-20-132-50.ec2.internal service-name=halting-penguin-rabbitmq err="Put http://10.20.62.49:8500/v1/catalog/register: dial tcp 10.20.62.49:8500: connect: connection refused"
```

IP - 10.20.62.49 is a primary ip of one of my eks-workers, however the pod that runs consul server (8500) is actually running on a secondary IP - 10.20.61.120 on that same eks-worker.

```
kubectl describe pod consul-server-2
Name:           consul-server-2
Namespace:      default
Node:           ip-10-20-62-49.ec2.internal/10.20.62.49
Start Time:     Wed, 17 Oct 2018 16:12:01 -0700
Labels:         app=consul
                chart=consul-0.1.0
                component=server
                controller-revision-hash=consul-server-66c8b8459c
                hasDNS=true
                release=consul
                statefulset.kubernetes.io/pod-name=consul-server-2
Annotations:    consul.hashicorp.com/connect-inject: false
Status:         Running
IP:             10.20.61.120
```

How can I register services the Consul cluster via the pod IP address which is on my eks-worker's secondary IP, and not connect to the primary ip?

Please let me know if I need to provide more information

Thank you

popopanda:
Also, this is my values.yml I used

```
global:
  enabled: true

  # Domain to register the Consul DNS server to listen for.
  domain: consul
  image: "consul:1.3.0"
  imageK8S: "hashicorp/consul-k8s:0.2.0"
  datacenter: dc1

server:
  enabled: true
  image: null
  replicas: 3
  bootstrapExpect: 3 # Should <= replicas count

  storage: 10Gi
  storageClass: null

  connect: true

  resources: {}
  updatePartition: 0
  disruptionBudget:
    enabled: true
    maxUnavailable: null
  extraConfig: |
    {}
  extraVolumes: []
    # - type: secret (or "configMap")
    #   name: my-secret
[...]

PR #44, opened 2018-10-18
update DaemonSet to use host network

jipperinbham:
addresses https://github.com/hashicorp/consul-helm/issues/43

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -48,7 +48,10 @@ spec:
             secretName: {{ .name }}
             {{- end }}
         {{- end }}
-
+      {{- if .Values.client.hostNetwork.enabled }}
+      dnsPolicy: ClusterFirstWithHostNet
+      hostNetwork: true
+      {{- end }}
       containers:
         - name: consul
           image: "{{ default .Values.global.image .Values.client.image }}"
--- a/test/unit/client-daemonset.bats
+++ b/test/unit/client-daemonset.bats
@@ -73,6 +73,38 @@ load _helpers
 }
 
 #--------------------------------------------------------------------
+# hostNetwork 
+
+@test "client/DaemonSet: hostNetwork default" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/client-daemonset.yaml  \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.hostNetwork' | tee /dev/stderr)
+  [ "${actual}" = null ]
+}
+
+@test "client/DaemonSet: hostNetwork enabled" {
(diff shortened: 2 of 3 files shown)


mitchellh:
Hey @jipperinbham, thanks. I am kind of curious why this is necessary as its not super clear to me.

I've previously understood `hostNetwork` as exposing the host network interfaces into the pod. However, we use `hostPort` instead which should only expose a _single port_ and bind it to the host IP. These two seem to be separate functionality, and I'm not sure why we want the former.

The Kubernetes networking model[1] also states that all containers should be able to communicate to their nodes (and vice versa) without NAT, point 2. We tested this daemonset on a variety of K8S installations and found it to work without issue, without this change.

So my question is: why isn't port 8500 properly binding and accessible to the host IP by setting `hostPort` alone? Why is `hostNetwork` necessary as well? 

[1]: https://kubernetes.io/docs/concepts/cluster-administration/networking/#kubernetes-model

mitchellh:
Yes, I think that'd be acceptable. Thanks for digging into this.

I'm curious if `hostNetwork` is more broadly supported than `hostPort`. If so, then just using `hostNetwork` by default might make sense too. But let's cross that bridge a different time. A toggle for now would be great.

jipperinbham:
Ok, I've updated the PR to make `hostNetwork` be opt-in.

As far as the use of `hostNetwork` in the broader sense, IMO, the use case of running a DaemonSet where you want the Pod to be accessible over a defined port(s) is when it's typically used. I did a [naive search](https://github.com/helm/charts/search?p=3&q=hostNetwork+%26%26+DaemonSet&unscoped_q=hostNetwork+%26%26+DaemonSet) of the main helm charts repo but it's hard to say one way or another if it's used more broadly in the community.

Art3mK:
`hostPort` works ok for me on EKS with VPC CNI. They have added `hostPort` support recently, see 
aws/amazon-vpc-cni-k8s/pull/153

adilyse:
Following up on this-- @jipperinbham, does the fix @Art3mK mentioned make the current `hostPort` implementation work for you?

jipperinbham:
I haven't been able to test it but good to know it's likely fixed. Since this is an "opt-in" change, is it an issue to make it an option?

adilyse:
In an effort to keep the helm chart understandable and easy to use, we'd like to be judicious about which configuration values we add. If your underlying issue has been solved by AWS, I would prefer not to add this option at this point in time. 

If things change in the future, we can always revisit that decision.

sunadm:
#194

jipperinbham:
Closed this one since it's quite out of date and I can't really sign the CLA for these changes at this point.

Issue #10, opened 2018-09-26
Consul fails to start


mmisztal1980:
In my cluster I have [rook](https://rook.io) running. When provisioning the consul cluster, my helm values file looks like this:

```yaml
global:
  enabled: true
  domain: consul
  image: "consul:1.2.3"
  datacenter: dc1

server:
  enabled: "-"
  image: null
  replicas: 3
  bootstrapExpect: 3
  storage: 10Gi
  storageClass: rook-ceph-block
  connect: true
  resources: {}
  updatePartition: 0
  disruptionBudget:
    enabled: true
    maxUnavailable: null
  extraConfig: |
    {}
  extraVolumes: []
client:
  enabled: "-"
  image: null
  join: null
  resources: {}
  extraConfig: |
    {}
  extraVolumes: []
dns:
  enabled: "-"

ui:
  enabled: "-"
  service:
    enabled: true
    type: null

connectInject:
  enabled: false # "-" disable this by default for now until the image is public
  image: "TODO"
  default: false # true will inject by default, otherwise requires annotation
  caBundle: "" # empty will auto generate the bundle
  namespaceSelector: null

  certs:
    secretName: null
    caBundle: ""
    certName: tls.crt
    keyName: tls.key
```
In order to start the chart, I use the following cmdline:

```bash
helm install -f ./helm/values.digitalocean.yaml --name consul --namespace service-discovery ./helm
NAME:   consul
LAST DEPLOYED: Wed Sep 26 07:46:32 2018
NAMESPACE: service-discovery
STATUS: DEPLOYED

RESOURCES:
==> v1beta1/PodDisruptionBudget
NAME           MIN AVAILABLE  MAX UNAVAILABLE  ALLOWED DISRUPTIONS  AGE
consul-server  N/A            0                0                    1s

==> v1/Pod(related)
NAME             READY  STATUS             RESTARTS  AGE
consul-56lvs     0/1    ContainerCreating  0         1s
consul-jttwp     0/1    ContainerCreating  0         1s
consul-qpgdn     0/1    ContainerCreating  0         1s
consul-server-0  0/1    Pending            0         1s
consul-server-1  0/1    Pending            0         1s
consul-server-2  0/1    Pending            0         1s

==> v1/ConfigMap
NAME                  DATA  AGE
consul-client-config  1     1s
consul-server-config  1     1s

==> v1/Service
[...]

PR #46, opened 2018-10-22
#10 consul-server readinessProbe settings moved to values file

mmisztal1980:
This PR makes the consul-servers' stateful set readiness check configurable via the chart values.

This way, the operator gets to tweak the settings in case the check fails under the default settings.

This PR aims to unblock cases similar to the one describe in #10

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -146,11 +146,11 @@ spec:
                 - |
                   curl http://127.0.0.1:8500/v1/status/leader 2>/dev/null | \
                   grep -E '".+"'
-            failureThreshold: 2
-            initialDelaySeconds: 5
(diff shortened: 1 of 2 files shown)


adilyse:
@mmisztal1980 Unfortunately we won't be able to support every possible configuration option in the `values.yaml` file without making it incredibly complicated for users. In this case, there are a number of ways a user might unknowing change these values that will cause them all sorts of issues. So for now, these check variables don't warrant inclusion in the `values.yaml` file. That said, we're opening to reevaluating this decision in the future if things change.

I've written up some [information](https://www.consul.io/docs/platform/k8s/helm.html#customizing-the-helm-chart) about extending the Helm chart in cases like these. Hopefully that will address your situation!

Issue #34, opened 2018-10-12
enhancement: AWS internal ALB and NLB configuration support


Vadim-Zenin:
Hello,

Enhancement: AWS internal ALB and NLB configuration support would be nice for Consul on AWS EKS.
Reference: https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer

lkysow:
Closed by https://github.com/hashicorp/consul-helm/pull/105

PR #49, opened 2018-10-25
provide annotation to ui-service to utilize aws loadbalancers

popopanda:
This enables use of annotations to utilize AWS load balancers for consul ui. 
Addresses https://github.com/hashicorp/consul-helm/issues/34

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -133,9 +133,8 @@ ui:
   service:
     enabled: true
     type: null
-    # Define annotations to utilize AWS Load balancers according to https://docs.aws.amazon.com/eks/latest/userguide/load-balancing.html
+    # Provide annotations for Consul's ui service, such as utilizing cloud load balancers if service is of type LoadBalancer
     annotations: null
-        # service.beta.kubernetes.io/aws-load-balancer-internal: 0.0.0.0/0
 
 # syncCatalog will run the catalog sync process to sync K8S with Consul
 # services. This can run bidirectional (default) or unidirectionally (Consul
--- a/values.yaml
+++ b/values.yaml
@@ -133,6 +133,9 @@ ui:
   service:
     enabled: true
     type: null
+    # Define annotations to utilize AWS Load balancers according to https://docs.aws.amazon.com/eks/latest/userguide/load-balancing.html
+    # annotations:
+        # service.beta.kubernetes.io/aws-load-balancer-internal: 0.0.0.0/0
 
 # syncCatalog will run the catalog sync process to sync K8S with Consul
 # services. This can run bidirectional (default) or unidirectionally (Consul
--- a/values.yaml
+++ b/values.yaml
@@ -134,7 +134,7 @@ ui:
     enabled: true
     type: null
     # Define annotations to utilize AWS Load balancers according to https://docs.aws.amazon.com/eks/latest/userguide/load-balancing.html
-    # annotations:
+    annotations: null
         # service.beta.kubernetes.io/aws-load-balancer-internal: 0.0.0.0/0
 
 # syncCatalog will run the catalog sync process to sync K8S with Consul
--- a/templates/ui-service.yaml
+++ b/templates/ui-service.yaml
@@ -13,6 +13,10 @@ metadata:
     chart: {{ template "consul.chart" . }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
+  {{- if .Values.ui.service.annotations }}
+  annotations:
+  {{ toYaml .Values.ui.service.annotations | indent 2 }}
+  {{- end }}
 spec:
   selector:
     app: {{ template "consul.name" . }}


values.yaml
```diff
@@ -133,6 +133,9 @@ ui:
   service:
     enabled: true
     type: null
+    # Define annotations to utilize AWS Load balancers according to https://docs.aws.amazon.com/eks/latest/userguide/load-balancing.html
+    # annotations:
+        # service.beta.kubernetes.io/aws-load-balancer-internal: 0.0.0.0/0
```
krancour: This just seems generally useful for _any_ sort of annotation. Does it make sense to call out ELB-related concerns specifically?

values.yaml
```diff
@@ -133,6 +133,9 @@ ui:
   service:
     enabled: true
     type: null
+    # Define annotations to utilize AWS Load balancers according to https://docs.aws.amazon.com/eks/latest/userguide/load-balancing.html
+    # annotations:
```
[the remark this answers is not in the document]
popopanda: Thanks for the feedback

values.yaml
popopanda: You're right, thanks for the feedback.  I've updated the comments and removed aws specific

adilyse:
@popopanda These changes look great! Once there is test support for this addition, it will be ready to merge. I'm happy to answer any questions you might have about the test setup, or I could add the tests and merge if you'd prefer.

adilyse:
Sure! Let me find some examples for you.

First, a couple of general formatting things:

- Put the tests in the same order as the variables appear in the `values.yaml`. In this case, the tests will be added at the end of the file since it's the last option in the `ui` section.
- Start tests for a chart value with a header that saws what is being tested, like this:
```
# annotations
```

This isn't currently the case everywhere, but it would be good to add it wherever tests are added to keep things readable.

- Name the test based on what it's testing in this format:
```
{ section being tested }: { short description of the test case }
```

When adding tests to an existing file, the first section will be the same as the other tests that are already there.

## Test specifics

[Bats](https://github.com/bats-core/bats-core) provides a way to run commands in a shell and inspect the output in an automated way. In all of the tests in this repo, the base command being run is [helm template](https://docs.helm.sh/helm/#helm-template) which turns the templated files into straight yaml output. In this way, we're able to test that the various conditionals in the templates render as we would expect.

Each test defines the files that are rendered using the `-x` flag, then it might adjust chart values by adding `--set` flags as well. This output is piped to [yq](https://pypi.org/project/yq/), which allows us to pull out just the information we're interested in verifying by referencing its position in the hierarchy or giving information about it (like its length). The test passes or fails based on the conditional at the end in the square brackets that compares the output of our `helm template` piped to `yq` to whatever we set as the expected value.

The `| tee /dev/stderr ` pieces direct any terminal output of the commands to stderr so that it doesn't interfere with `bats`.

## Examples
Here are some examples of things that show effective tests for the conditions I asked for.

- [Disabled by default](https://github.com/hashicorp/consul-helm/blob/master/test/unit/ui-service.bats#L77-L84)

```
@test "ui/Service: no type by default" {
  cd `chart_dir`
      -x templates/ui-service.yaml  \
      . | tee /dev/stderr |
      yq -r '.spec.type' | tee /dev/stderr)
  [ "${actual}" = "null" ]
}
```

In this example, nothing is changed from the default templates (no `--set` flags), then we use `yq` to retrieve the value we're checking, `.spec.type`. This output is then compared against our expected value (`null` in this case) in the assertion `[ "${actual}" = "null" ]`


- [Value matches a specific value](https://github.com/hashicorp/consul-helm/blob/master/test/unit/ui-service.bats#L86-L94)
```
@test "ui/Service: specified type" {
  cd `chart_dir`
      -x templates/ui-service.yaml  \
      --set 'ui.service.type=LoadBalancer' \
      . | tee /dev/stderr |
      yq -r '.spec.type' | tee /dev/stderr)
  [ "${actual}" = "LoadBalancer" ]
}
```

This is very similar to the last example, except we've changed a default value with the `--set` flag and correspondingly changed the expected value.

 - [Value contains multiple specific values](https://github.com/hashicorp/consul-helm/blob/master/test/unit/sync-catalog-deployment.bats#L115-L132)
```
@test "syncCatalog/Deployment: to-k8s only" {
  cd `chart_dir`
      -x templates/sync-catalog-deployment.yaml  \
      --set 'syncCatalog.enabled=true' \
      --set 'syncCatalog.toConsul=false' \
      . | tee /dev/stderr |
      yq '.spec.template.spec.containers[0].command | any(contains("-to-consul=false"))' | tee /dev/stderr)
  [ "${actual}" = "true" ]

      -x templates/sync-catalog-deployment.yaml  \
      --set 'syncCatalog.enabled=true' \
      --set 'syncCatalog.toConsul=false' \
      . | tee /dev/stderr |
      yq '.spec.template.spec.containers[0].command | any(contains("-to-k8s"))' | tee /dev/stderr)
  [ "${actual}" = "false" ]
}
```
In this case, the same command is run twice in the same test. This can be used to look for several things in the same field, or making sure that something is not present that shouldn't be.

Whoops, looks like I got carried away! Now that I've written this all up, I've realized that this should be a guide to contributing for this repo. I'll work on adding this information to the readme for other folks too!

adilyse:
Hi @popopanda,

Thank you for your work on this! With the merging of #105, this functionality is now supported in the Helm chart.

PR #52, opened 2018-10-30
Added PodSecurityPolicies

liviudm:
In clusters with PodSecurityPolicies enabled we also need the policies. A Helm variable like `psp.enabled` can probably be added as well, I can add that logic as well if it's desired.

Server and client tested in Kubernetes 1.10.9 deployed with kops.

### DIFF
--- a/templates/connect-inject-clusterrole.yaml
+++ b/templates/connect-inject-clusterrole.yaml
@@ -17,4 +17,12 @@ rules:
     - "list"
     - "watch"
     - "patch"
+- apiGroups:
+  - policy
+  resources:
+  - podsecuritypolicies
+  resourcesName:
+  - {{ template "consul.fullname" . }}-connect-injector-webhook
+  verbs:
+  - use
 {{- end }}
--- a/templates/connect-inject-deployment.yaml
+++ b/templates/connect-inject-deployment.yaml
@@ -25,6 +25,7 @@ spec:
         release: {{ .Release.Name }}
         component: connect-injector
     spec:
+      securityContext: {}
   {{- if not .Values.connectInject.certs.secretName }}
       serviceAccountName: {{ template "consul.fullname" . }}-connect-injector-webhook-svc-account
   {{- end }}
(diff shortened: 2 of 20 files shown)


templates/connect-inject-deployment.yaml
```diff
@@ -25,6 +25,7 @@ spec:
         release: {{ .Release.Name }}
         component: connect-injector
     spec:
+      securityContext: {}
```
adilyse: Same question as previously-- if this is empty, does it need to be defined at all?

templates/connect-inject-clusterrole.yaml
```diff
@@ -17,4 +17,12 @@ rules:
     - "list"
     - "watch"
     - "patch"
+- apiGroups:
+  - policy
+  resources:
+  - podsecuritypolicies
+  resourcesName:
+  - {{ template "consul.fullname" . }}-connect-injector-webhook
```
adilyse: To match the name defined in the pod security policy:
```suggestion
  - {{ template "consul.fullname" . }}-connect-injector-webhook-deployment
```

templates/server-statefulset.yaml
```diff
@@ -47,6 +47,8 @@ spec:
                   component: server
               topologyKey: kubernetes.io/hostname
       terminationGracePeriodSeconds: 10
+      serviceAccount: {{ template "consul.fullname" . }}-server
```
adilyse: As in the client DaemonSet, this line should be removed.

templates/sync-catalog-deployment.yaml
```diff
@@ -28,6 +28,7 @@ spec:
       annotations:
         "consul.hashicorp.com/connect-inject": "false"
     spec:
+      securityContext: {}
```
adilyse: Again, wondering what facilitates.

liviudm:
Thank you for your review and suggestions @adilyse. It might take a couple of weeks to implement everything as I'm busy with other projects right now. I'll ping you as soon as I have everything ready. If anyone from your team is interested in contributing to this feature please let me know and I can give them write access to my fork.

adilyse:
These changes have been merged as part of #121. Thanks @liviudm for the great work on this!

PR #53, opened 2018-10-30
Update to latest consul-k8s version

### LINKED ISSUE #50: Chart version does not match tag (minor issue)
I checked out tag `v0.3.0` and found that [Chart.yaml](https://github.com/hashicorp/consul-helm/blob/master/Chart.yaml#L3) still shows `version: 0.1.0`.

I wonder: when will the next release occur?  I also found https://github.com/hashicorp/consul-helm/issues/26 and forked from 0.3.0 to work around the issue.

adilyse:
Additional updates the Chart.yaml version to match the current 0.3.0
release (fixes #50).

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: consul
-version: 0.1.0
+version: 0.3.0
 description: Install and configure Consul on Kubernetes.
 home: https://www.consul.io
 sources:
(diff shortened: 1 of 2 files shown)


Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: consul
-version: 0.1.0
+version: 0.3.0
```
anubhavmishra: We aren't doing minor version bumps?

adilyse: We will be doing versioning as expected moving forward. I think this was overlooked when we released 0.2.0 and 0.3.0, so I'm updating it to bring it in line with what currently exists. Moving forward, this should get updated every time we release.

PR #55, opened 2018-11-06
Add information about writing bats unit tests

adilyse:
This documentation gives a description of how the bats tests are
structured, as well as giving several examples of common test patterns.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,3 +66,111 @@ that can be used to quickly bring up a GKE cluster and configure
 `kubectl` and `helm` locally. This can be used to quickly spin up a test
 cluster for acceptance tests. Unit tests _do not_ require a running Kubernetes
 cluster.
+
+### Writing Unit Tests
+
+Changes to the Helm chart should be accompanied by appropriate unit tests.
+
+#### Formatting
+
+- Put tests in the test file in the same order as the variables appear in the `values.yaml`. 
+- Start tests for a chart value with a header that says what is being tested, like this:
+    ```
+    #--------------------------------------------------------------------
+    # annotations
+    ```
+
+- Name the test based on what it's testing in the following format:
+    ```
+    { section being tested }: { short description of the test case }
+    ```
+
+    When adding tests to an existing file, the first section will be the same as the other tests in the file.
+
+#### Test Details
+
+[Bats](https://github.com/bats-core/bats-core) provides a way to run commands in a shell and inspect the output in an automated way.
+In all of the tests in this repo, the base command being run is [helm template](https://docs.helm.sh/helm/#helm-template) which turns the templated files into straight yaml output.
+In this way, we're able to test that the various conditionals in the templates render as we would expect.
+
+Each test defines the files that should be rendered using the `-x` flag, then it might adjust chart values by adding `--set` flags as well.
+The output from this `helm template` command is then piped to [yq](https://pypi.org/project/yq/). 
+`yq` allows us to pull out just the information we're interested in, either by referencing its position in the yaml file directly or giving information about it (like its length).
+
+The test passes or fails based on the conditional at the end that is in square brackets, which is a comparison of our expected value and the output of  `helm template` piped to `yq`.
+
+The `| tee /dev/stderr ` pieces direct any terminal output of the `helm template` and `yq` commands to stderr so that it doesn't interfere with `bats`.
+
+#### Test Examples
+
+Here are some examples of common test patterns:
+
+- Check that a value is disabled by default
+
+    ```
+    @test "ui/Service: no type by default" {
+      cd `chart_dir`
+      local actual=$(helm template \
+          -x templates/ui-service.yaml  \
+          . | tee /dev/stderr |
+          yq -r '.spec.type' | tee /dev/stderr)
+      [ "${actual}" = "null" ]
+    }
+    ```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -66,3 +66,111 @@ that can be used to quickly bring up a GKE cluster and configure
...
+- Put tests in the test file in the same order as the variables appear in the `values.yaml`. 
+- Start tests for a chart value with a header that says what is being tested, like this:
+    ```
+    #--------------------------------------------------------------------
+    # annotations
+    ```
+
+- Name the test based on what it's testing in the following format:
+    ```
+    { section being tested }: { short description of the test case }
```
banks: This is fine but it might be even more clear if this line included the actual syntax you need to name a test (more like an example) e.g.

```
@test "<section being tested>: <short description of test case>" {
```

Or alternatively an example is given right after to make it completely obvious that in the examples below that string is what was being referred to here.

README.md
```diff
@@ -66,3 +66,111 @@ that can be used to quickly bring up a GKE cluster and configure
...
+      cd `chart_dir`
+      local actual=$(helm template \
+          -x templates/ui-service.yaml  \
+          . | tee /dev/stderr |
+          yq -r '.spec.type' | tee /dev/stderr)
+      [ "${actual}" = "null" ]
+    }
+    ```
+
+    In this example, nothing is changed from the default templates (no `--set` flags), then we use `yq` to retrieve the value we're checking, `.spec.type`.
```
banks: Worth mentioning `-r` outputs it as a raw string value not a quoted Yaml one?

README.md
```diff
@@ -66,3 +66,111 @@ that can be used to quickly bring up a GKE cluster and configure
...
+    This output is then compared against our expected value (`null` in this case) in the assertion `[ "${actual}" = "null" ]`
+
+
+- Check that a template value is rendered to a specific value
```
banks: Is it worth making each example an H5 heading? That might visually separate them a bit more but most importantly would allow deep links to specific examples? Totally optional and I forgot this is a README not part of a doc site but still.

README.md
```diff
@@ -66,3 +66,111 @@ that can be used to quickly bring up a GKE cluster and configure
...
+    ```
+    @test "ui/Service: specified type" {
+      cd `chart_dir`
+      local actual=$(helm template \
+          -x templates/ui-service.yaml  \
+          --set 'ui.service.type=LoadBalancer' \
+          . | tee /dev/stderr |
```
banks: is this missing a trailing `\` (same for all the others actually, I could be misreading the syntax though but it looks like all part of the same subshell command?

README.md
```diff
@@ -66,3 +66,111 @@ that can be used to quickly bring up a GKE cluster and configure
...
+          -x templates/sync-catalog-deployment.yaml  \
+          --set 'syncCatalog.enabled=true' \
+          --set 'syncCatalog.toConsul=false' \
+          . | tee /dev/stderr |
+          yq '.spec.template.spec.containers[0].command | any(contains("-to-k8s"))' | tee /dev/stderr)
+      [ "${actual}" = "false" ]
+    }
+    ```
+    In this case, the same command is run twice in the same test.
+    This can be used to look for several things in the same field, or to check that something is not present that shouldn't be.
```
banks: Is there a reason to do it like that vs something like:

```sh
local actualYAML=$(helm template \
          -x templates/sync-catalog-deployment.yaml  \
          --set 'syncCatalog.enabled=true' \
          --set 'syncCatalog.toConsul=false' \
          . | tee /dev/stderr)

local notToConsul=$(echo "actualYAML" | \
    yq '.spec.template.spec.containers[0].command | any(contains("-to-consul=false"))' | tee /dev/stderr)
[ "${notToConsul}" = "true" ]

          -x templates/sync-catalog-deployment.yaml  \
          --set 'syncCatalog.enabled=true' \
          --set 'syncCatalog.toConsul=false' \
          . | tee /dev/stderr)
local containsToK8s=$(yq '.spec.template.spec.containers[0].command | any(contains("-to-k8s"))' | tee /dev/stderr)
[ "${containsToK8s}" = "false" ]
```

No big deal just seems like it would be marginally more efficient and less copy paste to just run once and make multiple assertions on the output?

README.md
adilyse: Completely reasonable. I've updated the formatting to include the full line and made a note to that effect.

README.md
adilyse: Added a note about this.

README.md
adilyse: I tried it out, but it seems to impede the readability a bit. It's something to keep in mind in the future, especially if we add a bunch of additional examples.

README.md
adilyse: You had me worried here that all of our tests were not working right! It turns out that `|` isn't a valid command terminator, so it automatically continues to the next line (see the second answer [here](https://stackoverflow.com/questions/3871332/how-to-tell-bash-that-the-line-continues-on-the-next-line)), so we should be good.

README.md
adilyse: (I imagine the second `actual` definition was a copy/paste slip, since as written this adds additional work rather than removing it.)

In this case, I don't think the minimal redundancy reduction makes it worth changing the established pattern. For many contributors, this will be their first time working with `bats`, so keeping things as consistent as possible will reduce the overhead to understanding what's going on. In cases where more than just two conditions being tested on the same `helm template` output, combining the base command would definitely be preferable. I'll make a note of that in the comment for this example.

Issue #58, opened 2018-11-09
disruptionBudget maxUnavailable value is ignored if 0


topfunky:
**What I did:**

I set my pod disruption budget's `maxUnavailable` to `0` in my Helm values override file. I then installed the helm chart to a minikube cluster.

**What I saw:**

When running `helm install`, I saw an error that told me the `maxUnavailable` had been set to `-1`, not the `0` that I had used.

```
$ helm install -f helm-consul-values.yaml ./consul-helm

Error: release pouring-monkey failed: PodDisruptionBudget.policy "pouring-monkey-consul
-server" is invalid: spec.maxUnavailable: Invalid value: -1: must be greater than or eq
ual to 0
```

**What I expected:**

I expected that setting `maxUnavailable` to `0` would be interpreted by the chart as a valid value.

**Other context:**

If I set the value to `1`, everything works as expected. This suggests that a piece of logic is interpreting `0` as `false` and fails to use the value. Values `1` and greater are interpreted correctly.

My Helm values file has (this is the config that causes the above error):

```
server:
  replicas: 1
  bootstrapExpect: 1
  disruptionBudget:
    enabled: true
    maxUnavailable: 0
```

The issue may be related to code in `_helpers.tpl` which checks `if .Values.server.disruptionBudget.maxUnavailable` but should look for `if .Values.server.disruptionBudget.enabled`.

adilyse:
I've addressed the invalid `-1` value in [this PR](https://github.com/hashicorp/consul-helm/pull/60). There is unfortunately still a larger issue of not being able to explicitly set this value to `0` that is a limitation of Helm templating (see PR for details).

adilyse:
I've addressed this with the following changes:

- I've changed the calculation of the default `maxUnavailable`to provide a valid value when replicas=1. This should fix the issue in the vast majoring of cases.

- I've added documentation about how someone can directly set this value to `0`, since it's more involved than it should be.

Hopefully that helps!

PR #60, opened 2018-11-10
Provide a valid maxUnavailable value when using a single replica

adilyse:
This adds a condition to set `maxUnavailable` to 0 when replicas=1.
Previously, the logic returned an invalid value of `-1`. (See #58 )

The larger issue of not being able to set something explicitly to 0
remains, as this is a limitation of the way helm templating works.
For more information, see:

Helm documentation that a numerical 0 is treated identically to null:
https://docs.helm.sh/chart_template_guide/#if-else

Github issue for this:
https://github.com/helm/helm/issues/3164

Additionally cleans up some test descriptions for greater clarity.


anubhavmishra:
I am sure this was an interesting journey to this PR!

PR #75, opened 2018-12-03
Add flag to deployment: -k8s-source-namespace

frankabee:
Sync-Catalog version 0.2.1 was throwing errors for me without this additional setting.

Error message:
ERROR: logging before flag.Parse: E1203 15:35:47.427485       7 reflector.go:205] pkg/mod/k8s.io/client-go@v8.0.0+incompatible/tools/cache/reflector.go:99: Failed to list *v1.Service: services is forbidden: User "system:serviceaccount:swfactory:default" cannot list services at the cluster scope

### DIFF
--- a/templates/sync-catalog-deployment.yaml
+++ b/templates/sync-catalog-deployment.yaml
@@ -57,5 +57,6 @@ spec:
                 {{- if .Values.syncCatalog.k8sPrefix }}
                 -k8s-service-prefix="{{ .Values.syncCatalog.k8sPrefix}}" \
                 {{- end }}
-                -k8s-write-namespace=${NAMESPACE}
+                -k8s-write-namespace=${NAMESPACE} \
+                -k8s-source-namespace=${NAMESPACE}
 {{- end }}


adilyse:
Hi @frankabee ,

Which version of the helm chart are you using? We've recently added RBAC support for the sync-catalog operation, so we now create a cluster role, cluster role binding and a service account with the correct permissions for the catalog sync. Based on your error message, it looks like it's not using this service account (it would end in `-consul-sync-catalog`). Can you give it a try with the latest helm chart and let me know if that makes a difference?

frankabee:
Hi @adilyse ,

I am using commit cd180898555042678cd54f273db57bb458d8a746 (tag: v0.3.0). I will check it. 👍

frankabee:
@adilyse I think that you are right about closing it. The setup with the RBAC makes a lot of sense, because otherwise I need to setup a synch-catalog in each namespace. Also with my supposed flag this would be the case. I will check how I can get this running and will come back to you if some good additions are coming out. 👍 Thanks a lot for your support! 😃

PR #84, opened 2018-12-10
expose server port to host

yacut:
related to https://github.com/hashicorp/consul-helm/issues/85


lkysow:
Since multi-cluster federation is now supported and this PR was for that use case (https://github.com/hashicorp/consul-helm/issues/85) this can now be closed.

PR #91, opened 2018-12-30
Added aclRequestToken

stevegszabo:
Not sure if this is the best approach however we need a way to set CONSUL_HTTP_TOKEN

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -178,6 +178,9 @@ syncCatalog:
   # Set this to false to skip syncing ClusterIP services.
   syncClusterIPServices: true
 
+  # aclRequestToken used to set CONSUL_HTTP_TOKEN environment
+  aclRequestToken: null
+
 # ConnectInject will enable the automatic Connect sidecar injector.
 connectInject:
   enabled: false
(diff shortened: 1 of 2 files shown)


stevegszabo:
From the [docs](https://www.consul.io/docs/platform/k8s/service-sync.html) we should be able to provide this as a secret:

For Consul, the process accepts both the standard CLI flag -token and the environment variable CONSUL_HTTP_TOKEN. This should be set to an Consul ACL token if ACLs are enabled. **This can also be configured using the Helm chart to read from a Kubernetes secret.**

values.yaml
```diff
@@ -178,6 +178,9 @@ syncCatalog:
   # Set this to false to skip syncing ClusterIP services.
   syncClusterIPServices: true
 
+  # aclRequestToken used to set CONSUL_HTTP_TOKEN environment
+  aclRequestToken: null
```
adilyse: Rather than putting the token directly in to the values file, it would be more secure to reference a kubernetes secret instead. There's an example of this in PR #104 for the `enterpriseLicense` option.

adilyse:
@stevegszabo Thank you for submitting this! This functionality has been added in #111, so I'm going to close this, but we appreciate your help improving the product!

PR #93, opened 2019-01-04
fix consul-k8s:nodeport-syncing get node ip forbidden

hidaen:
project: consul-k8s branch nodeport-syncing get node ip forbidden
"to-consul/source: error getting node info: error="nodes "X.X.X.X" is forbidden: User "system:serviceaccount:default:consul-sync-catalog" cannot get nodes at the cluster scope""

### DIFF
--- a/templates/sync-catalog-cluster-role.yaml
+++ b/templates/sync-catalog-cluster-role.yaml
@@ -14,6 +14,7 @@ rules:
     resources:
       - services
       - endpoints
+      - nodes
     verbs:
       - get
       - list


adilyse:
Hi @hidaen 

I appreciate you adding permissions for the NodePort syncing, but I've ended up adding the `nodes` with only `get` permissions since that's all that's need (see PR #98 ). Thanks!

hidaen:
> Hi @hidaen
> 
> I appreciate you adding permissions for the NodePort syncing, but I've ended up adding the `nodes` with only `get` permissions since that's all that's need (see PR #98 ). Thanks!

Yes, you are right 👍

PR #101, opened 2019-01-21
use consul server service instead of local agent

grzegorzlyczba:
allows to use sync in setup with servers only

### DIFF
--- a/templates/sync-catalog-deployment.yaml
+++ b/templates/sync-catalog-deployment.yaml
@@ -33,10 +33,12 @@ spec:
         - name: consul-sync-catalog
           image: "{{ default .Values.global.imageK8S .Values.syncCatalog.image }}"
           env:
+{{- if not .Values.syncCatalog.useServerService }}
             - name: HOST_IP
               valueFrom:
                 fieldRef:
                   fieldPath: status.hostIP
+{{- end }}
             - name: NAMESPACE
               valueFrom:
                 fieldRef:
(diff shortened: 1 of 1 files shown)


justinwalz:
I just made the same change locally - then checked PRs to see if someone had the same idea. Can we please get this merged? 

Going to tag @adilyse as you're the top contributor for the project.

lkysow:
Hi @soymo can you describe why you need this change?

We always recommend applications (including our own sync) talk to the local agent because otherwise performance issues arise.

justinwalz:
I can speak for soymo, but our use case is that we're only running consul in order to sync services with the k8s sync project. It's overkill for us to run a client on every node (especially for large clusters), so we prefer to just hit the three servers directly via the k8s service.

lkysow:
Thanks for the reply Justin.
A follow-up: which way are you syncing, consul => kubernetes or kubernetes => consul. Also do you have Consul servers and clients running outside of Kubernetes? Or are your servers only in Kubernetes and your clients are all outside of Kubernetes and using the Consul servers in Kube?

justinwalz:
We are syncing kubernetes => consul.

We use the consul service registry + consul-template to generate nginx upstreams for a reverse proxy we run. The kubernetes servers are in k8s, via this helm chart. The only user of this setup is the proxy using consul-template, no consul agents running in client mode.

lkysow:
Hi @grzegorzlyczba, thanks for the PR! This is now supported without the extra setting. If clients are `false` then we will automatically use the servers for sync catalog. So we're going to close this PR for now. Happy to re-open if this doesn't work for your use case.

PR #105, opened 2019-01-25
Add additional configuration options for UI Service

lawliet89:
Including additional annotations and spec for the service


lawliet89:
Could I get this looked at please? I use this to set additional annotations for my cloud provider LB.

PR #112, opened 2019-02-07
Making the metadata volume claim metadata name configurable

### LINKED ISSUE #110: Installing on multiple namespaces on EKS
While trying to install the helm chart on EKS I kept getting failures in provisioning the PVC, with the error that the storage class was not given (it was, in the chart).

After digging around, I remembered that we have the same chart installed on another namespace on the same cluster.
Changing the metadata name in the volumeClaimTemplates that the chart generates manually solved this problem.

I am not sure why this is happening as the actual PV generated at the end have different names and, of course, the pvc are namespaced. Still, it might me worthwhile to make that name a variable in the values.

### How to reproduce

Install, at the very least, the server from the helm chart. Then do that again in a different namespace.

kfirbreger:
This resolves #110 by adding the metadata name it the values.yaml

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -101,7 +101,7 @@ spec:
                 {{- end }}
                 -server
           volumeMounts:
-            - name: data
+            - name: {{ .Values.server.storageClaimName }}
               mountPath: /consul/data
             - name: config
               mountPath: /consul/config
@@ -153,7 +153,7 @@ spec:
           {{- end }}
   volumeClaimTemplates:
     - metadata:
-        name: data
+        name: {{ .Values.server.storageClaimName }}
       spec:
         accessModes:
           - ReadWriteOnce
--- a/values.yaml
+++ b/values.yaml
@@ -41,6 +41,7 @@ server:
   # to null (the Kube cluster will pick the default).
   storage: 10Gi
   storageClass: null
+  storageClaimName: "data"
 
   # connect will enable Connect on all the servers, initializing a CA
   # for Connect-related connections. Other customizations can be done


templates/server-statefulset.yaml
```diff
@@ -101,7 +101,7 @@ spec:
                 {{- end }}
                 -server
           volumeMounts:
-            - name: data
+            - name: {{ .Values.server.storageClaimName }}
```
adilyse:
```suggestion
            - name: data-{{ .Release.Namespace }}
```

templates/server-statefulset.yaml
```diff
@@ -153,7 +153,7 @@ spec:
           {{- end }}
   volumeClaimTemplates:
     - metadata:
-        name: data
+        name: {{ .Values.server.storageClaimName }}
```
adilyse:
```suggestion
        name: data-{{ .Release.Namespace }}
```

values.yaml
```diff
@@ -41,6 +41,7 @@ server:
   # to null (the Kube cluster will pick the default).
   storage: 10Gi
   storageClass: null
+  storageClaimName: "data"
```
adilyse: Making the above-suggested changes means that this is no longer necessary.

kfirbreger: Hi @adilyse 
I am no longer running two installations in different namespaces. If this solves the problem than its all good. I can't really test it though. Have you?

s-christoff:
Changes to support this have been merged as part of #132 
Thank you so much @kfirbreger  for the contribution!

SamDecrock:
I don't get it. Why is the namespace part of the pvc name if the pvc is namespaced anyway?

PR #114, opened 2019-02-08
Update consul and consul-k8s default versions

adilyse:


values.yaml
```diff
@@ -14,14 +14,14 @@ global:
   # servers below. This can be overridden per component.
   #
   # Examples:
-  #   image: "consul:1.4.0"
-  #   image: "consul:1.4.0-ent"   # Enterprise Consul image
-  image: "consul:1.3.1"
+  #   image: "consul:1.4.1"
+  #   image: "consul:1.4.1-ent"   # Enterprise Consul image
+  image: "consul:1.4.1"
```
pearkes: We should be doing 1.4.2 here as it included some important changes.

adilyse: Updated!

PR #119, opened 2019-02-18
Allow to specify custom annotations

MiLk:
[kube2iam](https://github.com/jtblin/kube2iam#kubernetes-annotation) and [kiam](https://github.com/uswitch/kiam#overview) require to be able to assign a role to the pod and have the necessary permissions to use the Cloud Auto Join feature on AWS.

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -28,6 +28,9 @@ spec:
         hasDNS: "true"
       annotations:
         "consul.hashicorp.com/connect-inject": "false"
+        {{- if .Values.client.annotations }}
+{{ toYaml .Values.client.annotations | indent 8 }}
+        {{- end }}
     spec:
       terminationGracePeriodSeconds: 10
 
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -29,7 +29,7 @@ spec:
       annotations:
         "consul.hashicorp.com/connect-inject": "false"
         {{- if .Values.client.annotations }}
-          {{- toYaml .Values.client.annotations | nindent 8 }}
+          {{- tpl .Values.client.annotations . | nindent 8 }}
         {{- end }}
     spec:
       terminationGracePeriodSeconds: 10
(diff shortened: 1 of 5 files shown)


templates/client-daemonset.yaml
```diff
@@ -28,6 +28,9 @@ spec:
         hasDNS: "true"
       annotations:
         "consul.hashicorp.com/connect-inject": "false"
+        {{- if .Values.client.annotations }}
+{{ toYaml .Values.client.annotations | indent 8 }}
```
s3than: You can use nindent which allows easier readability see https://github.com/hashicorp/consul-helm/blob/master/templates/server-statefulset.yaml#L43

MiLk:
Thank you for your review.
I've added simple tests.

pviniciusfm:
Any updates on that?

values.yaml
```diff
@@ -106,6 +106,9 @@ server:
               component: server
           topologyKey: kubernetes.io/hostname
 
+  # Extra annotations to attach to the server pods
+  annotations: {}
```
adilyse: For complicated inputs in the values.yaml file, we've standardized on using multi-line strings. For consistency, this value should use the same pattern. This would make the annotation definition work like this:
```
  annotations: |
    "my-annotation": "my-annotation-value"
    "foo": "bar"
```

These values are processed in the template files with `tpl` instead of `toYaml1`. A good recent example of this is #105.

values.yaml
```diff
@@ -106,6 +106,9 @@ server:
               component: server
           topologyKey: kubernetes.io/hostname
 
+  # Extra annotations to attach to the server pods
```
adilyse: This comment should include some information about the expected format of the input for the value.

values.yaml
```diff
@@ -137,6 +140,9 @@ client:
     #   name: my-secret
     #   load: false # if true, will add to `-config-dir` to load by Consul
 
+  # Extra annotations to attach to the client pods
+  annotations: {}
```
adilyse: Same note here as for the server annotations format.

MiLk:
I believe I've done the requested changes as expected.
Let me know if it's good.

MiLk:
I've rebased on master. It should be ready to merge now.

PR #121, opened 2019-02-26
Add podSecurityPolicy support to the Consul Helm chart

adilyse:
This builds on the work done in #52 by adding tests and a global variable to enable the feature. 

As part of this work, cluster roles, cluster role bindings and service accounts were added as the default for the client and server components. There was also some formatting and file name standardization as part of the change.


adilyse:
Hi @liviudm ,

Thanks for the great work getting the majority of the functionality added for this! I've added tests and a global option that allows folks to opt in or out of the feature.

Would you be willing to sign the new Contributor License Agreement so I can get this merged? Clicking on the `Details` link in the `license/cla` check should take you to the correct place.

Thanks again!

templates/client-podsecuritypolicy.yaml
```diff
@@ -0,0 +1,43 @@
...
+  hostPorts:
+  - min: 8500
+    max: 8502
+  hostIPC: false
+  hostPID: false
+  runAsUser:
+    # Require the container to run without root privileges.
+    rule: 'RunAsAny'
+  seLinux:
+    # This policy assumes the nodes are using AppArmor rather than SELinux.
```
lkysow: This comment isn't super clear to me. As a user is this telling me I should use apparmor?

Also from a quick scan of the docs, app armour profiles are enabled via annotations on the `PodSecurityPolicy`, but there's no annotations on this policy

adilyse: Ah, good catch! I've removed the AppArmor-specific comments.

adilyse:
Hi @liviudm,

PodSecurityPolicy support is ready to be merged as soon as you've signed the CLA. Looking foward to adding this to the Helm chart!

PR #122, opened 2019-02-27
Add support for custom environment variables

pviniciusfm:
To support [cloud auto join](https://www.consul.io/docs/agent/cloud-auto-join.html) capabilities in our infrastructure, we needed to add proxies in the client daemon set. This feature will also support custom consul parameters.

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -51,3 +51,15 @@ Add a special case for replicas=1, where it should default to 0 as well.
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Inject extra environment vars in the format key:value, if populated
+*/}}
+{{- define "consul.extraEnvironmentVars" -}}
+{{- if .extraEnvironmentVars -}}
+{{- range $key, $value := .extraEnvironmentVars }}
+- name: {{ $key }}
+  value: {{ $value | quote }}
+{{- end -}}
+{{- end -}}
+{{- end -}}
--- a/values.yaml
+++ b/values.yaml
@@ -38,6 +38,15 @@ server:
   replicas: 3
   bootstrapExpect: 3 # Should <= replicas count
 
+  # extraEnvVars is a list of extra enviroment variables to set with the stateful set. These could be
+  # used to include proxy settings required for cloud auto-join feature, 
+  # in case kubernetes cluster is behind egress http proxies. Additionally, it could be used to configure
+  # custom consul parameters.
+  extraEnvironmentVars: {}
+    # http_proxy: http://localhost:3128
+    # https_proxy: http://localhost:3128
+    # no_proxy: internal.domain.com
+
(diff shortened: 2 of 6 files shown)


pviniciusfm:
This feature and #119 allow consul cloud auto join feature to work in k8 clusters that are behind corporate proxies.

templates/_helpers.tpl
```diff
@@ -51,3 +51,15 @@ Add a special case for replicas=1, where it should default to 0 as well.
...
 {{- end -}}
+
+{{/*
+Inject extra environment vars in the format key:value, if populated
+*/}}
+{{- define "consul.extraEnvironmentVars" -}}
+{{- if .extraEnvironmentVars -}}
+{{- range $key, $value := .extraEnvironmentVars }}
+- name: {{ $key }}
+  value: {{ $value | quote }}
```
adilyse: Are there situations where a user might want their value passed through without quotes? If so, is there another way to pass quotes through when folks want them?

values.yaml
```diff
@@ -38,6 +38,15 @@ server:
   replicas: 3
   bootstrapExpect: 3 # Should <= replicas count
 
+  # extraEnvVars is a list of extra enviroment variables to set with the stateful set. These could be
+  # used to include proxy settings required for cloud auto-join feature, 
+  # in case kubernetes cluster is behind egress http proxies. Additionally, it could be used to configure
+  # custom consul parameters.
+  extraEnvironmentVars: {}
```
adilyse: To maintain consistency in variable ordering with the `client` section, this would fit better coming after the `extraVolumes` option under `server`.

values.yaml
```diff
@@ -38,6 +38,15 @@ server:
...
+    # http_proxy: http://localhost:3128
```
```suggestion
    # http_proxy: http://localhost:3128,
```

This set of example text is slightly misleading, since commas are required between each key/value pair.

values.yaml
```diff
@@ -38,6 +38,15 @@ server:
...
+    # https_proxy: http://localhost:3128
```
```suggestion
    # https_proxy: http://localhost:3128,
```

values.yaml
```diff
@@ -137,6 +146,15 @@ client:
     #   name: my-secret
     #   load: false # if true, will add to `-config-dir` to load by Consul
 
+  # extraEnvVars is a list of extra enviroment variables to set with the pod. These could be
+  # used to include proxy settings required for cloud auto-join feature, 
+  # in case kubernetes cluster is behind egress http proxies. Additionally, it could be used to configure
+  # custom consul parameters.
+  extraEnvironmentVars: {}
+    # http_proxy: http://localhost:3128
```
```suggestion
    # http_proxy: http://localhost:3128,
```

values.yaml
```diff
@@ -137,6 +146,15 @@ client:
...
+    # https_proxy: http://localhost:3128
```
```suggestion
    # https_proxy: http://localhost:3128,
```

PR #134, opened 2019-03-15
Support gossip encryption

### LINKED ISSUE #133: Support Gossip encryption in Helm chart
Add an option to `values.yaml` to allow enabling Gossip encryption between nodes.

tradel:
Fixes #133


values.yaml
```diff
@@ -37,6 +37,11 @@ global:
   # chart. See https://kubernetes.io/docs/concepts/policy/pod-security-policy/
   enablePodSecurityPolicies: false
 
+  # Gossip encryption key. To disable gossip encryption, set this to null.
+  # Note: DO NOT use the provided key in production. Generate your own with
+  # "consul keygen" and paste it here.
+  gossipKey: cg8StVXbQJ0gPvMd9o7yrg==
```
adilyse: It is generally not a good idea to include sensitive information in a configuration file since many people check it into source control. Like we do elsewhere in the chart, we should recommend folks create a Kubernetes secret that contains this information, and include `secretName` and `secretKey` fields here instead. The `.server.enterpriseLicense` or `syncCatalog.aclSyncToken` are good examples of this pattern.

tradel: Of course you're right. Not sure why I didn't do it this way in the first place.

test/unit/gossip-client-daemonset.bats
```diff
@@ -0,0 +1,71 @@
...
+      -x templates/client-daemonset.yaml  \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[] | select(.name=="consul") | .env[] | select(.name == "GOSSIP_KEY") | length > 0' | tee /dev/stderr)
+  [ "${actual}" = "" ]
+}
+
+@test "global/gossipEncryption: disabled in client DaemonSet when servers are disabled" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/enterprise-license.yaml  \
```
adilyse: This should presumably still be templating the client daemonset. Additionally, I would expect that client gossip encryption could be enabled even if servers within the Kubernetes cluster are disabled. If the clients are talking to an external Consul cluster, they still might need a gossip encryption key, right?

Oh, or maybe this was a copy/paste issue and you meant the clients were disabled?

```suggestion
      -x templates/client-daemonset.yaml  \
```

test/unit/gossip-client-daemonset.bats
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env bats
+
+load _helpers
+
+@test "global/gossipEncryption: disabled in client DaemonSet by default" {
```
adilyse: It would be good to check that the `-encrypt` flag isn't present in this case as well.

test/unit/gossip-server-statefulset.bats
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env bats
+
+load _helpers
+
+@test "global/gossipEncryption: disabled in server StatefulSet by default" {
```
adilyse: As with the client test, it would be good to check that the `-encrypted` flag does not get set in this case.

test/unit/gossip-client-daemonset.bats
tradel: Accepted your change

test/unit/gossip-server-statefulset.bats
tradel: Fixed in 87c49d849b8231ff5d628f518b2b20a44d19246e.

adilyse:
Shoot, looks like this is going to need a rebase. There's some info in the [readme](https://github.com/hashicorp/consul-helm#rebasing-contributions-against-master) if you'd be willing to do the update. Otherwise I can take a look tomorrow.

tradel:
Manually merged into master.

Issue #136, opened 2019-03-18
Support TLS encryption


tradel:
Goals:
* Enable HTTPS listener on servers with user-provided certs
* Enable HTTPS connections between client and server
* Support HTTPS in the UI service

PR #137, opened 2019-03-18
Support TLS encryption

tradel:
Adds support for TLS encryption between agents using self-generated, self-signed TLS certs. Also adds the option to disable HTTP completely if `global.tls.httpsOnly` is set.


tradel:
Implements #136.

templates/client-daemonset.yaml
```diff
@@ -85,6 +93,18 @@ spec:
                 -advertise="${POD_IP}" \
                 -bind=0.0.0.0 \
                 -client=0.0.0.0 \
+                {{- if .Values.global.tls.enabled }}
+                -hcl='ca_file = "/consul/tls/ca/tls.crt"' \
+                -hcl='cert_file = "/consul/tls/client/tls.crt"' \
+                -hcl='key_file = "/consul/tls/client/tls.key"' \
+                -hcl='verify_incoming = {{ .Values.global.tls.verifyIncoming }}' \
```
i0rek: There is also `verify_incoming_https` and `verify_incoming_rpc` which is useful if you want to use the UI via HTTPS.

tradel: I think I like that approach better. Less pain to go through with HTTPS clients.

tradel: Fixed with dd7a23536bc179d0f0b8a646c94bea291d3e3c00

asteven:
Any chance that this PR is merged in the near future?
TLS is like the most important missing bit to make these helm charts really useful.

FYI: As I want to test this and it didn't merge I have squashed the commits in this PR and rebased on master while fixing the conflicts. https://github.com/asteven/consul-helm/tree/tls-encryption

tradel:
Hi Steven. We delayed merging this because there are other parts of the Consul ecosystem (such as Consul-k8s and Envoy) that need to support TLS too. 

We are working on bringing everything together soon so that Consul can be fully secured.

zakkg3:
Hi, Guys, I am using this code to do something else in another chart. Thanks! :)
I found curl it's printing the whole secret to stdout.... so if you are using ELK or something collecting logs... well I guess you have my point... so maybe its better to create the secrets with:

    curl -s -X POST  bla bla bla > /dev/null

filintod:
one question, does this actually work? when I tried the rpc was not working because the ip advertised is not part of the cert and the ip also changes every time the pod is restarted.

banks:
Hey!

If you're having trouble you might get more visibility asking over in
https://discuss.hashicorp.com/c/consul.

To my knowledge this is working for others. In general Consul doesn't
validate IP addresses, only server names and that the cert was signed by
the right root.

janbotech:
Hello guys, what is the current status of tls support?

filintod:
I think you are missing updating the sync-catalog-deployment.yaml file or this whole TLS will be of no use with httpsOnly.

janbotech:
@tradel I've been trying to deploy consul using this helm chart on the EKS k8s cluster. 

Now the project that I've been working on requires me to implement only tls encrypted traffic between pods. 

Is it maybe possible that this is merged in the near future or you would suggest me to change the way of deployment (not use this helm chart) in order to implement encrypted traffic? 

This is really big decision and would appreciate if you can help me with a time frame and thank you very much is support great work on this repo!

ishustava:
Hi @janbotech,

Thanks for sharing your use case with us. Unfortunately, I don't have a specific time frame, but this feature is currently our top priority, and I'm hoping to merge this (or another PR based on this) very soon.

ishustava:
We've merged #313, which was heavily based on this PR (thanks @tradel)!

Please see the [latest Helm release](https://github.com/hashicorp/consul-helm/releases/tag/v0.16.0) for the list of all changes.

Thank you all!

PR #144, opened 2019-03-28
Tests are now optional

michaelgeorgeattard:
Tests can be toggled off to avoid changes across installations using https://github.com/roboll/helmfile and similar tools.

Alternatively, https://github.com/hashicorp/consul-helm/pull/143

(This is the safer option)

### DIFF
--- a/templates/tests/test-runner.yaml
+++ b/templates/tests/test-runner.yaml
@@ -1,3 +1,4 @@
+{{- if (.Values.tests.enabled) }}
 apiVersion: v1
 kind: Pod
 metadata:
@@ -29,3 +30,4 @@ spec:
             [ `consul kv get _consul_helm_test` = "$VALUE" ]
             consul kv delete _consul_helm_test
   restartPolicy: Never
+{{- end }}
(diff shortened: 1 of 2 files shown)


oceyral:
Pretty much must-have, since the test pod cannot be created PodSecurityPolicies are activated (the test pod doesn't have one, or a service account).

lkysow:
> Pretty much must-have, since the test pod cannot be created when PodSecurityPolicies are activated (the test pod doesn't have one, or a service account).

This is only created when you purposefully run `helm test` though right?

lkysow:
Closed by #143.

@oceyral if you can help me understand better why this is necessary then I can reopen!

oceyral:
@lkysow They are also generated with `helm template`. In our case, we are not using tiller, but the output of helm template.

I did not remember about them only being generated through `helm test` so it is less of an issue than I thought, but still nice to have.

lkysow:
> @lkysow They are also generated with `helm template`. In our case, we are not using tiller, but the output of helm template.
> 
> I did not remember about them only being generated through `helm test` so it is less of an issue than I thought, but still nice to have.

Ahh I understand. Yes it looks like using `helm template` causes it to be generated. I'll re-open.

PR #150, opened 2019-04-22
Remove the "enabled" flag for gossip encryption

tradel:


adilyse:
Merged as #162, which was a rebased version of this PR.

PR #151, opened 2019-04-23
Implement `bootstrapACLs` option

adilyse:
This new config option allows folks creating an entire Consul cluster
(both servers and clients, optionally catalog sync) in Kubernetes to
automatically enable ACLs and bootstrap the system with server, client
and (optionally) catalog sync tokens.

It does this by applying a `server-acl-init` job after the servers have
started up to bootstrap the servers and create client and sync tokens
that are written to Kubernetes secrets. The client and sync pods have
an init container when this option is enabled that waits for these
secrets to be populated, then applies them once they become available.

Manual creation of other ACL toekns is possible by extracting the original
bootstrap token out of its Kubernetes secret and connecting directly to
a Consul server.

At the moment, the secrets are not automatically cleaned up when the Helm
chart is deleted. So, like PersistentVolumeClaims, these will need to be
manually deleted.

### DIFF
--- a/templates/client-clusterrole.yaml
+++ b/templates/client-clusterrole.yaml
@@ -8,13 +8,24 @@ metadata:
     chart: {{ template "consul.chart" . }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
-{{- if .Values.global.enablePodSecurityPolicies }}
+{{- if (or .Values.global.enablePodSecurityPolicies .Values.global.bootstrapACLs) }}
 rules:
+{{- end }}
+{{- if .Values.global.enablePodSecurityPolicies }}
   - apiGroups: ["policy"]
     resources: ["podsecuritypolicies"]
     resourceNames:
     - {{ template "consul.fullname" . }}-client
     verbs:
     - use
+{{- end }}
+{{- if .Values.global.bootstrapACLs }}
+  - apiGroups: [""]
+    resources:
+      - secrets
+    resourceNames:
+      - {{ .Release.Name }}-consul-client-acl-token
+    verbs:
+      - get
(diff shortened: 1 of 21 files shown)


templates/client-clusterrole.yaml
```diff
@@ -8,13 +8,22 @@ metadata:
...
   - apiGroups: ["policy"]
     resources: ["podsecuritypolicies"]
     resourceNames:
     - {{ template "consul.fullname" . }}-client
     verbs:
     - use
 {{- end }}
+{{- if .Values.global.bootstrapACLs }}
+  - apiGroups: [""]
+    resources:
```
rboyer: If possible it might be good to figure out how to scope this to just _your_ secrets using `resourceNames` instead of "all secrets": https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources

values.yaml
```diff
@@ -24,7 +24,7 @@ global:
   # Note: support for the catalog sync's liveness and readiness probes was added
   # to consul-k8s v0.6.0. If using an older consul-k8s version, you may need to
   # remove these checks to make the sync work.
-  imageK8S: "hashicorp/consul-k8s:0.7.0"
+  imageK8S: "gcr.io/consul-k8s/consul-k8s-dev:clientacl4"
```
rboyer: intentional?

templates/client-clusterrole.yaml
adilyse: Restricted the client and catalog sync roles to only refer to their specific secret.

values.yaml
adilyse: Left in for testing purposes. Removed.

PR #152, opened 2019-04-29
Add K8sConsulPrefix

s-christoff:
Updates values file to allow users to prefix k8s services registered within consul with a string.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -274,6 +274,12 @@ syncCatalog:
   # prepended with "consul-". (Consul -> Kubernetes sync)
   k8sPrefix: null
 
+  # consulPrefix is the service prefix which preprends itself
+  # to Kubernetes services registered within Consul
+  # For example, "k8s-" will register all services peprended with "k8s-".
+  # (Kubernetes -> Consul sync)
+  consulPrefix: null
+
   # k8sTag is an optional tag that is applied to all of the Kubernetes services
   # that are synced into Consul. If nothing is set, defaults to "k8s".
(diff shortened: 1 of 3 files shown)


values.yaml
```diff
@@ -260,7 +260,7 @@ ui:
 # Consul agent.
 syncCatalog:
   # True if you want to enable the catalog sync. "-" for default.
-  enabled: false
+  enabled: true
```
adilyse: This should stay `false` for the default helm chart.

values.yaml
```diff
@@ -274,6 +274,12 @@ syncCatalog:
   # prepended with "consul-". (Consul -> Kubernetes sync)
   k8sPrefix: null
 
+  # consulServicePrefix is the service prefix which preprends itself 
+  # to Kubernetes services registered within Consul
```
adilyse: Nit:
```suggestion
  # to Kubernetes services registered within Consul.
```

values.yaml
```diff
@@ -274,6 +274,12 @@ syncCatalog:
...
+  # For example, "k8s-" will register all services peprended with "k8s-". 
+  # (Kubernetes -> Consul sync)
+  consulServicePrefix: null
```
adilyse: Does it make sense to name this `consulPrefix` to be consistent with `k8sPrefix`?

values.yaml
s-christoff: 🤦‍♀️sorry, I thought I fixed this too.

PR #154, opened 2019-05-01
added default logLevel and consulWriteInterval

FullGC:
log-level and consul-write-interval are important configuration properties that should be exposed to edit in the values.yaml (The defaults are similar to these in 'consul-k8s')


lkysow:
Hi Dani, thanks for the PR! I set the syncInterval to nil by default so that it uses the default from `consul-k8s` but like that the logLevel is set in the helm chart so users can see its value and know right away how to change it.

PR #155, opened 2019-05-01
Support auth method in connect-inject

adilyse:
Updates the server-acl-init job to also register Kuberentes as an
auth method for Connect. Additionally adds an init container for the
connect injection deployment to wait for the resultant acl token to
be available in a Kubernetes secret.

Updates rbac resources to use v1 instead of the beta tag. Removes the
helper method that stopped name duplication if the release is explicitly
named `consul`, since this breaks several necessary naming assumptions.

### DIFF
--- a/templates/connect-inject-authmethod-clusterrolebinding.yaml
+++ b/templates/connect-inject-authmethod-clusterrolebinding.yaml
@@ -1,0 +1,39 @@
+{{- if and (not .Values.connectInject.certs.secretName) (or (and (ne (.Values.connectInject.enabled | toString) "-") .Values.connectInject.enabled) (and (eq (.Values.connectInject.enabled | toString) "-") .Values.global.enabled)) }}
+{{- if .Values.global.bootstrapACLs }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ template "consul.fullname" . }}-connect-injector-authmethod-authdelegator-role-binding
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: "system:auth-delegator"
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "consul.fullname" . }}-connect-injector-authmethod-svc-account
+    namespace: {{ .Release.Namespace }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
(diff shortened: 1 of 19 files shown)


templates/connect-inject-deployment.yaml
```diff
@@ -57,8 +57,11 @@ spec:
                 -tls-key-file=/etc/connect-injector/certs/{{ .Values.connectInject.certs.keyName }}
 {{- else }}
                 -tls-auto=${CONSUL_FULLNAME}-connect-injector-cfg \
-                -tls-auto-hosts=${CONSUL_FULLNAME}-connect-injector-svc,${CONSUL_FULLNAME}-connect-injector-svc.${NAMESPACE},${CONSUL_FULLNAME}-connect-injector-svc.${NAMESPACE}.svc
+                -tls-auto-hosts=${CONSUL_FULLNAME}-connect-injector-svc,${CONSUL_FULLNAME}-connect-injector-svc.${NAMESPACE},${CONSUL_FULLNAME}-connect-injector-svc.${NAMESPACE}.svc \
```
rboyer: Does this trailing `\` break anything when `bootstrapACLs` is off now?

adilyse: I've not had any trouble with it, but probably better to be safe.

templates/connect-inject-authmethod-clusterrolebinding.yaml
```diff
@@ -0,0 +1,39 @@
+{{- if and (not .Values.connectInject.certs.secretName) (or (and (ne (.Values.connectInject.enabled | toString) "-") .Values.connectInject.enabled) (and (eq (.Values.connectInject.enabled | toString) "-") .Values.global.enabled)) }}
+{{- if .Values.global.bootstrapACLs }}
+apiVersion: rbac.authorization.k8s.io/v1beta1
```
rboyer:
```suggestion
apiVersion: rbac.authorization.k8s.io/v1
```
I don't think the beta1 suffix is required anymore.

templates/connect-inject-authmethod-clusterrolebinding.yaml
```diff
@@ -0,0 +1,39 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: "system:auth-delegator"
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "consul.fullname" . }}-connect-injector-authmethod-svc-account
+    namespace: {{ .Release.Namespace }}
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
```
rboyer:
```suggestion
apiVersion: rbac.authorization.k8s.io/v1
```
I don't think the beta1 suffix is required anymore.

templates/connect-inject-authmethod-clusterrolebinding.yaml
adilyse: Turns out there was a mix in all of the rbac files. I've removed all of the beta tags across multiple files.

PR #156, opened 2019-05-01
Support the option to create an enterprise license acl token

adilyse:
When bootstraACLs is enabled, this adds an init container to the
enterprise license job that will wait until the appropriate acl
token is populated. Additionally, it marks it to run after the
acl init job so that everything is ready for it.

### DIFF
--- a/templates/enterprise-license.yaml
+++ b/templates/enterprise-license.yaml
@@ -14,7 +14,7 @@ metadata:
     release: {{ .Release.Name }}
   annotations:
     "helm.sh/hook": post-install
-    "helm.sh/hook-weight": "0"
+    "helm.sh/hook-weight": "100"
     "helm.sh/hook-delete-policy": hook-succeeded
 spec:
   template:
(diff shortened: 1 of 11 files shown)


templates/enterprise-license.yaml
```diff
@@ -30,21 +30,42 @@ spec:
...
+      - name: ent-license-acl-init
+        image: {{ .Values.global.imageK8S }}
+        command:
+          - "/bin/sh"
+          - "-ec"
+          - |
+            consul-k8s acl-init \
+              -secret-name="{{ .Release.Name }}-consul-enterprise-license-acl-token" \
+              -k8s-namespace={{ .Release.Namespace }} \
+              -init-type="sync"
```
rboyer: Should this be "sync" for the licence container?

templates/server-acl-init-job.yaml
```diff
@@ -55,7 +55,10 @@ spec:
                 -create-inject-token=true \
                 {{- end }}
                 {{- if .Values.connectInject.aclBindingRuleSelector }}
-                -acl-binding-rule-selector={{ .Values.connectInject.aclBindingRuleSelector }}
+                -acl-binding-rule-selector={{ .Values.connectInject.aclBindingRuleSelector }} \
+                {{- end }}
```
rboyer: Try this with a binding selector configured but no licence. The stray continuation worries me

tradel: @rboyer I think it ends up being okay because there's a blank line at the end of the command. Here's the output from `kubectl describe pod`:

```
    Command:
      /bin/sh
      -ec
      consul-k8s server-acl-init \
        -release-name=consul \
        -k8s-namespace=default \
        -expected-replicas=1 \
        -allow-dns=true \
        -create-inject-token=true \
        -acl-binding-rule-selector=serviceaccount.name!=default \
      
```

The job appears to run okay, but I agree, the extra backslash looks wrong.

templates/enterprise-license.yaml
adilyse: Yes, `sync` vs. `client`. I'd originally named it to after the catalog sync process, but ended up using it several places. It waits for an acl token to exist then creates it as an environment variable. So it waits for the token to sync? Kinda stretches the meaning, but I'm not worried enough about it to change it now.

templates/server-acl-init-job.yaml
adilyse: I haven't had any issue with it, but switching it up just in case.

test/unit/server-acl-init-job.bats
```diff
@@ -101,4 +101,41 @@ load _helpers
...
+@test "serverACLInit/Job: ent license acl option disabled missing server.enterpriseLicense.secretKey" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/server-acl-init-job.yaml  \
+      --set 'global.bootstrapACLs=true' \
+      --set 'server.enterpriseLicense.secretName=foo' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command | any(contains("-create-enterprise-license-token"))' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+>>>>>>> Support the option to create an enterprise license acl token
```
rboyer: is that from a merge conflict?

adilyse: Fixed.

PR #159, opened 2019-05-03
Support central configuration and proxy defaults

adilyse:
Added in Consul 1.5, proxy defaults allow users to define config
options that will be automatically applied to all injected sidecar
proxies.

Additionally, it is possible to enable Connect services to be centrally
registered.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -371,3 +371,20 @@ connectInject:
   # default disallows using the default service account for ACl generation.
   # Requires Consul v1.5+ and consul-k8s v0.8.0+.
   aclBindingRuleSelector: "serviceaccount.name!=default"
+
+  # Requires Consul v1.5+ and consul-k8s v0.8.0+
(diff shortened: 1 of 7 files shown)


values.yaml
```diff
@@ -371,3 +371,20 @@ connectInject:
   # default disallows using the default service account for ACl generation.
   # Requires Consul v1.5+ and consul-k8s v0.8.0+.
   aclBindingRuleSelector: "serviceaccount.name!=default"
+
+  # Requires Consul v1.5+ and consul-k8s v0.8.0+
+  centralConfig:
+    enabled: false
+
+    # defaultProtocol allows you to specify a convenience default protocol if
+    # all of your service are of the same protocol type. The individual annotation
```
```suggestion
    # all of your services are of the same protocol type. The individual annotation
```

adilyse: That's a good clarification, thanks!

PR #172, opened 2019-05-16
Make service account on ent license job optional

adilyse:
If ACLs are not enabled, the enterprise license application job does
not need any special privileges. This makes the definition of the
service account in that spec key off of the `bootstrapACLs` value
rather than being blanket applied.


eikenb (approved):
Don't know much about it, but everything I understand looks good. Never heard of yq before, had to look that one up.

PR #184, opened 2019-06-05
Update values.yaml

dwaiba:
image in values.yaml bumped to consul:1.5.1
https://github.com/hashicorp/consul/blob/v1.5.1/CHANGELOG.md


lkysow:
We actually need to bump to 1.6.0 now so I'm going to close this.

PR #191, opened 2019-06-07
Support running snapshot agents from the Helm chart

adilyse:
This allows users to run the Consul Enterprise Snapshot Agents feature
from the Helm chart. The snapshot agent uses the same permissions and
affinities as the Consul clients because they need to be co-located.

The configuration for the snapshot agents need to be defined in a
Kubernetes secret. This must not include the `http_addr` field as it
will conflict with the routing that is automatically set up to allow
the snapshot agent to talk to the client on the same k8s node. Likewise,
if ACLs are enabled using the `global.bootstrapACLs` option, the `token`
field cannot be included either, since it is automatically populated.

This also assumes that the snapshot agent retains its defaults of:
"service": "consul-snapshot",
"lock_key": "consul-snapshot/lock"

Additionally, any cloud storage defined must be routable from the
Kubernetes cluster where the snapshot agents are running.

### DIFF
--- a/templates/client-snapshot-agent-deployment.yaml
+++ b/templates/client-snapshot-agent-deployment.yaml
@@ -1,0 +1,116 @@
+{{- if (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}
+{{- if .Values.client.snapshotAgent.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "consul.fullname" . }}-snapshot-agent
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+spec:
+  replicas: {{ .Values.client.snapshotAgent.replicas }}
+  selector:
+    matchLabels:
+      app: {{ template "consul.name" . }}
+      chart: {{ template "consul.chart" . }}
+      release: {{ .Release.Name }}
+      component: client-snapshot-agent
+  template:
+    metadata:
+      labels:
+        app: {{ template "consul.name" . }}
+        chart: {{ template "consul.chart" . }}
+        release: {{ .Release.Name }}
+        component: client-snapshot-agent
+      annotations:
+        "consul.hashicorp.com/connect-inject": "false"
+    spec:
+      {{- if .Values.client.tolerations }}
+      tolerations:
+        {{ tpl .Values.client.tolerations . | nindent 8 | trim }}
+      {{- end }}
+      terminationGracePeriodSeconds: 10
+      serviceAccountName: {{ template "consul.fullname" . }}-client
+
+      {{- if .Values.client.priorityClassName }}
+      priorityClassName: {{ .Values.client.priorityClassName | quote }}
+      {{- end }}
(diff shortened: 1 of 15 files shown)


templates/client-snapshot-agent-deployment.yaml
```diff
@@ -0,0 +1,116 @@
...
+                  fieldPath: status.hostIP
+            {{- if .Values.global.bootstrapACLs }}
+            - name: CONSUL_HTTP_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: "{{ .Release.Name }}-consul-client-snapshot-agent-acl-token"
+                  key: "token"
+            {{- end}}
+          command:
+            - "/bin/sh"
```
lkysow: why `/bin/sh`?

templates/client-snapshot-agent-deployment.yaml
```diff
@@ -0,0 +1,116 @@
...
+            - "-ec"
+            - |
+              exec /bin/consul snapshot agent \
+                {{- if (and .Values.client.snapshotAgent.configSecret.secretName .Values.client.snapshotAgent.configSecret.secretKey) }}
+                -config-dir=/consul/config \
+                {{- end }}
+                {{- if .Values.global.bootstrapACLs}}
+                -config-dir=/consul/aclconfig \
+                {{- end }}
+                -http-addr=http://${HOST_IP}:8500
```
lkysow: ahh, is the `/bin/sh` for `HOST_IP`?

lkysow: although not sure it's necessary https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config

templates/client-snapshot-agent-deployment.yaml
adilyse: It's standard across all commands we execute through the Helm chart.

templates/client-snapshot-agent-deployment.yaml
```diff
@@ -0,0 +1,116 @@
...
+      component: client-snapshot-agent
+  template:
+    metadata:
+      labels:
+        app: {{ template "consul.name" . }}
+        chart: {{ template "consul.chart" . }}
+        release: {{ .Release.Name }}
+        component: client-snapshot-agent
+      annotations:
+        "consul.hashicorp.com/connect-inject": "false"
```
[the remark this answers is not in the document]
adilyse: Agree, but this is not the only place where it's missing. It would be good to take a systematic look at everything in the future and update it. Deferring for now.

values.yaml
```diff
@@ -223,6 +223,27 @@ client:
...
     # no_proxy: internal.domain.com
 
+  # snaphotAgent contains settings for setting up and running snapshot agents
+  # within the Consul clusters. They are required to be co-located with Consul
+  # clients, so will inherit the clients' nodeSelector, tolerations and affinity.
+  # This is an Enterprise feature only.
+  snapshotAgent:
+    enabled: false
+
+    # replicas determines how many snapshot agent pods are created
```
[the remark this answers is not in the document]
adilyse: Going to leave this as-is for now since we don't seem to have solid guidelines for folks about how many of these to run at the moment. Since this is an enterprise feature, folks using it should have access to someone that will help guide them to a correct number.

templates/client-snapshot-agent-deployment.yaml
```diff
@@ -0,0 +1,116 @@
...
+    spec:
+      {{- if .Values.client.tolerations }}
+      tolerations:
+        {{ tpl .Values.client.tolerations . | nindent 8 | trim }}
+      {{- end }}
+      terminationGracePeriodSeconds: 10
+      serviceAccountName: {{ template "consul.fullname" . }}-client
```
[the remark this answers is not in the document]
adilyse: Separated it out.

lkysow (approved):

PR #194, opened 2019-06-10
fix aliyun.com, we need use hostNetwork for client

dongtianyi:


Mikulas:
We had to enable `hostNetwork` as well. Unless you have all agents inside the cluster, Serf will fail because the multicast will be rejected with
> memberlist: Refuting a suspect message (from: ...)

```diff
diff --git a/templates/client-daemonset.yaml b/helm/charts/consul-helm/templates/client-daemonset.yaml
index 5282825..401b7ea 100644
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -66,6 +66,7 @@ spec:
           emptyDir: {}
         {{- end }}
 
+      hostNetwork: true
       containers:
         - name: consul
           image: "{{ default .Values.global.image .Values.client.image }}"
@@ -74,6 +75,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: status.podIP
+            - name: HOST_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
             - name: NAMESPACE
               valueFrom:
                 fieldRef:
@@ -98,7 +103,7 @@ spec:
 
               exec /bin/consul agent \
                 -node="${NODE}" \
-                -advertise="${POD_IP}" \
+                -advertise="${HOST_IP}" \
                 -bind=0.0.0.0 \
                 -client=0.0.0.0 \
                 {{- if .Values.client.grpc }}
```

sunadm:
No, you don't need it, `hostNetwork` is not the recommended way to expose the host port, but `hostPort` is.

To support `hostPort` in AliCloud ACK, you need to load an additional CNI networking plugins `portmap` when `terway` init, and add related configurations for port mapping.

sunadm:
#44

lkysow:
I think the issue for this user is that they have nodes outside of the kube cluster and since the agents are advertising their Pod IPs to those out of cluster nodes, they can't be gossiped with. We'd either have to run hostNetwork or map all the ports to the host and advertise the host ip.

sunadm:
@lkysow , I think not actually

In fact, not like other distributions, Aliyun/AliCloud provided ACK Kubernetes did not enabled the `portmap` CNI networking plugins by default, so `hostPort` not effect by default is 'reasonable'.

Just need to load and enable it.

lkysow:
> @lkysow , I think not actually
> 
> In fact, not like other distributions, Aliyun/AliCloud provided ACK Kubernetes did not enabled the `portmap` CNI networking plugins by default, so `hostPort` not effect by default is 'reasonable'.
> 
> Just need to load and enable it.

Does the CNI plugin give the host IP as the Pod IP?

kschoche:
This issue was resolved and merged in #496 and should be available in helm release 0.22, thank you!
To enable you would now add to your config : 
```
client:
  hostNetwork: true
  dnsPolicy: ClusterFirstWithHostNet
```

PR #203, opened 2019-07-02
Add new meshGateway key for Connect Mesh Gateways

lkysow:
- [x] consul-k8s changes merged
- [x] test with bootstrap acl
        - NOTE: this worked *but* the gateways fail unless there's a connect-enabled service running **and** for some reason the connect inject container kept getting `Error logging in: Unexpected response code: 403 (rpc error making call: rpc error making call: Permission denied)` errors...
- [x] figure out how best to specify wan-address
- [x] finish Service config
- [x] decide on default resources
- [x] liveness/readiness probes (should be able to use localhost tcp 443)
- [x] docs for all the options in values.yaml
- [x] deregistration on exit
- [x] add pod security policy
- [x] website docs
- [x] test with multi-dc
- [x] tests

### DIFF
--- a/templates/mesh-gateway-clusterrole.yaml
+++ b/templates/mesh-gateway-clusterrole.yaml
@@ -1,0 +1,34 @@
+{{- if .Values.meshGateway.enabled }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "consul.fullname" . }}-mesh-gateway
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: mesh-gateway
+{{- if or .Values.global.bootstrapACLs .Values.global.enablePodSecurityPolicies }}
+rules:
+{{- if .Values.global.enablePodSecurityPolicies }}
+  - apiGroups: ["policy"]
+    resources: ["podsecuritypolicies"]
+    resourceNames:
+      - {{ template "consul.fullname" . }}-mesh-gateway
+    verbs:
+      - use
+{{- end }}
+{{- if .Values.global.bootstrapACLs }}
+  - apiGroups: [""]
+    resources:
+      - secrets
+    resourceNames:
+      - {{ .Release.Name }}-consul-mesh-gateway-acl-token
+    verbs:
+      - get
+{{- end }}
+{{- else }}
+rules: []
+{{- end }}
+{{- end }}
--- a/templates/mesh-gateway-clusterrolebinding.yaml
+++ b/templates/mesh-gateway-clusterrolebinding.yaml
@@ -1,0 +1,20 @@
+{{- if .Values.meshGateway.enabled }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ template "consul.fullname" . }}-mesh-gateway
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: mesh-gateway
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "consul.fullname" . }}-mesh-gateway
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "consul.fullname" . }}-mesh-gateway
+    namespace: {{ .Release.Namespace }}
+{{- end }}
--- a/templates/mesh-gateway-deployment.yaml
+++ b/templates/mesh-gateway-deployment.yaml
@@ -1,0 +1,172 @@
+{{- if .Values.meshGateway.enabled }}
+{{- if not .Values.connectInject.enabled }}{{ fail "connectInject.enabled must be true" }}{{ end -}}
+{{- if not .Values.client.grpc }}{{ fail "client.grpc must be true" }}{{ end -}}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "consul.fullname" . }}-mesh-gateway
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+    component: mesh-gateway
+  {{- if .Values.meshGateway.annotations }}
+  annotations:
+    {{- tpl .Values.meshGateway.annotations . | nindent 4 }}
+  {{- end }}
+spec:
+  replicas: {{ .Values.meshGateway.replicas }}
+  selector:
+    matchLabels:
+      app: {{ template "consul.name" . }}
+      chart: {{ template "consul.chart" . }}
+      release: {{ .Release.Name }}
+      component: mesh-gateway
+  template:
+    metadata:
+      labels:
+        app: {{ template "consul.name" . }}
+        chart: {{ template "consul.chart" . }}
+        release: {{ .Release.Name }}
+        component: mesh-gateway
+      annotations:
+        "consul.hashicorp.com/connect-inject": "false"
+    spec:
+      {{- if .Values.meshGateway.affinity }}
+      affinity:
+        {{ tpl .Values.meshGateway.affinity . | nindent 8 | trim }}
+      {{- end }}
+      {{- if .Values.meshGateway.tolerations }}
+      tolerations:
+        {{ tpl .Values.meshGateway.tolerations . | nindent 8 | trim }}
+      {{- end }}
+      terminationGracePeriodSeconds: 10
+      serviceAccountName: {{ template "consul.fullname" . }}-mesh-gateway
+      volumes:
+        - name: consul-bin
+          emptyDir: {}
+      {{- if .Values.meshGateway.hostNetwork }}
+      hostNetwork: {{ .Values.meshGateway.hostNetwork }}
+      {{- end }}
+      {{- if .Values.meshGateway.dnsPolicy }}
+      dnsPolicy: {{ .Values.meshGateway.dnsPolicy }}
+      {{- end }}
+      initContainers:
+        # We use the Envoy image as our base image so we use an init container to
+        # copy the Consul binary to a shared directory that can be used when
+        # starting Envoy.
+        - name: copy-consul-bin
+          image: {{ .Values.global.image | quote }}
+          command:
+          - cp
+          - /bin/consul
+          - /consul-bin/consul
+          volumeMounts:
+          - name: consul-bin
+            mountPath: /consul-bin
+        {{- if .Values.global.bootstrapACLs }}
+        # Wait for secret containing acl token to be ready.
+        # Doesn't do anything with it but when the main container starts we
+        # know that it's been created.
+        - name: mesh-gateway-acl-init
+          image: {{ .Values.global.imageK8S }}
+          command:
+            - "/bin/sh"
+            - "-ec"
+            - |
+              consul-k8s acl-init \
+                -secret-name="{{ .Release.Name }}-consul-mesh-gateway-acl-token" \
+                -k8s-namespace={{ .Release.Namespace }} \
+                -init-type="sync"
+        {{- end }}
+      containers:
+        - name: mesh-gateway
+          image: {{ .Values.meshGateway.imageEnvoy | quote }}
+          {{- if .Values.meshGateway.resources }}
+          resources:
+            {{ tpl .Values.meshGateway.resources . | nindent 12 | trim }}
+          {{- end }}
+          volumeMounts:
+          - name: consul-bin
+            mountPath: /consul-bin
+          env:
+            - name: HOST_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+            - name: NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+            {{- if .Values.global.bootstrapACLs }}
+            - name: CONSUL_HTTP_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: "{{ .Release.Name }}-consul-mesh-gateway-acl-token"
+                  key: "token"
+            {{- end}}
+          command:
+            # /bin/sh -c is needed so we can use the pod-specific environment
+            # variables.
+            - "/bin/sh"
+            - "-ec"
+            - |
+              exec /consul-bin/consul connect envoy \
+                -mesh-gateway \
+                -register \
+                -address="${POD_IP}:{{ .Values.meshGateway.containerPort }}" \
+                -http-addr="${HOST_IP}:8500" \
+                -grpc-addr="${HOST_IP}:8502" \
+                {{- if .Values.meshGateway.wanAddress.host }}
+                -wan-address="{{ .Values.meshGateway.wanAddress.host }}:{{ .Values.meshGateway.wanAddress.port }}" \
+                {{- else if .Values.meshGateway.wanAddress.useNodeName }}
+                -wan-address="${NODE_NAME}:{{ .Values.meshGateway.wanAddress.port }}" \
+                {{- else if .Values.meshGateway.wanAddress.useNodeIP }}
+                -wan-address="${HOST_IP}:{{ .Values.meshGateway.wanAddress.port }}" \
+                {{- end }}
+                {{- if .Values.meshGateway.consulServiceName }}
+                {{- if .Values.global.bootstrapACLs }}{{ fail "if global.bootstrapACLs is true, meshGateway.consulServiceName cannot be set" }}{{ end }}
+                -service={{ .Values.meshGateway.consulServiceName | quote }} \
+                {{- end }}
+          {{- if .Values.meshGateway.enableHealthChecks }}
+          livenessProbe:
+            tcpSocket:
+              port: {{ .Values.meshGateway.containerPort }}
+            failureThreshold: 3
+            initialDelaySeconds: 30
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 5
+          readinessProbe:
+            tcpSocket:
+              port: {{ .Values.meshGateway.containerPort }}
+            failureThreshold: 3
+            initialDelaySeconds: 10
+            periodSeconds: 10
+            successThreshold: 1
+            timeoutSeconds: 5
+          {{- end }}
+          ports:
+            - name: gateway
+              containerPort: {{ .Values.meshGateway.containerPort }}
+              {{- if .Values.meshGateway.hostPort }}
+              hostPort:  {{ .Values.meshGateway.hostPort }}
+              {{- end }}
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh", "-ec", "/consul-bin/consul services deregister -http-addr=\"${HOST_IP}:8500\" -id=\"{{ default "mesh-gateway" .Values.meshGateway.consulServiceName }}\""]
+      {{- if .Values.meshGateway.priorityClassName }}
+      priorityClassName: {{ .Values.meshGateway.priorityClassName | quote }}
+      {{- end }}
+      {{- if .Values.meshGateway.nodeSelector }}
+      nodeSelector:
+        {{ tpl .Values.meshGateway.nodeSelector . | indent 8 | trim }}
+      {{- end }}
+{{- end }}
--- a/templates/mesh-gateway-deployment.yaml
+++ b/templates/mesh-gateway-deployment.yaml
@@ -1,6 +1,8 @@
 {{- if .Values.meshGateway.enabled }}
 {{- if not .Values.connectInject.enabled }}{{ fail "connectInject.enabled must be true" }}{{ end -}}
 {{- if not .Values.client.grpc }}{{ fail "client.grpc must be true" }}{{ end -}}
+{{- /* The below test checks if clients are disabled (and if so, fails). We use the conditional from other client files and prepend 'not' */ -}}
+{{- if not (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}{{ fail "clients must be enabled" }}{{ end -}}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -12,10 +14,6 @@ metadata:
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
     component: mesh-gateway
-  {{- if .Values.meshGateway.annotations }}
-  annotations:
-    {{- tpl .Values.meshGateway.annotations . | nindent 4 }}
-  {{- end }}
 spec:
   replicas: {{ .Values.meshGateway.replicas }}
   selector:
@@ -33,6 +31,9 @@ spec:
         component: mesh-gateway
       annotations:
         "consul.hashicorp.com/connect-inject": "false"
+        {{- if .Values.meshGateway.annotations }}
+        {{- tpl .Values.meshGateway.annotations . | nindent 8 }}
+        {{- end }}
     spec:
       {{- if .Values.meshGateway.affinity }}
       affinity:
@@ -100,10 +101,12 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: status.podIP
+            {{- if .Values.meshGateway.wanAddress.useNodeName }}
             - name: NODE_NAME
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
+            {{- end }}
             {{- if .Values.global.bootstrapACLs }}
             - name: CONSUL_HTTP_TOKEN
               valueFrom:
@@ -130,8 +133,8 @@ spec:
                 {{- else if .Values.meshGateway.wanAddress.useNodeIP }}
                 -wan-address="${HOST_IP}:{{ .Values.meshGateway.wanAddress.port }}" \
                 {{- end }}
-                {{- if .Values.meshGateway.consulServiceName }}
-                {{- if .Values.global.bootstrapACLs }}{{ fail "if global.bootstrapACLs is true, meshGateway.consulServiceName cannot be set" }}{{ end }}
+                {{- if and .Values.meshGateway.consulServiceName }}
+                {{- if and .Values.global.bootstrapACLs (ne .Values.meshGateway.consulServiceName "mesh-gateway") }}{{ fail "if global.bootstrapACLs is true, meshGateway.consulServiceName cannot be set" }}{{ end }}
                 -service={{ .Values.meshGateway.consulServiceName | quote }} \
                 {{- end }}
           {{- if .Values.meshGateway.enableHealthChecks }}
--- a/test/unit/mesh-gateway-clusterrole.bats
+++ b/test/unit/mesh-gateway-clusterrole.bats
@@ -1,0 +1,77 @@
+#!/usr/bin/env bats
+
+load _helpers
+
+@test "meshGateway/ClusterRole: disabled by default" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/mesh-gateway-clusterrole.yaml  \
+      . | tee /dev/stderr |
+      yq 'length > 0' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+}
+
+@test "meshGateway/ClusterRole: enabled with meshGateway.enabled=true" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/mesh-gateway-clusterrole.yaml  \
+      --set 'meshGateway.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      --set 'client.grpc=true' \
+      . | tee /dev/stderr |
+      yq 'length > 0' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
+@test "meshGateway/ClusterRole: rules for PodSecurityPolicy" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/mesh-gateway-clusterrole.yaml  \
+      --set 'meshGateway.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      --set 'client.grpc=true' \
+      --set 'global.enablePodSecurityPolicies=true' \
+      . | tee /dev/stderr |
+      yq -r '.rules[0].resources[0]' | tee /dev/stderr)
+  [ "${actual}" = "podsecuritypolicies" ]
+}
+
+
+@test "meshGateway/ClusterRole: rules for global.bootstrapACLs=true" {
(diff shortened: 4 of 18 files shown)


templates/client-daemonset.yaml
```diff
@@ -101,7 +101,7 @@ spec:
                 -advertise="${POD_IP}" \
                 -bind=0.0.0.0 \
                 -client=0.0.0.0 \
-                {{- if .Values.client.grpc }}
+                {{- if or .Values.client.grpc .Values.meshGateway.enabled }}
```
lkysow: gateways require gRPC. I can either do this or require `client.grpc` to be set to true. Requiring them to set this in client.grpc is cleaner because then they don't get grpc unexpectedly. But then we should find a way to require that if meshGateway.enabled is true, that client.grpc is true.

values.yaml
```diff
@@ -167,7 +167,7 @@ client:
   join: null
 
   # grpc should be set to true if the gRPC listener should be enabled.
-  # This should be set to true if connectInject is enabled.
+  # This will be set to true if connectInject or meshGateway is enabled.
```
```suggestion
  # This will automatically be set to true if connectInject or meshGateway is enabled.
```

This statement isn't entirely true unless we fix it so that connectInject turns this on as well (mentioned in the comment on the client daemonset file).

templates/mesh-gateway-deployment.yaml
```diff
@@ -0,0 +1,162 @@
+{{- if and .Values.meshGateway.enabled }}
```
adilyse:
```suggestion
{{- if .Values.meshGateway.enabled }}
```

templates/mesh-gateway-clusterrolebinding.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if and .Values.meshGateway.enabled }}
```
adilyse:
```suggestion
{{- if .Values.meshGateway.enabled }}
```

values.yaml
```diff
@@ -387,3 +387,116 @@ connectInject:
...
+
+  # The service key configures the Service that fronts the Gateway Deployment.
+  service:
+    # Whether to create a Service or not.
+    enabled: false
+    type: LoadBalancer
+    port: 443
+
+    # Optional YAML string for additional annotations.
+    annotations: |
```
adilyse: This and `additionalSpec` should be null, though it would be good to have an example of formatting in the comment above it. Elsewhere in the Helm chart we've set the standard that none of the options are commented out, but the ones not in use are `null` or empty. It would be good to be consistent.

values.yaml
```diff
@@ -387,3 +387,116 @@ connectInject:
...
+    # Optional YAML string that will be appended to the Service spec.
+    additionalSpec: |
+
+  # Envoy image to use.
+  imageEnvoy: envoyproxy/envoy:v1.8.0
```
adilyse: Per other discussion, we should be using envoy 1.10, or at least 1.9.1+

values.yaml
```diff
@@ -387,3 +387,116 @@ connectInject:
...
+
+    # If true, each Gateway Pod will advertise its NodeIP
+    # (as provided by the Kubernetes downward API) as the wan address.
+    # This is useful if the node IPs are routable from other DCs.
+    # The port key will be used as the port for the wan address.
+    useNodeIP: true
+
+    # If true, each Gateway Pod will advertise its NodeName
+    # (as provided by the Kubernetes downward API) as the wan address.
+    # Again, this is useful if the node names are DNS entries that are
```
```suggestion
    # This is useful if the node names are DNS entries that are
```

values.yaml
```diff
@@ -387,3 +387,116 @@ connectInject:
...
+    # routable from other DCs.
+    # The port key will be used as the port for the wan address.
```
adilyse: Same comment about `key`.

values.yaml
```diff
@@ -387,3 +387,116 @@ connectInject:
...
+    # Must set useNodeIP to false.
+    # useNodeName: false
+
+    # If set, this will be the host of the wan address for all gateways.
+    # Users must ensure that this address routes to the Gateway pods,
+    # for example via a DNS
+    # entry that routes to the Service fronting the Deployment.
+    # The port key will be used as the port for the wan address.
```
adilyse: Same comment about `key`.

values.yaml
```diff
@@ -387,3 +387,116 @@ connectInject:
...
+    # (as provided by the Kubernetes downward API) as the wan address.
+    # Again, this is useful if the node names are DNS entries that are
+    # routable from other DCs.
+    # The port key will be used as the port for the wan address.
+    # Must set useNodeIP to false.
+    # useNodeName: false
+
+    # If set, this will be the host of the wan address for all gateways.
+    # Users must ensure that this address routes to the Gateway pods,
+    # for example via a DNS
```
adilyse: Some odd line wrapping in this comment.

values.yaml
```diff
@@ -387,3 +387,116 @@ connectInject:
...
+
+    # If true, each Gateway Pod will advertise its NodeName
+    # (as provided by the Kubernetes downward API) as the wan address.
+    # Again, this is useful if the node names are DNS entries that are
+    # routable from other DCs.
+    # The port key will be used as the port for the wan address.
+    # Must set useNodeIP to false.
+    # useNodeName: false
+
+    # If set, this will be the host of the wan address for all gateways.
```
adilyse: I'm having a hard time parsing "will be the host of the wan address for all gateways." Is there another way to phrase this to make it clearer?

values.yaml
```diff
@@ -387,3 +387,116 @@ connectInject:
...
+
+  # What gets registered as wan address for the gateway.
+  wanAddress:
+    # Port that gets registered.
+    port: 443
+
+    # If true, each Gateway Pod will advertise its NodeIP
+    # (as provided by the Kubernetes downward API) as the wan address.
+    # This is useful if the node IPs are routable from other DCs.
+    # The port key will be used as the port for the wan address.
```
adilyse: What do you mean by `port key`? Something like this? 

```suggestion
    # The port configured above will be used as the port for the wan address.
```

We use `key` elsewhere in this file in several different ways, so overloading it here as well will get confusing.

values.yaml
```diff
@@ -387,3 +387,116 @@ connectInject:
...
+
+    # If set, this will be the host of the wan address for all gateways.
+    # Users must ensure that this address routes to the Gateway pods,
+    # for example via a DNS
+    # entry that routes to the Service fronting the Deployment.
+    # The port key will be used as the port for the wan address.
+    # Must set useNodeIP and useNodeName to false.
+    # host: myhostname.com
+
+  # The service key configures the Service that fronts the Gateway Deployment.
```
adilyse: Same comment about `key`.

values.yaml
```diff
@@ -387,3 +387,116 @@ connectInject:
...
+  imageEnvoy: envoyproxy/envoy:v1.8.0
+
+  # If set to true, gateway Pods will run on the host network.
+  hostNetwork: false
+
+  # dnsPolicy to use.
+  dnsPolicy: ClusterFirst
+
+  # Override the default 'mesh-gateway' service name registered in Consul.
+  # Cannot be used if bootstrapACLs is true since the ACL token generated
```
adilyse: There should definitely be a test for this, since it would be good to have a hard requirement on this in the templating.

values.yaml
```diff
@@ -387,3 +387,116 @@ connectInject:
...
+  # is only for the name 'mesh-gateway'.
+  # consulServiceName: your-name
+
+  # Port that the gateway will run on inside the container.
+  containerPort: 443
+
+  # Optional hostPort for the gateway to be exposed on.
+  # This can be used with wanAddress.port and wanAddress.useNodeIP
+  # to expose the gateways directly from the node.
+  # hostPort: null
```
adilyse: Does this potentially conflict with the `hostNetwork` option?

templates/mesh-gateway-deployment.yaml
```diff
@@ -0,0 +1,162 @@
...
+      tolerations:
+        {{ tpl .Values.meshGateway.tolerations . | nindent 8 | trim }}
+      {{- end }}
+      terminationGracePeriodSeconds: 10
+      serviceAccountName: {{ template "consul.fullname" . }}-mesh-gateway
+      volumes:
+      - name: consul-bin
+        emptyDir: {}
+      hostNetwork: {{ .Values.meshGateway.hostNetwork }}
+      dnsPolicy: {{ .Values.meshGateway.dnsPolicy }}
```
adilyse: Should this be option in case someone sets the `dnsPolicy` to null?

templates/mesh-gateway-deployment.yaml
```diff
@@ -0,0 +1,162 @@
...
+        env:
+        - name: HOST_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: status.hostIP
+        - name: POD_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: status.podIP
+        - name: NODE_NAME
```
adilyse: This should be optional, only if `.Values.meshGateway.wanAddress.useNodeName` is true.

lkysow:
> Big open question from me: Does the mesh gateway need to have Connect and/or central config enabled to work correctly? If so, there are some places that need slightly different logic.
Yes both connect and central config are needed. Have you commented on all the places that need slightly different logic or does that warrant further discussion?


> Also: Do the gateways need to run on a node with a client available? If so, it would be good to note that somewhere in the affinity documentation so that folks don't accidentally partition things in an unusable way.
Yes they do, good point.

templates/client-daemonset.yaml
lkysow: Okay, I'll do that in a separate PR.

values.yaml
```diff
@@ -387,3 +387,116 @@ connectInject:
...
+  enabled: false
+
+  # Which mode the gateway should run in. Can be either 'remote' or 'local'.
+  # If set to remote, Connect traffic to services in other datacenters will
+  # be sent directly to the other datacenter's gateway.
+  # If set to local, Connect traffic to services in other datacenters will
+  # be proxied via the local gateway. Use 'local' if services cannot directly
+  # connect to the other datacenter's gateway and so must proxy through the local
+  # gateway, i.e. if only the gateways can communicate between datacenters.
+  mode: remote
```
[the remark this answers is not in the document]
lkysow: ✅

values.yaml
lkysow: ✅

templates/mesh-gateway-clusterrole.yaml
```diff
@@ -0,0 +1,24 @@
...
+rules:
+- apiGroups: [""]
+  resources:
+    - secrets
+  resourceNames:
+    - {{ .Release.Name }}-consul-mesh-gateway-acl-token
+  verbs:
+    - get
+{{- else }}
+rules: []
```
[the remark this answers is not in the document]
lkysow: ✅

templates/mesh-gateway-deployment.yaml
lkysow: ✅

templates/mesh-gateway-deployment.yaml
lkysow: Is there a downside to having it here regardless? My thought was that it's less conditionals == less complexity.

values.yaml
lkysow: ✅

values.yaml
lkysow: Changed to: "# If set, each gateway Pod will use this host as its wan address."

values.yaml
lkysow: ✅

values.yaml
lkysow: ✅

values.yaml
lkysow: ✅

values.yaml
lkysow: ✅

values.yaml
lkysow: How can I do a hard requirement?

values.yaml
lkysow: A bit, they have to set it to the containerPort if using `hostNetwork: true`. I'll add that to the docs.

Kube docs:
> If HostNetwork is specified, this must match ContainerPort.

templates/mesh-gateway-deployment.yaml
```diff
@@ -0,0 +1,169 @@
...
+          env:
+            - name: HOST_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+            - name: NODE_NAME
```
adilyse: We shouldn't add unnecessary things to a deployment if it's not needed, even if it adds a bit of complexity. Tests are a good way to make sure the logic does the right thing.

test/unit/mesh-gateway-deployment.bats
```diff
@@ -0,0 +1,220 @@
...
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/mesh-gateway-deployment.yaml  \
+      --set 'meshGateway.enabled=true' \
+      . | tee /dev/stderr |
+      yq 'length > 0' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
+@test "meshGateway/Deployment: annotations can be set" {
```
adilyse: Might be good to make sure that setting more than one annotation will be added and formatted correctly.

values.yaml
```diff
@@ -387,3 +389,130 @@ connectInject:
     # configured proxy.
     proxyDefaults: |
       {}
+
+# Mesh Gateways enable Consul Connect to work across Consul datacenters.
+meshGateway:
+  # If mesh gateways are enabled, a Deployment will be created that runs
+  # gateways and Consul Connect will be configured to use gateways.
+  # See {todo doc link} for more information.
```
adilyse: Reminder that this is a todo link.

values.yaml
adilyse: Where the `consulServiceName` is used and `bootstrapACLs` is true, you can require that the service name is equal to `mesh-gateway` or that section is not rendered.

values.yaml
```diff
@@ -387,3 +389,130 @@ connectInject:
...
+    useNodeName: false
+
+    # If set, each gateway Pod will use this host as its wan address.
+    # Users must ensure that this address routes to the Gateway pods,
+    # for example via a DNS entry that routes to the Service fronting the Deployment.
+    # meshGateway.wanAddress.port will be used as the port for the wan address.
+    # useNodeIP and useNodeName must be false.
+    host: ""
+
+  # The service key configures the Service that fronts the Gateway Deployment.
```
```suggestion
  # The service option configures the Service that fronts the Gateway Deployment.
```

templates/client-daemonset.yaml
lkysow: I'm actually going to remove this because it feels wrong to set grpc to true automatically. Instead I'll document that it must be set to true for gateways to work *and* I've found a neat trick to have helm fail if these things aren't set properly. I'll tag you on it when it's committed.

values.yaml
lkysow: I'm going to do this via a `fail` call. I'll tag you on it, lmk what you think.

test/unit/mesh-gateway-deployment.bats
lkysow: We can't set multiline strings via `--set` unfortunately. I could write a `values.yaml` file to disk I guess? Might be overkill, what do you think?

values.yaml
lkysow: ✅

templates/mesh-gateway-deployment.yaml
lkysow: ✅

templates/mesh-gateway-deployment.yaml
```diff
@@ -0,0 +1,172 @@
+{{- if .Values.meshGateway.enabled }}
+{{- if not .Values.connectInject.enabled }}{{ fail "connectInject.enabled must be true" }}{{ end -}}
```
lkysow: @adilyse this is what I'm doing to enforce requirements. What do you think?

templates/mesh-gateway-deployment.yaml
```diff
@@ -0,0 +1,172 @@
...
+                -grpc-addr="${HOST_IP}:8502" \
+                {{- if .Values.meshGateway.wanAddress.host }}
+                -wan-address="{{ .Values.meshGateway.wanAddress.host }}:{{ .Values.meshGateway.wanAddress.port }}" \
+                {{- else if .Values.meshGateway.wanAddress.useNodeName }}
+                -wan-address="${NODE_NAME}:{{ .Values.meshGateway.wanAddress.port }}" \
+                {{- else if .Values.meshGateway.wanAddress.useNodeIP }}
+                -wan-address="${HOST_IP}:{{ .Values.meshGateway.wanAddress.port }}" \
+                {{- end }}
+                {{- if .Values.meshGateway.consulServiceName }}
+                {{- if .Values.global.bootstrapACLs }}{{ fail "if global.bootstrapACLs is true, meshGateway.consulServiceName cannot be set" }}{{ end }}
```
lkysow: also here

values.yaml
lkysow: ✅

test/unit/mesh-gateway-deployment.bats
```diff
@@ -0,0 +1,505 @@
...
+      --set 'meshGateway.wanAddress.useNodeIP=false' \
+      --set 'meshGateway.wanAddress.useNodeName=false' \
+      --set 'meshGateway.wanAddress.host=myhost' \
+      --set 'meshGateway.wanAddress.port=4444' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.containers[0].command[2]' | tee /dev/stderr)
+  [[ "${actual}" =~ '-wan-address="myhost:4444"' ]]
+}
+
+@test "meshGateway/Deployment: fails if consulServiceName is set and bootstrapACLs is true" {
```
adilyse: As mentioned elsewhere, it would be nice for this to work if someone happened to set the service name to `mesh-gateway`. A test for that case would be good.

test/unit/mesh-gateway-clusterrolebinding.bats
```diff
@@ -0,0 +1,37 @@
...
+      -x templates/mesh-gateway-clusterrolebinding.yaml  \
+      --set 'meshGateway.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      --set 'client.grpc=true' \
+      . | tee /dev/stderr |
+      yq 'length > 0' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
+@test "meshGateway/ClusterRoleBinding: subject name is correct" {
```
adilyse: This test fails locally for me:

```
`[ "${actual}" = "release-name-consul-mesh-gateway" ]' failed

RELEASE-NAME-consul-mesh-gateway
```

test/unit/mesh-gateway-podsecuritypolicy.bats
```diff
@@ -0,0 +1,25 @@
...
+  local actual=$(helm template \
+      -x templates/mesh-gateway-podsecuritypolicy.yaml  \
+      --set 'meshGateway.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      --set 'client.grpc=true' \
+      --set 'global.enablePodSecurityPolicies=true' \
+      . | tee /dev/stderr |
+      yq 'length > 0' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
```
adilyse: Again, it would be good to make sure this is disabled if connectInject and client.grpc are not enabled.

test/unit/mesh-gateway-podsecuritypolicy.bats
```diff
@@ -0,0 +1,25 @@
...
+@test "meshGateway/PodSecurityPolicy: disabled by default" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/mesh-gateway-podsecuritypolicy.yaml  \
+      . | tee /dev/stderr |
+      yq 'length > 0' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+}
+
+@test "meshGateway/PodSecurityPolicy: enabled with meshGateway.enabled and global.enablePodSecurityPolicies=true" {
```
adilyse:
```suggestion
@test "meshGateway/PodSecurityPolicy: enabled with meshGateway, connectInject and client.grpc enabled and global.enablePodSecurityPolicies=true" {
```

test/unit/mesh-gateway-service.bats
```diff
@@ -0,0 +1,186 @@
...
+@test "meshGateway/Service: disabled by default" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/mesh-gateway-service.yaml  \
+      . | tee /dev/stderr |
+      yq 'length > 0' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+}
+
+@test "meshGateway/Service: disabled with meshGateway.enabled=true" {
```
adilyse:
```suggestion
@test "meshGateway/Service: disabled by default with meshGateway, connectInject and client.grpc enabled" {
```

test/unit/mesh-gateway-serviceaccount.bats
```diff
@@ -0,0 +1,25 @@
...
+@test "meshGateway/ServiceAccount: disabled by default" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/mesh-gateway-serviceaccount.yaml  \
+      . | tee /dev/stderr |
+      yq 'length > 0' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+}
+
+@test "meshGateway/ServiceAccount: enabled with meshGateway.enabled = true" {
```
adilyse:
```suggestion
@test "meshGateway/ServiceAccount: enabled with meshGateway, connectInject and client.grpc enabled" {
```

test/unit/mesh-gateway-serviceaccount.bats
```diff
@@ -0,0 +1,25 @@
...
+  local actual=$(helm template \
+      -x templates/mesh-gateway-serviceaccount.yaml  \
+      --set 'meshGateway.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      --set 'client.grpc=true' \
+      . | tee /dev/stderr |
+      yq 'length > 0' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
```
adilyse: Should test that this is disabled with connectInject and/or client.grpc turned off.

test/unit/server-configmap.bats
```diff
@@ -111,3 +111,48 @@ load _helpers
...
+
+@test "server/ConfigMap: proxyDefaults should have no gateway mode if disabled" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/server-config-configmap.yaml  \
+      --set 'connectInject.enabled=true' \
+      --set 'connectInject.centralConfig.enabled=true' \
+      --set 'connectInject.centralConfig.proxyDefaults="{\"hello\": \"world\"}"' \
+      --set 'meshGateway.enabled=true' \
+      --set 'meshGateway.globalMode=' \
```
adilyse: Should this be set to `null`? Or `local`?

test/unit/mesh-gateway-clusterrole.bats
```diff
@@ -0,0 +1,77 @@
...
+  local actual=$(helm template \
+      -x templates/mesh-gateway-clusterrole.yaml  \
+      --set 'meshGateway.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      --set 'client.grpc=true' \
+      . | tee /dev/stderr |
+      yq 'length > 0' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
```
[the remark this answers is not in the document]
lkysow: The whole chart will fail to render if that's not the case. That's tested in the deployment bats tests.

test/unit/mesh-gateway-serviceaccount.bats
lkysow: The whole chart will fail in that case

test/unit/mesh-gateway-podsecuritypolicy.bats
lkysow: The whole chart will fail in that case

templates/mesh-gateway-deployment.yaml
lkysow: Yeah it's not great but not sure what else to do. I think this is better than not rendering the template since they'll get an error and can fix it rather than wondering why there's no gateways spinning up

templates/mesh-gateway-deployment.yaml
lkysow: yep, I thought I added that in there but must have got deleted somehow

test/unit/mesh-gateway-clusterrolebinding.bats
lkysow: Maybe a helm version difference?

```
Client: &version.Version{SemVer:"v2.14.1", GitCommit:"5270352a09c7e8b6e8c9593002a73535276507c0", GitTreeState:"clean"}
	Server: &version.Version{SemVer:"v2.14.1", GitCommit:"5270352a09c7e8b6e8c9593002a73535276507c0", GitTreeState:"clean"}
```

test/unit/server-configmap.bats
lkysow: If set to null or empty string then it shouldn't be set. Added a test for null.

values.yaml
```diff
@@ -408,3 +409,134 @@ connectInject:
...
+
+  # Port that the gateway will run on inside the container.
+  containerPort: 443
+
+  # Optional hostPort for the gateway to be exposed on.
+  # This can be used with wanAddress.port and wanAddress.useNodeIP
+  # to expose the gateways directly from the node.
+  # If hostNetwork is true, this must be null or set to the same port as
+  # containerPort.
+  hostPort: null
```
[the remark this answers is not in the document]
lkysow: ✅

values.yaml
```diff
@@ -408,3 +409,134 @@ connectInject:
...
+  # gateways and Consul Connect will be configured to use gateways.
+  # See https://www.consul.io/docs/connect/mesh_gateway.html
+  # Requirements: consul >= 1.6.0 and consul-k8s >= 0.8.2 if using global.bootstrapACLs.
+  enabled: false
+
+  # Globally configure which mode the gateway should run in.
+  # Can be set to either "remote", "local", "none" or "".
+  # See https://consul.io/docs/connect/mesh_gateway.html#modes-of-operation for
+  # a description of each mode.
+  # If set, connectInject.centralConfig.enabled should be set to true so that
```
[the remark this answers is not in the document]
lkysow: Made clearer.

values.yaml
```diff
@@ -408,3 +409,134 @@ connectInject:
     # configured proxy.
     proxyDefaults: |
       {}
+
+# Mesh Gateways enable Consul Connect to work across Consul datacenters.
+meshGateway:
+  # If mesh gateways are enabled, a Deployment will be created that runs
+  # gateways and Consul Connect will be configured to use gateways.
+  # See https://www.consul.io/docs/connect/mesh_gateway.html
+  # Requirements: consul >= 1.6.0 and consul-k8s >= 0.8.2 if using global.bootstrapACLs.
```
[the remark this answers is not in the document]
lkysow: ✅

test/unit/mesh-gateway-clusterrolebinding.bats
```diff
@@ -0,0 +1,37 @@
...
+  local actual=$(helm template \
+      -x templates/mesh-gateway-clusterrolebinding.yaml  \
+      --set 'meshGateway.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      --set 'client.grpc=true' \
+      . | tee /dev/stderr |
+      yq 'length > 0' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
```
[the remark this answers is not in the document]
lkysow: the whole chart will fail to render in this case, it's tested in the deployment

test/unit/mesh-gateway-clusterrolebinding.bats
lkysow: I've set it explicitly via `--name`

test/unit/mesh-gateway-clusterrole.bats
```diff
@@ -0,0 +1,77 @@
...
+      -x templates/mesh-gateway-clusterrole.yaml  \
+      --set 'meshGateway.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      --set 'client.grpc=true' \
+      --set 'global.enablePodSecurityPolicies=true' \
+      . | tee /dev/stderr |
+      yq -r '.rules[0].resources[0]' | tee /dev/stderr)
+  [ "${actual}" = "podsecuritypolicies" ]
+}
+
```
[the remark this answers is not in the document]
lkysow: ✅

test/unit/mesh-gateway-deployment.bats
```diff
@@ -0,0 +1,505 @@
...
+}
+
+@test "meshGateway/Deployment: fails if client.grpc=false" {
+  cd `chart_dir`
+  run helm template \
+      -x templates/mesh-gateway-deployment.yaml  \
+      --set 'meshGateway.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      --set 'client.grpc=false' \
+      --set 'connectInject.enabled=true' \
```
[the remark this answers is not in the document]
lkysow: ✅

test/unit/mesh-gateway-deployment.bats
```diff
@@ -0,0 +1,505 @@
...
+      --set 'meshGateway.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      --set 'client.grpc=true' \
+      --set 'meshGateway.tolerations=- key: value' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.tolerations[0].key' | tee /dev/stderr)
+  [ "${actual}" = "value" ]
+}
+
+@test "meshGateway/Deployment: hostNetwork can be set" {
```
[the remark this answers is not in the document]
lkysow: ✅

test/unit/mesh-gateway-deployment.bats
lkysow: ✅

lkysow:
Updates:
- added `fail` if clients are disabled, tested
- moved annotations to Pod from Deployment, tested multiple annotations
- added conditional render for NODE_NAME
- added special case for setting `consulServiceName` to `mesh-gateway`, tested
- added headings for test cases
- added test for `hostNetwork` not set by default
- added test for `globalMode=null`


> Realized that there's yet another dependency to keep track off-- the logic also needs to enforce that clients are enabled for mesh gateways. This one is slightly more complicated because both true and - are valid values for client.enabled.

Added a check to the top of the deployment spec and tests.

> Is having only the one error on the deployment enough? Or should the logic checking for connectInject and client.grpc be attached to all mesh-gateway components?

One is enough because even if you're rendering a single template via `helm template` it will fail. Helm must be rendering everything each time:
```
$ helm template \
  -x templates/mesh-gateway-clusterrole.yaml \
  --set 'meshGateway.enabled=true' .
Error: render error in "consul/templates/mesh-gateway-deployment.yaml": template: consul/templates/mesh-gateway-deployment.yaml:2:46: executing "consul/templates/mesh-gateway-deployment.yaml" at <fail "connectInject.enabled must be true">: error calling fail: connectInject.enabled must be true
```

> It would be good to break up the test files with headers for the various components being tested as defined here.

Done

templates/mesh-gateway-deployment.yaml
adilyse: Agreed!

PR #207, opened 2019-07-24
add aclTokens parameter

monchaio:
Add aclTokens as parameter under client section.
In case consul client join external server that required acl tokens.

### DIFF
--- a/templates/client-config-configmap.yaml
+++ b/templates/client-config-configmap.yaml
@@ -20,4 +20,18 @@ data:
       "enable_central_service_config": true
     }
   {{- end }}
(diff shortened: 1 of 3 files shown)


lkysow:
Hi Monchai, thanks for the PR. What about using the `client.extraConfig` key to set these?
```yaml
# values.yaml
client:
  extraConfig: |
    {
      "acl": {
          "enabled": true,
          "tokens": {
            "name": "token",
            "name2": "token"
          }
      }
    }    
```

monchaio:
@lkysow Okay. agree to use existing parameter

lkysow:
👍 let us know if you have any issues with that.

PR #208, opened 2019-07-24
Add annotations configuration to DNS service.

sunadm:
Many cloud providers support bind Load Balancer address by using service
annotations. Such as Amazon AWS, Alibaba AliCloud etc.

So, then we can set a fixed LB address in coredns configMap settings, not a non-stable dynamic cluster ip.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -254,6 +254,11 @@ client:
 dns:
   enabled: "-"
 
+  # Extra annotations to attach to the dns service
+  # This should be a multi-line string of
+  # annotations to apply to the dns Service
+  annotations: null
+
 ui:
   # True if you want to enable the Consul UI. The UI will run only
   # on the server nodes. This makes UI access via the service below (if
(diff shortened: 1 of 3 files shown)


lkysow:
Thanks for the PR. Please add two tests similar to https://github.com/hashicorp/consul-helm/blob/master/test/unit/mesh-gateway-service.bats#L42-L67 but in this file https://github.com/hashicorp/consul-helm/blob/master/test/unit/dns-service.bats

values.yaml
```diff
@@ -254,6 +254,11 @@ client:
 dns:
   enabled: "-"
 
+  # Extra annotations to attach to the dns service
+  # This should be a multi-line string mapping directly to the a map of
+  # the annotations to apply to the dns service
```
```suggestion
  # annotations to apply to the dns Service
```

lkysow:
Looking good, please squash your changes into a single commit and I'll merge.

sunadm:
> Looking good, please squash your changes into a single commit and I'll merge.

Done.

lkysow:
Thanks! Apologies for taking a while to get this merged.

Issue #120, opened 2019-02-21
Use recommended labels in chart


rmacduff:
Are there any plans to start using the recommended standard labels in the templates per the best practices doc: https://github.com/helm/helm/blob/master/docs/chart_best_practices/labels.md

```
app.kubernetes.io/name
helm.sh/chart
app.kubernetes.io/managed-by
app.kubernetes.io/instance
app.kubernetes.io/version
app.kubernetes.io/component
app.kubernetes.io/part-of
```

This would make it easier to search for kubernetes resources and in logs.

adilyse:
Hi @rmacduff,

This hadn't come up before, so there is no current plan. It hasn't been a priority, since we're still working to fix some known issues and add vital functionality. I'd be happy to take a look at a PR, though, if you'd like to add them.

One thing to check first though-- we've already had an issue with a label that changed per version interfering with upgrading via Helm (see #86). If adding these recommended labels causes the same problem, we'll continue without them.

PR #214, opened 2019-08-10
Use recommended labels as defined in kubernetes docs

jason-riddle:
Fixes https://github.com/hashicorp/consul-helm/issues/120.

Use labels as defined in https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels. There are now 5 labels for the outer template and 3 labels for the inner template (selector, matchLabels, etc.). Labels are not alphabetically organized in order to minimize the number of labels copying/pasting from the outer template to the inner template. Style of ordering was copied from https://github.com/hashicorp/vault-helm/pull/8.

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -6,24 +6,25 @@ metadata:
   name: {{ template "consul.fullname" . }}
   namespace: {{ .Release.Namespace }}
   labels:
-    app: {{ template "consul.name" . }}
-    chart: {{ template "consul.chart" . }}
-    heritage: {{ .Release.Service }}
-    release: {{ .Release.Name }}
+    helm.sh/chart: {{ include "consul.chart" . }}
+    app.kubernetes.io/name: {{ include "consul.name" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/version: {{ .Chart.Version | quote }}
 spec:
   selector:
     matchLabels:
-      app: {{ template "consul.name" . }}
-      chart: {{ template "consul.chart" . }}
-      release: {{ .Release.Name }}
+      helm.sh/chart: {{ template "consul.chart" . }}
+      app.kubernetes.io/name: {{ template "consul.name" . }}
+      app.kubernetes.io/instance: {{ .Release.Name }}
       component: client
       hasDNS: "true"
   template:
     metadata:
       labels:
-        app: {{ template "consul.name" . }}
-        chart: {{ template "consul.chart" . }}
-        release: {{ .Release.Name }}
+        helm.sh/chart: {{ template "consul.chart" . }}
+        app.kubernetes.io/name: {{ template "consul.name" . }}
+        app.kubernetes.io/instance: {{ .Release.Name }}
         component: client
         hasDNS: "true"
       annotations:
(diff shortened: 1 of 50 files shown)


templates/client-daemonset.yaml
```diff
@@ -6,24 +6,25 @@ metadata:
...
   labels:
-    app: {{ template "consul.name" . }}
-    chart: {{ template "consul.chart" . }}
-    heritage: {{ .Release.Service }}
-    release: {{ .Release.Name }}
+    helm.sh/chart: {{ include "consul.chart" . }}
+    app.kubernetes.io/name: {{ include "consul.name" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/version: {{ .Chart.Version | quote }}
```
lkysow: This prevents upgrades because `labels` is an immutable field for daemonset and statefulsets

jason-riddle:
> Is there a way to factor the labels into a helper function?

This seems reasonable. Two questions.

1. What should the helper function look like? Should it only include labels that can be use for deployments and not stateful/daemonsets? Maybe

```yaml
{{/*
Generate standard helm labels
*/}}
{{- define "consul.labels" }}
helm.sh/chart: {{ include "consul.chart" . }}
app.kubernetes.io/name: {{ include "consul.name" . }}
app.kubernetes.io/instance: {{ .Release.Name }}
app.kubernetes.io/managed-by: {{ .Release.Service }}
app.kubernetes.io/version: {{ .Chart.Version | quote }}
{{- if .Values.global.commonLabels }}
{{ toYaml .Values.global.commonLabels }}
{{- end }}
{{- end }}
```

2. How should it be used for top-level labels and selector-based labels? For example, for the dns service

```yaml
apiVersion: v1
kind: Service
metadata:
  name: {{ template "consul.fullname" . }}-dns
  namespace: {{ .Release.Namespace }}
  labels:
    {{- include "consul.labels" . | indent 4 -}}
spec:
  ports:
    - name: dns-tcp
      port: 53
      protocol: "TCP"
      targetPort: dns-tcp
    - name: dns-udp
      port: 53
      protocol: "UDP"
      targetPort: dns-udp
  selector:
    {{- include "consul.labels" . | indent 4 -}}
    hasDNS: "true"
```

and for the deployment snapshot agent

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: {{ template "consul.fullname" . }}-snapshot-agent
  namespace: {{ .Release.Namespace }}
  labels:
    {{- include "consul.labels" . | indent 4 -}}
spec:
  replicas: {{ .Values.client.snapshotAgent.replicas }}
  selector:
    matchLabels:
      {{- include "consul.labels" . | indent 6 -}}
      component: client-snapshot-agent
  template:
    metadata:
      labels:
        {{- include "consul.labels" . | indent 6 -}}
        component: client-snapshot-agent
      annotations:
```

templates/client-daemonset.yaml
jason-riddle: Ok, what do you recommend this should be changed to?

lkysow: Remove any label that will change if the chart version is upgraded.

jason-riddle:
@lkysow I'm going to close this PR and break this down into smaller pieces.

filintod:
I wanted to comment as I did something similar and one issue is with the ACL server init subcommand in consul-k8s where it tries to find the label selectors:

https://github.com/hashicorp/consul-k8s/blob/master/subcommand/server-acl-init/command.go#L113

So unless the consul k8s subcommand is updated/patched this labels change will break the acl init.

PR #217, opened 2019-08-14
Allow hardcoding DNS service's IP

gmaiztegi:
In order to use Consul's DNS inside a Kubernetes cluster, we need to manually add the service's IP address after it is created to CoreDNS's or kube-dns's configuration. If the service is recreated, we need to manually change the configuration to reflect this and forgetting this will cause issues.

With this we can hardcode the DNS service's cluster IP to a known value and avoid the extra manual work.

PS: I do not know if I should upgrade the chart's version. Just let me know if that is the case.

### DIFF
--- a/templates/dns-service.yaml
+++ b/templates/dns-service.yaml
@@ -11,6 +11,9 @@ metadata:
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
 spec:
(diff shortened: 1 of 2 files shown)


lkysow:
Thanks for the PR! I just merged https://github.com/hashicorp/consul-helm/pull/219 with the same fix so I'm going to close this PR.

PR #223, opened 2019-08-21
Mount in extra-configs as files instead of a whole folder

Sriharivignesh:
Mounting extra-from-values.json as a whole folder makes the folder read-only preventing people from adding in more configs to the config directory.

This PR modifies the chart to mount extra-from-values.json as a file and allows people to mount in other extra configs into the config directory as necessary.

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -134,7 +134,8 @@ spec:
             - name: data
               mountPath: /consul/data
             - name: config
-              mountPath: /consul/config
+              mountPath: /consul/config/extra-from-values.json
+              subPath: extra-from-values.json
             {{- range .Values.client.extraVolumes }}
             - name: userconfig-{{ .name }}
               readOnly: true
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -126,7 +126,8 @@ spec:
             - name: data-{{ .Release.Namespace }}
               mountPath: /consul/data
             - name: config
-              mountPath: /consul/config
+              mountPath: /consul/config/extra-from-values.json
+              subPath: extra-from-values.json
             {{- range .Values.server.extraVolumes }}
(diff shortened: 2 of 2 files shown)


lkysow:
Hi @Sriharivignesh,
We need to mount the whole config map instead of only the extra-values key because we support multiple keys that translate to multiple files:

https://github.com/hashicorp/consul-helm/blob/24d3115a2f04e8b5058f43b1d67afe6eff20df19/templates/server-config-configmap.yaml#L13-L17

https://github.com/hashicorp/consul-helm/blob/24d3115a2f04e8b5058f43b1d67afe6eff20df19/templates/client-config-configmap.yaml#L14-L21


If you need to add something to the `/consul/config` directory, you will need to do that using:
```yaml
# values.yaml
server:
  # extraConfig is a raw string of extra configuration to set with the
  # server. This should be JSON.
  extraConfig: |
    {}

client:
  # extraConfig is a raw string of extra configuration to set with the
  # client. This should be JSON.
  extraConfig: |
    {}
```

If you can't use the above method, you could add your own volume:
```yaml
server:
  # extraVolumes is a list of extra volumes to mount. These will be exposed
  # to Consul in the path `/consul/userconfig/<name>/`. The value below is
  # an array of objects, examples are shown below.
  extraVolumes: []
    # - type: secret (or "configMap")
    #   name: my-secret
client:
  # extraVolumes is a list of extra volumes to mount. These will be exposed
  # to Consul in the path `/consul/userconfig/<name>/`. The value below is
  # an array of objects, examples are shown below.
  extraVolumes: []
    # - type: secret (or "configMap")
    #   name: my-secret
```

And set `load: true`. This will add the `-config-dir=/consul/userconfig/<name>` flag to Consul so it will read the files in that volume.

I'm going to close this due to the reasons above but thanks for the PR and hopefully these solutions work for you.

PR #224, opened 2019-09-04
Less privileges when syncCatalog.toK8S = false

f84anton:


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ BUG FIXES:
 
   * Fix enterprise license application when ACLs are turned off
 
+IMPROVEMENTS:
```
lkysow: We'll add this change when we do a release. Please remove it for now because it causes issues with merging.

Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: consul
-version: 0.9.0-beta
+version: 0.9.1-beta
```
lkysow: We'll make this change when we do a release. Please remove it for now because it causes issues with merging.

PR #233, opened 2019-09-24
Configuration to save persistent data on node, updateStrategy and dnsPolicy configuration

AlexShemeshWix:
**Configuration to save persistent data on node**
During consul agent upgrade , consul agent pod gets restarted and since /consul/data folder is mounted to emptyDir - all the data is erased, and new agent will try to register as a completely new agent.
If we save /consul/data content between pod restarts, new pod will continue with id and configuration that previous pod had.
This will decrease pressure on consul server during consul-agent upgrades.
This pull request adds hostPath parameter that can be configured with path to host machine folder to save contents of /consul/data folder.If this parameter is not defined everything will continue to work as before.

**dnsPolicy configuration**
If you use external DNS service you may want to explicitly contol DNS resolution priorities

**updateStrategy configuration**
On big clusters it may take hours to upgrade daemonset with 500+ pods, modifying updateStrategy  can drastically reduce rollout time.

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -11,6 +11,10 @@ metadata:
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
 spec:
+  {{- if .Values.client.updateStrategy }}
+  updateStrategy:
+    {{ toYaml .Values.client.updateStrategy | nindent 4 | trim }}
+  {{- end }}
   selector:
     matchLabels:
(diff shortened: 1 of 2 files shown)


labria:
Amazing, especially the `hostPath` setting, because otherwise consul agents will start with a blank state, and you have to explicitly re-register all services on the node.

AlexShemeshWix:
Any update?

AlexShemeshWix:
Ay update?

PR #249, opened 2019-10-15
Add flag for turning off tests

lkysow:
Useful in the case where users are running `helm template` and they don't want the test Pod submitted. When running `helm install`, helm won't submit the Pod, only if users manually run `helm test`.

Adds docs to #144, closes #144


lkysow:
> This looks good, but I think it'd be nice to add a unit test for this.

Good call, I just added a test for the flag on/off rather than the contents of the Pod manifest since that unnecessary but happy to add a bit more template testing if you think it's warranted.

PR #254, opened 2019-10-16
Add CircleCI config

ishustava:
This PR adds CircleCI config to run unit and acceptance tests along with some other changes:

- Update `Test.dockerfile` to install terraform `0.12` so we can run terraform commands from within the container
- Upgrade terraform templates used in acceptance tests to terraform `0.12`
- Remove deprecated properties in resources from terraform templates


test/docker/Test.dockerfile
```diff
@@ -31,6 +31,11 @@ RUN curl -OL https://dl.google.com/dl/cloudsdk/channels/rapid/install_google_clo
     bash install_google_cloud_sdk.bash --disable-prompts --install-dir='/root/' && \
     ln -s /root/google-cloud-sdk/bin/gcloud /usr/local/bin/gcloud
 
+# terraform
+RUN curl -sSL https://releases.hashicorp.com/terraform/0.12.10/terraform_0.12.10_linux_amd64.zip -o /tmp/tf.zip \
```
alvin-huang: setting the version with an `ENV` might be easier to find in case this file gets larger

ishustava: Thanks so much for a quick review. I've updated the config's file name and the terraform version.

alvin-huang:
Looks like the test errors are probably from not setting a working directory. The way Circle does it is setting up a separate user and using it's home directory: https://github.com/CircleCI-Public/circleci-dockerfiles/blob/62eb93e6af5efa502580106f9ba131e3d163abbf/golang/images/1.13.0-stretch/Dockerfile#L87-L103

.circleci/config.yml
```diff
@@ -0,0 +1,42 @@
+version: 2
+jobs:
+  unit:
+    docker:
+      - image: hashicorpdev/consul-helm-test:0.2.0
```
lkysow: Is this created by `test/docker/Test.dockerfile`? If so can you put a comment at the top of this file about where the `hashicorpdev/consul-helm-test` image comes from.

.circleci/config.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - checkout
+
+      - run:
+          name: Run Acceptance Tests
+          command: |
+            set -e
+
+            terraform init
+            terraform apply -var 'init_cli=true' -var 'project=${GOOGLE_PROJECT} --auto-approve
```
lkysow: the `'` are unbalanced

.circleci/config.yml
```diff
@@ -0,0 +1,42 @@
...
+
+            bats ./test/acceptance
+
+            terraform destroy -var 'project=${GOOGLE_PROJECT} --force
```
lkysow: I think there's a way to have this always run, regardless of pass/fail. Also `'` is unbalanced

.circleci/config.yml
```diff
@@ -0,0 +1,42 @@
...
+          environment:
+            TF_VAR_init_cli: true
+
+workflows:
+  version: 2
+  test:
+    jobs:
+      - unit
+      - acceptance:
```
lkysow: I thought we only wanted to run this on merges to master?

test/docker/Test.dockerfile
```diff
@@ -31,6 +32,11 @@ RUN curl -OL https://dl.google.com/dl/cloudsdk/channels/rapid/install_google_clo
     bash install_google_cloud_sdk.bash --disable-prompts --install-dir='/root/' && \
     ln -s /root/google-cloud-sdk/bin/gcloud /usr/local/bin/gcloud
 
+# terraform
+RUN curl -sSL https://releases.hashicorp.com/terraform/${TERRAFORM_VERSION}/terraform_${TERRAFORM_VERSION}_linux_amd64.zip -o /tmp/tf.zip \
```
lkysow: We could use a multistage Docker build and copy it from the official image:
`COPY --from=hashicorp/terraform:0.12.10 <not sure exact path>/terraform /usr/local/bin/terraform`

https://docs.docker.com/develop/develop-images/multistage-build/#use-an-external-image-as-a-stage

Not sure which one would be faster so it probably doesn't matter.

.circleci/config.yml
ishustava: Yep it is, I've added a comment 🙏

.circleci/config.yml
alvin-huang: https://circleci.com/docs/2.0/configuration-reference/#the-when-attribute you can use `when:` inside a run step. In this case, the destroy can just be a separate `run` key so the when attribute is only applied to destroy and not everything else.

.circleci/config.yml
```diff
@@ -0,0 +1,44 @@
+version: 2
+jobs:
+  unit:
+    docker:
+      # This image is build from test/docker/Test.dockerfile
```
lkysow:
```suggestion
      # This image is built from test/docker/Test.dockerfile
```

.circleci/config.yml
ishustava: I've added a filter, as you have suggested, thank you!

.circleci/config.yml
ishustava: This should be fixed now.

.circleci/config.yml
ishustava: I've split out the mono-step into multiple steps, and now `terraform destroy` can run always. Thank you both 😃

test/docker/Test.dockerfile
ishustava: I didn't know about this feature, this is awesome 💯. In this case though, it doesn't matter that much. The build time with `curl` was a few seconds faster a couple of times I tried it, but the results were not consistent.

Issue #253, opened 2019-10-16
Support External Consul Servers


lkysow:
Support exposing Consul clients running in Kube to Consul servers running outside of Kube

stvdilln:
I'm working on the PR doing some final testing.  I would like to fix as many scenarios as possible.

Do you see any flaws to this:
If we are only installing client (and not server) then client 8301,8302 are going to listen on NodePorts and can communicate with external server.

If we are only installing the server then 8300,8310,8302 are going to listen on NodePorts and this exposes the server cluster to outside world.

If we are installing both client and server:
The server listens on NodePorts, The client listens on Pod Ports.   When the server does a connect to the client it should open a socket on the pod port and the connection is pod port<->pod port.  When the client wants to connect it would be PodPort (client)<->NodePort (server).

If you want I can just fix the client connecting to an external server in this PR, but I think I can fix both the client to an external server and allowing external access to the server nodes running in K8s in one pass.

lkysow:
> If you want I can just fix the client connecting to an external server in this PR, but I think I can fix both the client to an external server and allowing external access to the server nodes running in K8s in one pass.

We're interested only in the client to external server right now. We need to think more about the other cases.

Notes:
1. We don't want the clients listening on host ports (other than 8500 and 8502) by default, this should be optional
1. It should be a separate PR for exposing the servers. There's already some work done in https://github.com/hashicorp/consul-helm/pull/27.
1. If installing both client and server... we already support this so I wouldn't want to make any changes that aren't optional. We're not exposing the server on hostPorts right now because the default configuration has Server pods run on the same nodes as the client daemonset so there would be port conflicts.

stvdilln:
I had started the work on doing an external server and it got messy quick so I abandoned that.  The possibilty of Client, or Server or Both running on same node quickly exploded the complexity.

I've submitted a PR that if you are not installing the server also, ports 8301 and 8302 are exposed via hostPorts so that the external consul cluster can connect to the k8s based clients.  Before this there was a one way communcation where the clients could connect to the external servers, but not the other way around.
The opening on 8301 and 8302 is contingent on server.enabled=false.
We can create a new explicit flag  client.useExternalServer or key the exposing of the ports to client.
We could also key off of client.join which indirectly implies an external server.  

Look though the PR 256 and let me know if you would like any changes.

mekza:
[...]

PR #256, opened 2019-10-17
Add support for servers not in k8s cluster

stvdilln:
The consul client was formerly unable to communicate with a consul server cluster outside of k8s.  This change specifies that if a server is not going to be installed in k8s, the 8301 and 8302 ports are allocated on the host machine, which will allow the existing consul servers to connect to the k8s consul clients.

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -175,14 +175,30 @@ spec:
               # We are not installing server, take the host port which allows an external server to connect.
               hostPort: 8301
               {{- end }}
-              name: serflan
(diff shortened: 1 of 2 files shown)


stvdilln:
This is tracked in issue [253](https://github.com/hashicorp/consul-helm/issues/253)

stvdilln:
Hey,
I was out of office last week (at Kubecon, Yea!).  Didn't see the comment till this morning.  I see that you have the changes already complete, so this PR is probably dead.  Nice to see this feature making it's way in.

PR #259, opened 2019-10-17
Run `helm test` in acceptance tests

lkysow:
`helm test` will check that the servers are functional (at least for KV). I know our acceptance tests are pretty minimal right now but this adds a bit more testing.


lkysow:
🤞 I don't know if this actually passes

PR #261, opened 2019-10-18
Document aclBindingRuleSelector

lkysow:
Thought it worth to expand a bit here.


values.yaml
```diff
@@ -420,10 +420,16 @@ connectInject:
   #   beta.kubernetes.io/arch: amd64
   nodeSelector: null
 
-  # aclBindingRuleSelector is a string that defines the automatic binding
-  # rule to control the allowed authentication for Connect injection. The
-  # default disallows using the default service account for ACl generation.
-  # Requires Consul v1.5+ and consul-k8s v0.8.0+.
```
ishustava: I think it's helpful to keep the minimum Consul and consul-k8s versions here.

lkysow: Yes, that was deleted by accident. Adding back in.

Issue #264, opened 2019-10-22
Consul members unable to elect a leader


tsturzl:
Basically what I'm seeing is all of the consul pods in the statefulset are running, but their readiness probes never succeed because the consul peers never elect a leader. I had previously had a cluster up and running, but removed it since it wasn't needed immediately. This was about 3 days ago. I deleted the PVC's from the stateful set after purging through helm, and when I recreated I'm unable to get the cluster working again. Nothing in the helm values have changed.

All of the logs from each of the 3 nodes looks like this:

```
bootstrap_expect > 0: expecting 3 servers
==> Starting Consul agent...
           Version: 'v1.6.1'
           Node ID: 'c63f2fbf-6a8a-647d-d4e2-61ae0b7f5276'
         Node name: 'vault-consul-consul-server-0'
        Datacenter: 'dc1' (Segment: '<all>')
            Server: true (Bootstrap: false)
       Client Addr: [0.0.0.0] (HTTP: 8500, HTTPS: -1, gRPC: -1, DNS: 8600)
      Cluster Addr: 100.96.12.37 (LAN: 8301, WAN: 8302)
           Encrypt: Gossip: false, TLS-Outgoing: false, TLS-Incoming: false, Auto-Encrypt-TLS: false

==> Log data will now stream in as it occurs:

    2019/10/22 20:36:39 [INFO]  raft: Initial configuration (index=0): []
    2019/10/22 20:36:39 [INFO]  raft: Node at 100.96.12.37:8300 [Follower] entering Follower state (Leader: "")
    2019/10/22 20:36:39 [INFO] serf: EventMemberJoin: vault-consul-consul-server-0.dc1 100.96.12.37
    2019/10/22 20:36:39 [INFO] serf: EventMemberJoin: vault-consul-consul-server-0 100.96.12.37
    2019/10/22 20:36:39 [INFO] agent: Started DNS server 0.0.0.0:8600 (udp)
    2019/10/22 20:36:39 [INFO] consul: Adding LAN server vault-consul-consul-server-0 (Addr: tcp/100.96.12.37:8300) (DC: dc1)
    2019/10/22 20:36:39 [INFO] agent: Started DNS server 0.0.0.0:8600 (tcp)
    2019/10/22 20:36:39 [INFO] consul: Handled member-join event for server "vault-consul-consul-server-0.dc1" in area "wan"
    2019/10/22 20:36:39 [INFO] agent: Started HTTP server on [::]:8500 (tcp)
[...]

Issue #265, opened 2019-10-23
PodDisruptionBudget calculation is not working


s3than:
As per #71 the latest version of helm https://github.com/helm/helm/releases/tag/v2.15.0 includes the following `Updated Sprig to 2.20.0`

This causes the previous fix to pod disruption budget to return -1 now on a new installation.

Updating the _helpers.tpl to the following resolves however if you run it in helm < 2.15.0 it's broken again..

I'll hopefully have a fix shortly that works in both and will advise.

```
{{- define "consul.pdb.maxUnavailable" -}}
{{- if eq (int .Values.server.replicas) 1 -}}
{{ 0 }}
{{- else if .Values.server.disruptionBudget.maxUnavailable -}}
{{ .Values.server.disruptionBudget.maxUnavailable -}}
{{- else -}}
{{- if eq (int .Values.server.replicas) 3 -}}
{{- 1 -}}
{{- else -}}
{{- div (int .Values.server.replicas) 2 -}}
{{- end -}}
{{- end -}}
{{- end -}}
```

s3than:
Resolution is same as #264

lkysow:
Yeah looks like helm has reverted this change in their latest release (v2.15.1): https://github.com/helm/helm/releases/tag/v2.15.1

I don't think there's a good workaround on our side other than asking users to update Helm.

s3than:
Updated pull request to just include tests, happy to close as resolved when using 2.15.1 of helm

PR #266, opened 2019-10-23
Fix to retry join not populating with helm 2.15.0

s3than:
Resolves issues
#264 
#265

### DIFF
--- a/test/unit/client-daemonset.bats
+++ b/test/unit/client-daemonset.bats
@@ -75,7 +75,7 @@ load _helpers
 #--------------------------------------------------------------------
 # retry-join
 
-@test "server/StatefulSet: retry join gets populated" {
+@test "client/DaemonSet: retry join gets populated" {
   cd `chart_dir`
   local actual=$(helm template \
       -x templates/client-daemonset.yaml  \
--- a/test/unit/client-daemonset.bats
+++ b/test/unit/client-daemonset.bats
@@ -73,6 +73,21 @@ load _helpers
 }
 
 #--------------------------------------------------------------------
+# retry-join
+
+@test "server/StatefulSet: retry join gets populated" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/client-daemonset.yaml  \
+      --set 'server.replicas=3' \
+      . | tee /dev/stderr |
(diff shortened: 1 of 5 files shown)


filintod:
There is a new version of Helm 2.15.1 that revert changes that created the regression issue (https://github.com/helm/helm/releases/tag/v2.15.1). I think we should just state that people should not use Helm 2.15.0 as it is flawed.

lkysow:
Thanks @filintod I agree that since it's a bug in Helm that they've reverted we shouldn't add a workaround to our chart. How does that sound @s3than?

s3than:
I'll re-run all the tests as it uses helm latest tag for testing. If it all works out happy to just add test coverage to a PR.

s3than:
@lkysow I've reverted the change to the helm charts and left the tests.

The changes still worked under 2.15.1 but I'd rather just have the test coverage.

lkysow:
@s3than good call, thanks!!

test/unit/server-statefulset.bats
```diff
@@ -42,6 +42,20 @@ load _helpers
...
+#--------------------------------------------------------------------
+# retry-join
+
+@test "server/StatefulSet: retry join gets populated" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/server-statefulset.yaml  \
+      --set 'server.replicas=3' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.containers[0].command | any(contains("-retry-join"))' | tee /dev/stderr)
```
lkysow: Same for here (see above)

test/unit/client-daemonset.bats
```diff
@@ -72,6 +72,21 @@ load _helpers
...
+#--------------------------------------------------------------------
+# retry-join
+
+@test "client/DaemonSet: retry join gets populated" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/client-daemonset.yaml  \
+      --set 'server.replicas=3' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.containers[0].command | any(contains("-retry-join"))' | tee /dev/stderr)
```
lkysow: Not sure where my comment went! I asked if it was possible to match exactly on the retry-join but now I realize it's not easy because it's just a string at that point. Nevermind!

PR #267, opened 2019-10-23
Allow configuration of hostPorts on client daemonSet

ncouse:
The `hostPort`s for the client daemonSet are hardcoded (ports 8500 and 8502).

This can cause a potential port conflict. While normally the defaults values are desirable, this PR allows you to modify these defaults.

To demonstrate this, you can deploy 2 Consul clusters, where the `hostPort`s are different, and don't conflict.

```
helm upgrade --install consul1 --namespace consul1 .
helm upgrade --install consul2 --namespace consul2 . --set client.hostPort.http=9500 --set client.hostPort.grpc=9502
```

Feel free to change names of values if not consistent.

PR #268, opened 2019-10-24
Allow configuration of hostPorts on client daemonSet

ncouse:
Re-creation of #267 from new branch.

---

The `hostPort`s for the client daemonSet are hardcoded (ports 8500 and 8502).

This can cause a potential port conflict. While normally the defaults values are desirable, this PR allows you to modify these defaults.

To demonstrate this, you can deploy 2 Consul clusters, where the `hostPort`s are different, and don't conflict.

```
helm upgrade --install consul1 --namespace consul1 .
helm upgrade --install consul2 --namespace consul2 . --set client.hostPort.http=9500 --set client.hostPort.grpc=9502
```

Feel free to change names of values if not consistent.

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -157,10 +157,10 @@ spec:
                 - consul leave
           ports:
             - containerPort: 8500
-              hostPort: 8500
+              hostPort: {{ .Values.client.hostPort.http }}
               name: http
             - containerPort: 8502
-              hostPort: 8502
+              hostPort: {{ .Values.client.hostPort.grpc }}
               name: grpc
             - containerPort: 8301
               name: serflan
--- a/values.yaml
+++ b/values.yaml
@@ -167,7 +167,8 @@ client:
   image: null
   join: null
 
-  # Host ports uses by daemon set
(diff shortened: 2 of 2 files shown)


lkysow:
We'll need to check that we're passing these values into Consul so that it's actually using those ports. Right now we're not changing how Consul is being configured so I'm not sure this actually works?

ncouse:
Yes, it is just the `hostPort` that I wanted to make configurable. The default is still 8500/8502. 

In the case of multiple installs or where for some reason there is a port conflict on any one of the nodes, making this configurable helps. Sometimes we have to install into k8s clusters (e.g. on a customer site that we don't have full control of) that may have port conflicts, or another Consul cluster installed.

The comment for these properties in `values.yml` probably should be updated to note that this may break any applications that depend on the default values (such as `consul-k8s`).

The intent was to at least allow us to install if there are problem, knowing we need to update our own application configuration.

There are probably other enhancements, such as not specifying a gRPC `hostPort` if gRPC is disabled. However, I wanted to keep this PR as simple as possible.

t-eckert:

PR #280, opened 2019-11-04
syncCatalog: expose add-k8s-namespace-suffix flag

ishustava:
Depends on hashicorp/consul-k8s#160.

### DIFF
--- a/templates/sync-catalog-deployment.yaml
+++ b/templates/sync-catalog-deployment.yaml
@@ -94,6 +94,9 @@ spec:
                 {{- if .Values.syncCatalog.consulPrefix}}
                 -consul-service-prefix="{{ .Values.syncCatalog.consulPrefix}}" \
                 {{- end}}
+                {{- if .Values.syncCatalog.addK8SNamespaceSuffix}}
+                -add-k8s-namespace-suffix="{{ .Values.syncCatalog.addK8SNamespaceSuffix}}" \
+                {{- end}}
           livenessProbe:
             httpGet:
               path: /health/ready
--- a/templates/sync-catalog-deployment.yaml
+++ b/templates/sync-catalog-deployment.yaml
@@ -95,7 +95,7 @@ spec:
                 -consul-service-prefix="{{ .Values.syncCatalog.consulPrefix}}" \
                 {{- end}}
                 {{- if .Values.syncCatalog.addK8SNamespaceSuffix}}
-                -add-k8s-namespace-suffix="{{ .Values.syncCatalog.addK8SNamespaceSuffix}}" \
+                -add-k8s-namespace-suffix \
                 {{- end}}
           livenessProbe:
             httpGet:
--- a/test/unit/sync-catalog-deployment.bats
+++ b/test/unit/sync-catalog-deployment.bats
@@ -382,16 +382,16 @@ load _helpers
       --set 'syncCatalog.enabled=true' \
       . | tee /dev/stderr |
       yq '.spec.template.spec.containers[0].command | any(contains("-add-k8s-namespace-suffix"))' | tee /dev/stderr)
-  [ "${actual}" = "false" ]
+  [ "${actual}" = "true" ]
 }
 
-@test "syncCatalog/Deployment: can set addK8SNamespaceSuffix to true" {
+@test "syncCatalog/Deployment: can set addK8SNamespaceSuffix to false" {
   cd `chart_dir`
   local actual=$(helm template \
       -x templates/sync-catalog-deployment.yaml  \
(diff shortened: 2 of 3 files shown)


values.yaml
```diff
@@ -328,10 +328,24 @@ syncCatalog:
...
+  # addK8SNamespaceSuffix appends Kubernetes namespace suffix to
+  # each service name synced to Consul, separated by an underscore.
+  # For example, for a service 'foo' in the default namespace,
+  # the sync process will create a Consul service named 'foo_default'.
+  # Set this flag to true to avoid registering services with the same name
+  # but in different namespaces as instances for the same Consul service.
+  # Namespace suffix is not added if 'annotationServiceName' is provided.
+  # NOTE: Setting this property to 'true' will result in deregistering of existing
+  # synced services in Consul and registering them with a new name.
+  addK8SNamespaceSuffix: false
```
lkysow: I think we should set this to `true` by default. The current default has the major problem that a service with the same name in any namespace will be synced to the same Consul service.

We'd need to document the breaking change and tell people how to upgrade safely (set this value to false) but I think it's worth it to not put new users in a bad state.

templates/sync-catalog-deployment.yaml
```diff
@@ -94,6 +94,9 @@ spec:
                 {{- if .Values.syncCatalog.consulPrefix}}
                 -consul-service-prefix="{{ .Values.syncCatalog.consulPrefix}}" \
                 {{- end}}
+                {{- if .Values.syncCatalog.addK8SNamespaceSuffix}}
+                -add-k8s-namespace-suffix="{{ .Values.syncCatalog.addK8SNamespaceSuffix}}" \
```
lkysow: This is a bool flag so can this be 
```
{{- if .Values.syncCatalog.addK8SNamespaceSuffix}}
-add-k8s-namespace-suffix \
```

values.yaml
ishustava: @lkysow After thinking about this more, I agree that this should be true by default and that it's worth making this a breaking change.

values.yaml
```diff
@@ -328,10 +328,24 @@ syncCatalog:
...
+  # synced services in Consul and registering them with a new name.
+  addK8SNamespaceSuffix: true
+
+  # consulPrefix is the service prefix which prepends itself
   # to Kubernetes services registered within Consul
-  # For example, "k8s-" will register all services peprended with "k8s-".
+  # For example, "k8s-" will register all services prepended with "k8s-".
   # (Kubernetes -> Consul sync)
+  # consulPrefix is ignored when 'annotationServiceName' is provided.
+  # NOTE: Setting this property to a non-null value will result in deregistering
```
```suggestion
  # NOTE: Updating this property to a non-null value for an existing installation will result in deregistering
```

Presumably the situation this comment is warning about doesn't happen during a first installation.

values.yaml
```diff
@@ -328,10 +328,24 @@ syncCatalog:
...
 
-  # consulPrefix is the service prefix which preprends itself
+  # addK8SNamespaceSuffix appends Kubernetes namespace suffix to
+  # each service name synced to Consul, separated by a dash.
+  # For example, for a service 'foo' in the default namespace,
+  # the sync process will create a Consul service named 'foo-default'.
+  # Set this flag to true to avoid registering services with the same name
+  # but in different namespaces as instances for the same Consul service.
+  # Namespace suffix is not added if 'annotationServiceName' is provided.
+  # NOTE: Setting this property to 'true' will result in deregistering of existing
```
adilyse: This is only a concern when someone is updating an existing Consul installation. This comment reads to me like it might happen if someone sets it to `true` the first time they install. 

Also, given that we're now defaulting this to `true`, the wording is a little odd.

test/unit/sync-catalog-deployment.bats
```diff
@@ -371,3 +371,27 @@ load _helpers
       yq -r '.command | any(contains("consul-k8s acl-init"))' | tee /dev/stderr)
   [ "${actual}" = "true" ]
 }
+
+#--------------------------------------------------------------------
+# addK8SNamespaceSuffix
+
+@test "syncCatalog/Deployment: no k8s namespace prefix by default" {
```
adilyse:
```suggestion
@test "syncCatalog/Deployment: k8s namespace suffix enabled by default" {
```

values.yaml
ishustava: Yep, good point!

values.yaml
ishustava: True. I removed that note entirely. This comment will be more appropriate in the Changelog.

PR #282, opened 2019-11-07
Prevent incorrect connectInject config

lkysow:
Connect injection requires clients to be enabled and the gRPC port to be
enabled. These requirements are documented on the website although not
in the values.yaml itself. If users do not enable clients, their init
pods will fail. If users do not enable gRPC, their sidecars will be
reported as unhealthy. Both these outcomes are hard to debug.

This change makes it harder to get into this misconfiguration. It will
cause a helm error if clients or grpc are not enabled when connect is
enabled.

It also defaults client.grpc to true. There are no downsides to having
the grpc port open even if not using connect inject. This change means
enabling connect simply requires setting connectInject to true rather
than also setting client.grpc to true (something that's unintuitive and
easy to miss).

This is a breaking change, however any users that had connect enabled
but clients or gRPC not enabled would be running into errors so it's
not going to break anything for anyone that isn't already in a broken
state.

The error users would see if using inject but not clients:
```
$ helm template ./ --set client.enabled=false --set connectInject.enabled=true
Error: render error in "consul/templates/connect-inject-deployment.yaml": template: \
consul/templates/connect-inject-deployment.yaml:3:169: executing \
 "consul/templates/connect-inject-deployment.yaml" at \
<fail "clients must be enabled for connect injection">:\
 error calling fail: clients must be enabled for connect injection
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 BREAKING CHANGES
   * `client.grpc` defaults to `true` now instead of `false`. This is to make it
-     harder to misconfigure connect [[GH-282](https://github.com/hashicorp/consul-helm/pull/282)]
+     harder to misconfigure Connect [[GH-282](https://github.com/hashicorp/consul-helm/pull/282)]
      
-     If you wish to not have client grpc ports exposed, set `client.grpc` to
+     If you do not wish to enable gRPC for clients, set `client.grpc` to
      `false` in your local values file.
 
 ## 0.12.0 (Oct 28, 2019)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## Unreleased
+
+BREAKING CHANGES
+  * `client.grpc` defaults to `true` now instead of `false`. This is to make it
+     harder to misconfigure connect [[GH-282](https://github.com/hashicorp/consul-helm/pull/282)]
+     
+     If you wish to not have client grpc ports exposed, set `client.grpc` to
+     `false` in your local values file.
+
 ## 0.12.0 (Oct 28, 2019)
 
 IMPROVEMENTS:
--- a/templates/connect-inject-deployment.yaml
+++ b/templates/connect-inject-deployment.yaml
@@ -1,5 +1,7 @@
 # The deployment for running the Connect sidecar injector
 {{- if (or (and (ne (.Values.connectInject.enabled | toString) "-") .Values.connectInject.enabled) (and (eq (.Values.connectInject.enabled | toString) "-") .Values.global.enabled)) }}
+{{- if not (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}{{ fail "clients must be enabled for connect injection" }}{{ end }}
+{{- if not .Values.client.grpc }}{{ fail "client.grpc must be true for connect injection" }}{{ end }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
--- a/test/unit/client-daemonset.bats
+++ b/test/unit/client-daemonset.bats
@@ -90,23 +90,23 @@ load _helpers
 #--------------------------------------------------------------------
 # grpc
 
-@test "client/DaemonSet: grpc is disabled by default" {
+@test "client/DaemonSet: grpc is enabled by default" {
   cd `chart_dir`
   local actual=$(helm template \
       -x templates/client-daemonset.yaml  \
-      . | tee /dev/stderr |
-      yq '.spec.template.spec.containers[0].command | any(contains("grpc"))' | tee /dev/stderr)
-  [ "${actual}" = "false" ]
-}
-
-@test "client/DaemonSet: grpc can be enabled" {
-  cd `chart_dir`
-  local actual=$(helm template \
-      -x templates/client-daemonset.yaml  \
-      --set 'client.grpc=true' \
       . | tee /dev/stderr |
       yq '.spec.template.spec.containers[0].command | any(contains("grpc"))' | tee /dev/stderr)
   [ "${actual}" = "true" ]
+}
+
+@test "client/DaemonSet: grpc can be disabled" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/client-daemonset.yaml  \
+      --set 'client.grpc=false' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command | any(contains("grpc"))' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
 }
 
 #--------------------------------------------------------------------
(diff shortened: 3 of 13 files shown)


test/unit/client-daemonset.bats
```diff
@@ -109,6 +109,48 @@ load _helpers
...
+@test "client/DaemonSet: grpc is disabled if global.enabled=false" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/client-daemonset.yaml  \
+      --set 'global.enabled=false' \
+      --set 'client.enabled=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command | any(contains("grpc"))' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+}
```
ishustava: Hmmm this test case is interesting. Currently the client daemon set template is disabled iff `global.enabled==false` AND `client.enabled=="-"`. So when client is explicitly enabled or disabled, the helm chart will ignore the global value. This means that the template will be rendered with whatever value `grpc` is set to in `values.yaml`, which happens to be `false` by default. Maybe a better test would be setting `client.enabled` to `-` and setting `client.grpc` to `true` and then checking that `grpc` is not present in the rendered template.

templates/connect-inject-deployment.yaml
```diff
@@ -1,5 +1,6 @@
 # The deployment for running the Connect sidecar injector
 {{- if (or (and (ne (.Values.connectInject.enabled | toString) "-") .Values.connectInject.enabled) (and (eq (.Values.connectInject.enabled | toString) "-") .Values.global.enabled)) }}
+{{- if not (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}{{ fail "clients must be enabled for connect injection" }}{{ end }}
```
ishustava: This is great!

test/unit/connect-inject-deployment.bats
```diff
@@ -42,6 +43,27 @@ load _helpers
...
+  cd `chart_dir`
+  run helm template \
+      -x templates/connect-inject-deployment.yaml  \
+      --set 'global.enabled=false' \
+      --set 'connectInject.enabled=true' .
+  [ "$status" -eq 1 ]
+  [[ "$output" =~ "clients must be enabled for connect injection" ]]
+}
+
+@test "connectInject/Deployment: fails if global.enabled=true client.enabled=false" {
```
ishustava:
```suggestion
@test "connectInject/Deployment: fails if global.enabled=true and client.enabled=false" {
```

test/unit/connect-inject-deployment.bats
```diff
@@ -42,6 +43,27 @@ load _helpers
...
+  cd `chart_dir`
+  run helm template \
+      -x templates/connect-inject-deployment.yaml  \
+      --set 'global.enabled=true' \
+      --set 'client.enabled=false' \
+      --set 'connectInject.enabled=true' .
+  [ "$status" -eq 1 ]
+  [[ "$output" =~ "clients must be enabled for connect injection" ]]
+}
+
```
ishustava: This failure will also occur when `global.enabled=false` and `client.enabled=false`, which I believe is not tested.

test/unit/client-daemonset.bats
lkysow: > . Currently the client daemon set template is disabled iff global.enabled==false AND client.enabled=="-".

That seems like a bug. If client.enabled=false then it should be disabled.

> Maybe a better test would be setting client.enabled to - and setting client.grpc to true and then checking that grpc is not present in the rendered template.

In this case what's global.enabled set to? If true then we would expect client.grpc to be equal to true and if false then the whole template wouldn't render.

I'm not exactly sure what you're asking, maybe you can write the test you think this should be changed to?

test/unit/connect-inject-deployment.bats
lkysow: done!

lkysow:
@adilyse I think the tradeoff is worth it because any users that had connect enabled
but no clients or gRPC not enabled would be running into errors so it's
not going to break anything for anyone that isn't already in a broken
state.

The tradeoff between failing the template and this change is that you are connect-ready with this values.yaml:
```yaml
connectInject:
  enabled: true
```

vs. the experience that doubtless many of our users would run into which would be using that values.yaml, then getting this error which is hard to understand:
```
Error: render error in "consul/templates/connect-inject-deployment.yaml": template: \
consul/templates/connect-inject-deployment.yaml:3:169: executing \
 "consul/templates/connect-inject-deployment.yaml" at \
<fail "client.grpc must be true for connect injection">:\
 error calling fail: client.grpc must be true for connect injection
```

What if we document that `client.grpc` will be set to true if connectInject is enabled?

test/unit/client-daemonset.bats
ishustava: Sorry I misspoke, client template is disabled in two cases:
1. `client.enabled==false`
2. `global.enabled==false` AND `client.enabled=="-"`

I was mostly talking about the second case, but didn't mention it explicitly. 

I was a little confused by which logic this test is testing. In the test both `client.enabled` and `global.enabled` are set. The logic for the `global.enabled` value is never executed because it just short-circuits once it sees that `client.enabled` is true. 

There is already a test that tests that the template is rendered if `global.enabled` is `false` and `client.enabled` is `true`. I think we can assume that this will apply to all values set for the client and we don't have to test each value individually.

So maybe this test isn't necessary at all, unless you had some other intention behind it.

lkysow:
I've made the change to default `grpc` to true. Please re-review @ishustava @adilyse

test/unit/connect-inject-clusterrolebinding.bats
```diff
@@ -16,6 +16,7 @@ load _helpers
   local actual=$(helm template \
       -x templates/connect-inject-clusterrolebinding.yaml  \
       --set 'global.enabled=false' \
+      --set 'client.enabled=true' \
```
lkysow: These had to be set to true because otherwise the connect-inject-* templates don't render now.

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## Unreleased
+
+BREAKING CHANGES
+  * `client.grpc` defaults to `true` now instead of `false`. This is to make it
+     harder to misconfigure connect [[GH-282](https://github.com/hashicorp/consul-helm/pull/282)]
```
ishustava:
```suggestion
     harder to misconfigure Connect [[GH-282](https://github.com/hashicorp/consul-helm/pull/282)]
```

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
...
+     
+     If you wish to not have client grpc ports exposed, set `client.grpc` to
```
ishustava:
```suggestion
     If you do not wish to enable gRPC for clients, set `client.grpc` to
```

PR #289, opened 2019-11-26
Support external-to-k8s Consul servers

adilyse:
Allows Consul clients to expose their gossip ports as hostPorts. This allows them to connect to Consul servers that are external to Kubernetes in environments where pod ips are not directly accessible.

Builds on the work done in #256.

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -77,13 +77,11 @@ spec:
             - name: ADVERTISE_IP
               valueFrom:
                 fieldRef:
-                  {{- if (or (and (ne (.Values.server.enabled | toString) "-") .Values.server.enabled) (and (eq (.Values.server.enabled | toString) "-") .Values.global.enabled)) }}
-                  # The server is enabled, which we make the assumption that this is a contained environment and  
-                  # the pods do not need to be exposed, and need to run on Pod IP's to talk to the server (also on pod ips)
+                  {{- if not .Values.client.exposeGossipPorts }}
                   fieldPath: status.podIP
                   {{- else }}
-                  # we are not going to be running a server on this cluster, so have the client listen on NodePorts
-                  # so that it can communicate bidirectionally with servers outside the cluster. 
+                  # Clients will be exposed on their node's hostPort for external-to-k8s communication,
+                  # so they need to advertise their host ip instead of their pod ip.
                   fieldPath: status.hostIP
                   {{- end }}
             - name: NAMESPACE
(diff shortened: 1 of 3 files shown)


values.yaml
```diff
@@ -171,6 +171,11 @@ client:
   # This should be set to true if connectInject or meshGateway is enabled.
   grpc: false
 
+  # exposeGossipPorts exposes the clients' gossip ports as hostPorts.
+  # This is only necessary if pod ips in the k8s cluster are not directly
+  # routable and the Consul servers are outside of the k8s cluster.
+  exposeGossipPorts: false
```
lkysow: should this be `exposeGossipPort` (singular) since there's only one port?

lkysow (approved):
Looks great.

One thing I was wondering was if we should expand the value to something like:
```yaml
client:
  gossipPort:
    mapToHost: true # or could do hostPort: 8301 and if it's null or <0 don't map to host
    advertiseAddress: "hostIP"
```

My thinking is that there may be cases in the future where users need something other than `status.hostIP` exposed. For example, if I wanted to expose my node's public IP or it has multiple IPs due to enis, etc. In that case we could simply add a new setting to `advertiseAddress` rather than needing to break backwards compatibility.

values.yaml
```diff
@@ -171,6 +171,11 @@ client:
   # This should be set to true if connectInject or meshGateway is enabled.
   grpc: false
 
+  # exposeGossipPorts exposes the clients' gossip ports as hostPorts.
+  # This is only necessary if pod ips in the k8s cluster are not directly
```
```suggestion
  # This is only necessary if pod IPs in the k8s cluster are not directly
```

values.yaml
```diff
@@ -171,6 +171,11 @@ client:
   # This should be set to true if connectInject or meshGateway is enabled.
   grpc: false
 
+  # exposeGossipPorts exposes the clients' gossip ports as hostPorts.
```
```suggestion
  # exposeGossipPorts exposes the clients' gossip ports as hostPorts
     and clients' IPs as hostIPs.
```

ishustava (approved):
Looks great! I've added a couple of suggestions for comments, but otherwise looks really good :)

values.yaml
marcdeop: There are actually two [ports](https://www.consul.io/docs/agent/options.html#ports) used for Serf. These are `8301` and `8302` for `lan` and `wan` respectively.

lkysow: The wan port is only for the servers though. I see we have it here but it's not used by the agents. This is because servers are the only ones that will gossip outside of their datacenter (i.e. to other datacenters).

test/unit/client-daemonset.bats
```diff
@@ -491,3 +491,54 @@ load _helpers
...
+# client.exposeGossipPorts
+
+@test "client/DaemonSet: client uses podIP when client.exposeGossipPorts=false" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/client-daemonset.yaml  \
+      --set 'client.enabled=true' \
+      --set 'client.exposeGossipPorts=false' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.containers | map(select(.name=="consul")) | .[0].env | map(select(.name=="ADVERTISE_IP")) | .[0] | .valueFrom.fieldRef.fieldPath'  |
```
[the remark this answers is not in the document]
adilyse: All credit to @stvdilln!

values.yaml
adilyse: From these [docs](https://kubernetes.io/docs/concepts/configuration/overview/#services): `each <hostIP, hostPort, protocol> combination must be unique`. Technically we're exposing just 8301, but because we need to expose both tcp and udp as hostPorts, we have to specify them separately. I figured it would be confusing to leave this singular given that there are two definitions.

templates/client-daemonset.yaml
```diff
@@ -163,7 +169,17 @@ spec:
...
+              hostPort: 8301
+              {{- end }}
+              protocol: "TCP"
+              name: serflan-tcp
+            - containerPort: 8301
+              {{- if .Values.client.exposeGossipPorts }}
+              hostPort: 8301
+              {{- end }}
+              protocol: "UDP"
+              name: serflan-udp
```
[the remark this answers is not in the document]
adilyse: This is a quirk of defining `hostPorts` because each <hostIP, hostPort, protocol> combination must be unique. It shouldn't be necessary for the servers, though we could if we wanted to be explicit about the protocols being used.

values.yaml

values.yaml
```diff
@@ -171,6 +171,11 @@ client:
...
+  # This is only necessary if pod ips in the k8s cluster are not directly
+  # routable and the Consul servers are outside of the k8s cluster.
```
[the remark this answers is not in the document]

templates/client-daemonset.yaml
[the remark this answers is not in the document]
lkysow: I was thinking for consistency as well

values.yaml
lkysow: It's the same port though, just different protocols.

It's not a blocker, just when I read "exposes the clients' gossip ports" I think there's two ports being exposed but there isn't, there's just one: 8301.

mozai:
What if the servers are inside the k8s cluster and the clients are outside?

ishustava:
Hi @mozai,

We support that when your out-of-cluster clients can connect to the Consul servers on the Pod IPs. Take a look at [these docs](https://www.consul.io/docs/platform/k8s/out-of-cluster-nodes.html).  

If you would like the Helm chart to support out-of-cluster clients when server Pod IPs aren't routable, then it would be super helpful if you could open an issue describing your use case.

Thanks!

PR #292, opened 2019-11-26
Do not run server-acl-init during server rollout

### LINKED ISSUE #274: StatefulSet partition value update doesn't force server redeployment so it shouldn't block acl bootstrapper job
I like the work that's been done on the bootstrap-acls job but I'm running into one quirk.
When the partition value of the statefulset is upgraded but nothing else in the pod spec, the servers do not redeploy.

This perceived difference in the partition value causes the acl-init job to get stuck indefinitely with the following repeating message:
```
2019-10-28T20:30:16.131Z [ERROR] Failure: waiting for rollout of statefulset vault-consul-consul-server: err="rollout is in progress (CurrentRevision=vault-consul-consul-server-844f64bff6 UpdateRevision=vault-consul-consul-server-86c7b4f99)"
2019-10-28T20:30:16.131Z [INFO ] Retrying in 1s
2019-10-28T20:30:17.135Z [ERROR] Failure: waiting for rollout of statefulset vault-consul-consul-server: err="rollout is in progress (CurrentRevision=vault-consul-consul-server-844f64bff6 UpdateRevision=vault-consul-consul-server-86c7b4f99)"
2019-10-28T20:30:17.135Z [INFO ] Retrying in 1s
...
``` 
In my opinion, the partition value shouldn't block the acl-init job and instead a new condition should be looked for.
Alternative 1: add logic to the job to detect that a rollout isn't actually occurring and that the job should continue (eg: no server redeployments in at least a minute).
Alternative 2: add something to the Consul server (eg: an env) to force redeployment of the consul servers when the partition value is updated.

lkysow:
Prior to this change, running helm upgrade with server.updatePartition >
0 and global.bootstrapACLs=true would cause the command to hang
indefinitely.

It would hang because it waited for the server-acl-init-cleanup hook to
complete. This hook would not complete until the server-acl-init job
finished. This job would not finish because it waits until the server
rollout is complete. The server rollout would not complete because the
updatePartition causes one server at a time to roll out.

In order for the server rollout to complete, the user needs to wait for
the first server to be upgraded, then they need to decrease the
updatePartition by one and then run helm upgrade again and repeat until
all the servers are upgraded.

In order to fix this, this change causes the server-acl-init job to only
run when the updatePartition is 0. This means during a server rollout,
i.e. updatePartition > 0, the job won't run and so the cleanup hook will
exit immediately and the helm upgrade command will exit cleanly.

Once the last server is ready to be upgraded, the updatePartition will
be set to 0. When helm upgrade runs, the server-acl-init Job will be
rendered but it will be able to complete successfully once that final
server comes up because the rollout will be complete.

It's worth noting that the reason we run server-acl-init during helm
upgrade is to help in the case when users have enabled a new feature,
for example mesh gateways, and need acl tokens for that feature.

Fixes #274

### DIFF
--- a/templates/server-acl-init-cleanup-job.yaml
+++ b/templates/server-acl-init-cleanup-job.yaml
@@ -1,5 +1,7 @@
 {{- if (or (and (ne (.Values.server.enabled | toString) "-") .Values.server.enabled) (and (eq (.Values.server.enabled | toString) "-") .Values.global.enabled)) }}
 {{- if .Values.global.bootstrapACLs }}
+{{- /* See reason for this in server-acl-init-job.yaml */ -}}
+{{- if eq (int .Values.server.updatePartition) 0 }}
 # This job deletes the server-acl-init job once it completes successfully.
 # It runs as a helm hook because it only needs to run when the server-acl-init
 # Job gets recreated which only happens during an install or upgrade.
@@ -49,5 +51,6 @@ spec:
             - delete-completed-job
             - -k8s-namespace={{ .Release.Namespace }}
             - {{ template "consul.fullname" . }}-server-acl-init
+  {{- end }}
   {{- end }}
   {{- end }}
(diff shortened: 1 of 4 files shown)


lkysow:
> This looks good. Have you thought about adding the same check to the ACL cleanup job?

Yeah that makes sense. Will do.

PR #302, opened 2019-12-05
Default centralConfig to enabled

lkysow:
This change will only affect users who have enabled connectInject
because we only enable central config if connectInject.enabled is true.
Users that are using connectInject are more likely than not going to
want to create L7 rules and service-defaults configs because that is
where most of the value of Consul Connect lies. Thus enabling this by
default makes sense.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 ## Unreleased
 
+IMPROVEMENTS:
+
+  * `connectInject.centralConfig` defaults to `true` now instead of `false`. This is to make it
+     easier to configure Connect via `service-defaults` and other routing
+     config [[GH-302](https://github.com/hashicorp/consul-helm/pull/302)].
+     See https://www.consul.io/docs/agent/options.html#enable_central_service_config.
+
+     If you wish to disable central config, set `connectInject.centralConfig` to
+     false in your local values file. NOTE: If `connectInject.enabled` is false,
+     then central config is not enabled so this change will not affect you. 
+
 ## 0.14.0 (Dec 10, 2019)
 
 IMPROVEMENTS:
--- a/values.yaml
+++ b/values.yaml
@@ -504,6 +504,9 @@ connectInject:
     # proxyDefaults is a raw json string that will be written as the value of
     # the "config" key of the global proxy-defaults config entry.
     # See: https://www.consul.io/docs/agent/config-entries/proxy-defaults.html
+    # NOTE: Changes to this value after the chart is first installed have *no*
+    # effect. In order to change the proxy-defaults config after installation,
+    # you must use the Consul API.
     proxyDefaults: |
       {}
 
--- a/values.yaml
+++ b/values.yaml
@@ -493,6 +493,10 @@ connectInject:
 
   # Requires Consul >= v1.5 and consul-k8s >= v0.8.1.
   centralConfig:
+    # enabled controls whether central config is enabled on all servers and clients.
+    # See https://www.consul.io/docs/agent/options.html#enable_central_service_config.
+    # If changing this after installation, servers and clients must be restarted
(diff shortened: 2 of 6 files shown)


values.yaml
```diff
@@ -454,18 +454,17 @@ connectInject:
 
   # Requires Consul >= v1.5 and consul-k8s >= v0.8.1.
   centralConfig:
-    enabled: false
+    enabled: true
 
     # defaultProtocol allows you to specify a convenience default protocol if
     # most of your services are of the same protocol type. The individual annotation
-    # on any given pod will override this value. A protocol must be provided,
```
lkysow: I'm going to confirm this with the team whether protocol must be provided.

lkysow: It does not.

CHANGELOG.md
```diff
@@ -3,10 +3,19 @@
...
      harder to misconfigure Connect [[GH-282](https://github.com/hashicorp/consul-helm/pull/282)]
-     
+
      If you do not wish to enable gRPC for clients, set `client.grpc` to
      `false` in your local values file.
 
+  * `connectInject.centralConfig` defaults to `true` now instead of `false`. This is to make it
+     easier to configure Connect via `service-defaults` and other routing
+     config [[GH-302](https://github.com/hashicorp/consul-helm/pull/302)]
+     See https://www.consul.io/docs/agent/options.html#enable_central_service_config
```
ishustava:
```suggestion
     See [Consul Docs](https://www.consul.io/docs/agent/options.html#enable_central_service_config).
```

CHANGELOG.md
```diff
@@ -3,10 +3,19 @@
...
   * `client.grpc` defaults to `true` now instead of `false`. This is to make it
      harder to misconfigure Connect [[GH-282](https://github.com/hashicorp/consul-helm/pull/282)]
-     
+
      If you do not wish to enable gRPC for clients, set `client.grpc` to
      `false` in your local values file.
 
+  * `connectInject.centralConfig` defaults to `true` now instead of `false`. This is to make it
+     easier to configure Connect via `service-defaults` and other routing
+     config [[GH-302](https://github.com/hashicorp/consul-helm/pull/302)]
```
ishustava:
```suggestion
     config [[GH-302](https://github.com/hashicorp/consul-helm/pull/302)].
```

CHANGELOG.md
```diff
@@ -3,10 +3,19 @@
 BREAKING CHANGES
   * `client.grpc` defaults to `true` now instead of `false`. This is to make it
      harder to misconfigure Connect [[GH-282](https://github.com/hashicorp/consul-helm/pull/282)]
-     
+
      If you do not wish to enable gRPC for clients, set `client.grpc` to
      `false` in your local values file.
 
+  * `connectInject.centralConfig` defaults to `true` now instead of `false`. This is to make it
```
ishustava: This probably needs to be updated because we've released a couple of times.

ishustava (approved):
This looks good, Luke. One question I had is whether there's any impact if you upgrade from `centralConfig.enabled` set to `false`.

CHANGELOG.md
lkysow: This will be converted to a link by github anyway and I thought it best to make it clear where users are being linked to.

CHANGELOG.md
lkysow: ✅

lkysow:
> One question I had is whether there's any impact if you upgrade from centralConfig.enabled set to false.

It will redeploy the injector since it will add the `-enable-central-config=true` flag. It will only update the configmaps for client and server which won't affect anything. According to our reload docs (https://www.consul.io/docs/agent/options.html#reloadable-configuration) running `consul reload` won't pick up the change either. Thus I think this would require restart all clients and servers to take effect.

PR #304, opened 2019-12-06
Update values docs to match website

lkysow:
This PR is a companion to: https://github.com/hashicorp/consul/pull/6891

It ensures they both have the same documentation for the helm values. I've only did the `global` key. This is what happens when you order a coffee and *then* find out there's no WiFi.

I've also updated some out of date info.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -1,55 +1,75 @@
 # Available parameters and their default values for the Consul chart.
 
+# global holds values that affect multiple components of the chart.
 global:
-  # enabled is the master enabled switch. Setting this to true or false
-  # will enable or disable all the components within this chart by default.
-  # Each component can be overridden using the component-specific "enabled"
-  # value.
+  # enabled is the master enabled/disabled setting.
+  # If true, servers, clients, Consul DNS and the Consul UI will be enabled.
+  # Each component can override this default via its component-specific
+  # "enabled" config.
+  # If false, no components will be installed by default and per-component
+  # opt-in is required, such as by setting `server.enabled` to true.
   enabled: true
 
-  # Domain to register the Consul DNS server to listen for.
+  # domain is the domain Consul will answer DNS queries for
+  # (see https://www.consul.io/docs/agent/options.html#_domain) and the domain
+  # services synced from Consul into Kubernetes will have,
+  # e.g. `service-name.service.consul`.
   domain: consul
 
-  # Image is the name (and tag) of the Consul Docker image for clients and
-  # servers below. This can be overridden per component.
+  # image is the name (and tag) of the Consul Docker image for clients and
+  # servers. This can be overridden per component.
+  # This should be pinned to a specific version tag, otherwise you may
+  # inadvertently upgrade your Consul version.
   #
   # Examples:
+  #   # Consul 1.5.0
   #   image: "consul:1.5.0"
-  #   image: "hashicorp/consul-enterprise:1.5.0-ent"   # Enterprise Consul image
+  #   # Consul Enterprise 1.5.0
+  #   image: "hashicorp/consul-enterprise:1.5.0-ent"
   image: "consul:1.6.2"
 
   # imageK8S is the name (and tag) of the consul-k8s Docker image that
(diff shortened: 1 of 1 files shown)


values.yaml
```diff
@@ -1,55 +1,75 @@
 # Available parameters and their default values for the Consul chart.
 
+# global holds values that affect multiple components of the chart.
 global:
-  # enabled is the master enabled switch. Setting this to true or false
-  # will enable or disable all the components within this chart by default.
-  # Each component can be overridden using the component-specific "enabled"
-  # value.
+  # enabled is the master enabled/disabled setting.
+  # If true, servers, clients, Consul DNS and the Consul UI will be enabled.
```
lkysow: specified exactly which components will be enabled because it previously said they'd all be enabled but things like meshgateway, connect wouldn't be so thought it best to be clear

values.yaml
```diff
@@ -1,55 +1,75 @@
...
   image: "consul:1.6.2"
 
   # imageK8S is the name (and tag) of the consul-k8s Docker image that
-  # is used for functionality such as the catalog sync. This can be overridden
-  # per component below.
+  # is used for functionality such as catalog sync. This can be overridden
+  # per component.
   # Note: support for the catalog sync's liveness and readiness probes was added
-  # to consul-k8s v0.6.0. If using an older consul-k8s version, you may need to
+  # to consul-k8s 0.6.0. If using an older consul-k8s version, you may need to
```
lkysow: our versions don't have `v` so this will probably save someone typing it wrong

values.yaml
```diff
@@ -1,55 +1,75 @@
...
-  # key with the "consul keygen" command.
-  # See https://www.consul.io/docs/commands/keygen.html
+  # gossipEncryption configures which Kubernetes secret to retrieve Consul's
+  # gossip encryption key from (see https://www.consul.io/docs/agent/options.html#_encrypt).
+  # If secretName or secretKey are not set, gossip encryption will not be enabled.
+  # The secret must be in the same namespace that Consul is installed into.
+  #
+  # The secret can be created by running:
+  #    kubectl create secret generic consul-gossip-encryption-key \
+  #      --from-literal=key=$(consul keygen)`.
```
lkysow: There was at least one issue of someone not creating the secret correctly so thought it best to document a way

values.yaml
ishustava: You could also set `syncCatalog.enabled`, `connectInject.enabled` (didn't check others) to `-`, and they will respect the `global.enabled` value.

values.yaml
```diff
@@ -1,55 +1,75 @@
...
+  # as. This can't be changed once the Consul cluster is up and running
   # since Consul doesn't support an automatic way to change this value
-  # currently: https://github.com/hashicorp/consul/issues/1858
+  # currently: https://github.com/hashicorp/consul/issues/1858.
   datacenter: dc1
 
-  # enablePodSecurityPolicies is a boolean flag that controls whether pod
-  # security policies are created for the consul components created by this
-  # chart. See https://kubernetes.io/docs/concepts/policy/pod-security-policy/
+  # enablePodSecurityPolicies controls whether Pod
```
ishustava: Should be lower case because it's a feature name. It's not consistent, but they use lower case in [k8s docs](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#policy-order) too.
```suggestion
  # enablePodSecurityPolicies controls whether pod
```

values.yaml
```diff
@@ -1,55 +1,75 @@
...
+  # Security Policies are created for the Consul components created by this
```
```suggestion
  # security policies are created for the Consul components created by this
```

values.yaml
lkysow: True. I just tried to document that here
>   # connectInject and syncCatalog's enabled keys can be set to "-" to inherit from this value as well

But it didn't really fit imo. If we have those docs on those keys do you think that's sufficient?

lkysow: I'll merge this and update based on your response.

lkysow: I'm going to merge this as is but come back and update based on your response.

PR #305, opened 2019-12-06
Update CHANGELOG.md

lkysow:
Add a bit more information to what's going to happen. I think it's easier for users in this case because they don't need to click into the PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,13 @@ BREAKING CHANGES:
     If you do not wish to enable gRPC for clients, set `client.grpc` to
     `false` in your local values file.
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -25,7 +27,6 @@ IMPROVEMENTS:
 
 BUG FIXES:
 
-  * Add `connectInject.overrideAuthMethodName` to allow setting the `-acl-auth-method flag` [[GH-278](https://github.com/hashicorp/consul-helm/pull/278)]
```
lkysow: This was already listed as an improvement (which I think it is rather than a bugfix)

PR #307, opened 2019-12-10
Support global.name

### LINKED ISSUE #286: fullnameOverride prevents acl job from completing successfully
Companion issue reported in hashicorp/consul-k8s#165

Given the following configuration values.yaml provided to helm:
```
fullnameOverride: consul
global:
  bootstrapACLs: true
```

The StatefulSet will be named `consul-server` and the bootstrap job will fail

lkysow:
There was a previously undocumented setting `fullnameOverride` that
would allow controlling the prefix added to each resource. There were
some places where this wasn't being used. Specifically in our
clusterrole's and when running server-acl-init.

This change fixes those locations and utilizes a new flag in
server-acl-init that will respect the override.

It also adds a new value: `global.name` and documents this
value. This value is the same as the `fullnameOverride` value, but it
lives under the global key which fits with the rest of our config
management philosophy (i.e. config that affects the whole chart lives
under the global key).

* Fixes https://github.com/hashicorp/consul-helm/issues/286
* Corresponding consul-k8s PR: https://github.com/hashicorp/consul-k8s/pull/174

### DIFF
--- a/templates/server-acl-init-job.yaml
+++ b/templates/server-acl-init-job.yaml
@@ -45,7 +45,8 @@ spec:
             - "-ec"
             - |
               consul-k8s server-acl-init \
-                -release-name={{ .Release.Name }} \
+                -server-label-selector=component=server,app={{ template "consul.name" . }},release={{ .Release.Name }} \
+                -resource-prefix={{ template "consul.fullname" . }} \
                 -k8s-namespace={{ .Release.Namespace }} \
                 {{- if .Values.syncCatalog.enabled }}
                 -create-sync-token=true \
--- a/values.yaml
+++ b/values.yaml
@@ -38,7 +38,7 @@ global:
   # Note: support for the catalog sync's liveness and readiness probes was added
   # to consul-k8s 0.6.0. If using an older consul-k8s version, you may need to
   # remove these checks to make the sync work.
-  # If using mesh gateways and bootstrapACLs then must be >= 0.9.0.
+  # If using bootstrapACLs then must be >= 0.10.1.
   imageK8S: "hashicorp/consul-k8s:0.9.5"
 
   # datacenter is the name of the datacenter that the agents should register
@@ -73,7 +73,7 @@ global:
 
   # bootstrapACLs will automatically create and assign ACL tokens within
   # the Consul cluster. This requires servers to be running inside Kubernetes.
-  # Additionally requires Consul >= 1.4 and consul-k8s >= 0.8.0.
+  # Additionally requires Consul >= 1.4 and consul-k8s >= 0.10.1.
   bootstrapACLs: false
 
 # Server, when enabled, configures a server cluster to run. This should
--- a/values.yaml
+++ b/values.yaml
@@ -10,6 +10,10 @@ global:
   # opt-in is required, such as by setting `server.enabled` to true.
   enabled: true
 
+  # name sets the prefix used for all resources in the helm chart.
+  # If not set, the prefix will be "<helm release name>-consul".
+  name: null
+
   # domain is the domain Consul will answer DNS queries for
   # (see https://www.consul.io/docs/agent/options.html#_domain) and the domain
   # services synced from Consul into Kubernetes will have,
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -1,12 +1,14 @@
 {{/*
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to
-this (by the DNS naming spec). If release name contains chart name it will
-be used as a full name.
+this (by the DNS naming spec). Supports the legacy fullnameOverride setting
+as well as the global.name setting.
 */}}
 {{- define "consul.fullname" -}}
 {{- if .Values.fullnameOverride -}}
 {{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- else if .Values.global.name -}}
+{{- .Values.global.name | trunc 63 | trimSuffix "-" -}}
 {{- else -}}
 {{- $name := default .Chart.Name .Values.nameOverride -}}
 {{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
--- a/templates/client-clusterrole.yaml
+++ b/templates/client-clusterrole.yaml
@@ -23,7 +23,7 @@ rules:
     resources:
       - secrets
     resourceNames:
-      - {{ .Release.Name }}-consul-client-acl-token
+      - {{ template "consul.fullname" . }}-client-acl-token
     verbs:
       - get
 {{- end }}
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -228,7 +228,7 @@ spec:
           - "-ec"
           - |
             consul-k8s acl-init \
-              -secret-name="{{ .Release.Name }}-consul-client-acl-token" \
+              -secret-name="{{ template "consul.fullname" . }}-client-acl-token" \
               -k8s-namespace={{ .Release.Namespace }} \
               -init-type="client"
         volumeMounts:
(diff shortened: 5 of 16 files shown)


templates/server-acl-init-job.yaml
```diff
@@ -46,6 +46,7 @@ spec:
             - |
               consul-k8s server-acl-init \
                 -release-name={{ .Release.Name }} \
+                -resource-prefix={{ template "consul.fullname" . }} \
```
lkysow: Here is the use of the new flag.

lkysow:
@hashicorp/consul-eco-platform looking for feedback on some things before submitting:
1. Creating a new key under `global` called `nameOverride` that works the same as `fullnameOverride` but is in (what I think) the correct location under `global`.
1. Not testing this. We could test
  a) that all the resources have the expected names with the default, with fullNameOverride and with `global.nameOverride` and the corresponding precedence however this would add a lot of boilerplate testing to every single resource. We previously weren't testing this since it simply tests the `consul.fullname` template, over and over again. Instead I propose we test that template in isolation once.
  b) that the clusterroles target the right `resourceNames`. This wasn't being tested previously. I personally think this is too much detail in the tests and again would be testing the consul.fullname template. On the other hand, if we had tested this previously it would have found the bugs that this  PR fixes (where it used the ReleaseName directly instead of the consul.fullname template)

ishustava:
@lkysow

As a relatively new Helm user, I find that putting it under global is more intuitive. I think it's good that it's documented and exposed rather than hidden in the `helpers` file. 

However, it seems that `fullnameOverride` is some sort of convention [among the official helm charts](https://github.com/helm/charts/search?q=fullnameOverride&unscoped_q=fullnameOverride) and changing that would be confusing to more experienced Helm users, assuming we are going to deprecate it eventually.

With regards to the testing strategy:
a) I think we'd also need to add tests to make sure all templates are using the helper function for names rather than constructing them from release name.
b) I think it would be good to add tests for it. I was thinking about improving clusterrole tests because currently they only assert on the resource type, e.g. `secrets`, which could be ambiguous.

lkysow:
> However, it seems that fullnameOverride is some sort of convention among the official helm charts and changing that would be confusing to more experienced Helm users, assuming we are going to deprecate it eventually.

I wasn't planning on deprecating it.

> a) I think we'd also need to add tests to make sure all templates are using the helper function for names rather than constructing them from release name.

So we use `{{ template "consul.fullname" . }}` 109 times in our chart files. We'd need to test the default, `global.name` and `fullnameOverride` so 327 tests. That seems like overkill to test this that many times?

ishustava:
If we're not deprecating it, then I'd prefer having just one way of overriding the name. We could still document it.

Could we instead write one comprehensive test, something like a grep command to check that templates aren't using `{{ .Release.Name }}`. 

```
$ grep -r '{{ .Release.Name }}' templates/*.yaml | grep -v 'release: '
templates/client-clusterrole.yaml:      - {{ .Release.Name }}-consul-client-acl-token
templates/client-daemonset.yaml:              -secret-name="{{ .Release.Name }}-consul-client-acl-token" \
templates/client-snapshot-agent-clusterrole.yaml:      - {{ .Release.Name }}-consul-client-snapshot-agent-acl-token
templates/client-snapshot-agent-deployment.yaml:                  name: "{{ .Release.Name }}-consul-client-snapshot-agent-acl-token"
templates/client-snapshot-agent-deployment.yaml:              -secret-name="{{ .Release.Name }}-consul-client-snapshot-agent-acl-token" \
templates/connect-inject-deployment.yaml:                -acl-auth-method="{{ .Release.Name }}-consul-k8s-auth-method" \
templates/enterprise-license-clusterrole.yaml:      - {{ .Release.Name }}-consul-enterprise-license-acl-token
templates/enterprise-license.yaml:                  name: "{{ .Release.Name }}-consul-enterprise-license-acl-token"
templates/enterprise-license.yaml:              -secret-name="{{ .Release.Name }}-consul-enterprise-license-acl-token" \
templates/mesh-gateway-clusterrole.yaml:      - {{ .Release.Name }}-consul-mesh-gateway-acl-token
templates/mesh-gateway-deployment.yaml:                -secret-name="{{ .Release.Name }}-consul-mesh-gateway-acl-token" \
templates/mesh-gateway-deployment.yaml:                  name: "{{ .Release.Name }}-consul-mesh-gateway-acl-token"
templates/server-acl-init-job.yaml:                -release-name={{ .Release.Name }} \
templates/sync-catalog-clusterrole.yaml:      - {{ .Release.Name }}-consul-catalog-sync-acl-token
templates/sync-catalog-deployment.yaml:                  name: "{{ .Release.Name }}-consul-catalog-sync-acl-token"
templates/sync-catalog-deployment.yaml:              -secret-name="{{ .Release.Name }}-consul-catalog-sync-acl-token"
```
I ran this command on master, so it doesn't have your recent changes.

templates/server-acl-init-job.yaml
```diff
@@ -45,7 +45,8 @@ spec:
             - "-ec"
             - |
               consul-k8s server-acl-init \
-                -release-name={{ .Release.Name }} \
+                -server-label-selector=component=server,app={{ template "consul.name" . }},release={{ .Release.Name }} \
```
lkysow: @ishustava your idea of using the label selector already fixed a bug! Previously we were assuming the `app` label was always `consul` however it can actually be overridden by the `nameOverride` config value! 🍻

lkysow:
@ishustava okay I've added tests for the `consul.fullname` helper *and* used your sneaky grep as a test to ensure we're only using `.Release.Name` (outside of as a label) where we expect. The idea  is that if someone uses it where `template consul.fullname` should be used, that test will fail. If they actually need to use `.Release.Name`, they can update the expected result of the `grep`.

Here's my reasoning for `global.name`: I think it makes sense to always recommend setting this in our docs and guides. As we talked about above, this is the right place for this setting so it will make sense to users when we recommend it. We could remove the `fullnameOverride` setting, however that would break backwards compatibility. It doesn't cost us much to continue to support that so I think it's worth supporting it.

lkysow:
I've also bumped consul-k8s to 0.10.0 in anticipation that that's the version we'll release. I'll cut a release of that before merging this.

templates/server-acl-init-job.yaml
ishustava: Nice!

values.yaml
```diff
@@ -34,7 +38,7 @@ global:
   # Note: support for the catalog sync's liveness and readiness probes was added
   # to consul-k8s 0.6.0. If using an older consul-k8s version, you may need to
   # remove these checks to make the sync work.
-  # If using mesh gateways and bootstrapACLs then must be >= 0.9.0.
+  # If using bootstrapACLs then must be >= 0.10.1.
```
ishustava: Does it have to be? it's only if you're overriding the name, right? Also, should we keep the note about mesh gateways?

values.yaml
```diff
@@ -10,6 +10,10 @@ global:
   # opt-in is required, such as by setting `server.enabled` to true.
   enabled: true
 
+  # name sets the prefix used for all resources in the helm chart.
+  # If not set, the prefix will be "${helm release name}-consul".
```
ishustava: Should the notation for this be `{{ .Release.Name }}-consul` or `<helm-release-name>-consul`? Also, if this is not set, you could still set `fullNameOverride`.

ishustava (commented):
Luke, thanks for making the improvements. I totally hear your point about `global.name` being less confusing. If it's low cost for us to maintain both, then I think it'd be a good improvement.

I've added a couple of comments about documentation, but otherwise it looks good.

values.yaml
lkysow: It has to be because we're setting the new flags which won't work with the older consul-k8s

lkysow: I removed mesh gateways because previously it was only if you were using both, not it's if you're using acls at all which is inclusive of acls and mesh gateways

values.yaml
lkysow: sure, I'll check what we use throughout the file (if anything).

> Also, if this is not set, you could still set fullNameOverride

True but thought it best not to mention it!

ishustava: It's like our secret menu 😄

ishustava (approved):

PR #313, opened 2019-12-13
Enable TLS

### LINKED ISSUE #136: Support TLS encryption

ishustava:
This PR builds on the work done in #137 and proposes the following changes:

* Optionally allows you to turn on TLS by setting `global.tls.enabled` to `true`.
* Exposes the default HTTPS port 8501 on clients and servers if TLS enabled.
* Optionally allows you to disable HTTP ports on clients and servers by setting `global.tls.httpsOnly` to `true`. This is set to `true` by default.
* Adds TLS bootstrapping job that generates Consul cluster CA and server certificate.
* Adds `tls-init-cleanup` job that deletes certs created by the `tls-init` job when the Helm chart is deleted.
* Consul client certs are generated in an init container, so that we can add `HOST_IP` as a SAN to enable other clients, e.g. sync-catalog, to talk to it. This will be replaced by `auto_encrypt` in the future once it supports providing additional SANs to the client certificate.
* Doesn't expose `verify_*` config options in the Helm chart as it was proposed in the original PR #137. Instead, it assumes secure defaults if `global.tls.enabled` is `true`.

Note 1: This currently doesn't support external CAs. It also doesn't work if servers aren't deployed on k8s. Once we support external CAs, we can also support various deployment configurations. Hence, if servers are disabled, `tls-init` will not run.

Note 2: We're not using mTLS for HTTP API. That is because in [Consul's threat model](https://www.consul.io/docs/internals/security.html#secure-configuration) we recommend using ACLs instead. To prevent clients from sending unencrypted RPC requests to the servers, we set `verify_incoming_rpc` to `true`.

Fixes #136


templates/client-daemonset.yaml
```diff
@@ -71,6 +71,14 @@ spec:
         - name: config
           configMap:
             name: {{ template "consul.fullname" . }}-client-config
+        {{- if .Values.global.tls.enabled }}
+        - name: tls-ca-cert
+          secret:
+            secretName: {{ template "consul.fullname" . }}-ca-cert
+        - name: tls-client-cert
+          emptyDir:
+            medium: "Memory"
```
ishustava: Using `tmpfs` to store client certs.

values.yaml
```diff
@@ -52,6 +52,35 @@ global:
   # consul-k8s v0.8.0+.
   bootstrapACLs: false
 
+  # Enables TLS encryption across the cluster to verify authenticity of the
+  # servers and clients that connect. Note: It is HIGHLY recommended that you also
+  # enable Gossip encryption.
```
lkysow: This should really link to our own docs page on enabling gossip encryption... if we had one.

templates/client-daemonset.yaml
lkysow: fancy! Do you know which versions of kube that's supported with?

templates/tls-init-job.yaml
```diff
@@ -0,0 +1,65 @@
...
+              curl -s -X POST --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
+                https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/${NAMESPACE}/secrets \
+                -H "Authorization: Bearer $( cat /var/run/secrets/kubernetes.io/serviceaccount/token )" \
+                -H "Content-Type: application/json" \
+                -H "Accept: application/json" \
+                -d "{ \"kind\": \"Secret\", \"apiVersion\": \"v1\", \"metadata\": { \"name\": \"{{ template "consul.fullname" . }}-server-cert\", \"namespace\": \"${NAMESPACE}\" }, \"type\": \"kubernetes.io/tls\", \"data\": { \"tls.crt\": \"$( cat {{ .Values.global.datacenter }}-server-{{ .Values.global.domain }}-0.pem | base64 | tr -d '\n' )\", \"tls.key\": \"$( cat {{ .Values.global.datacenter }}-server-{{ .Values.global.domain }}-0-key.pem | base64 | tr -d '\n' )\" } }" > /dev/null
```
lkysow: There's one cert for all the servers right? So isn't this weird it's suffixed with -0?

templates/tls-init-cleanup-serviceaccount.yaml
```diff
@@ -0,0 +1,16 @@
...
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "consul.fullname" . }}-tls-init-cleanup
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
```
lkysow: missing `annotations:`

templates/tls-init-cleanup-clusterrole.yaml
```diff
@@ -0,0 +1,25 @@
...
+kind: ClusterRole
+metadata:
+  name: {{ template "consul.fullname" . }}-tls-init-cleanup
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  "helm.sh/hook": post-delete
```
lkysow: need annotations:

templates/tls-init-cleanup-clusterrolebinding.yaml
```diff
@@ -0,0 +1,24 @@
...
+kind: ClusterRoleBinding
+metadata:
+  name: {{ template "consul.fullname" . }}-tls-init-cleanup
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  "helm.sh/hook": post-delete
```
lkysow: annotations:

ishustava:
This PR requires `connectInject` to support TLS before merging.

templates/client-daemonset.yaml
ishustava: good question! It's supported in 1.13+, don't know about earlier versions.

templates/tls-init-job.yaml
ishustava: That is what consul CLI does by default. It doesn't matter so much though because we're uploading only the contents here. 

It's maybe weird for client certs. Each client cert has the -0 suffix, and so it could be a bit confusing. We could rename them, but with auto_encrypt we won't need this anymore.

templates/tls-init-cleanup-serviceaccount.yaml
ishustava: Good catch! That's funny, it looks like they are not needed here actually. So will remove them altogether.

test/unit/client-daemonset.bats
```diff
@@ -418,9 +418,205 @@ load _helpers
       --set 'global.gossipEncryption.secretName=bar' \
       . | tee /dev/stderr |
       yq '.spec.template.spec.containers[] | select(.name=="consul") | .command | join(" ") | contains("encrypt")' | tee /dev/stderr)
+  [ "${actual}" == "true" ]
+}
+
+#--------------------------------------------------------------------
+# global.tls.enabled
+
+@test "client/DaemonSet: no secret volumes when TLS is disabled" {
```
lkysow: seems fragile, what if I add a secret for something unrelated? can you test just that it doesn't have the tls volumes? Or... delete this test since checking the absence of things for each feature seems like overkill

test/unit/client-daemonset.bats
```diff
@@ -418,9 +418,205 @@ load _helpers
...
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/client-daemonset.yaml  \
+      --set 'global.tls.enabled=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.volumes[] | select(.name == "tls-client-cert")' | tee /dev/stderr)
+  [ "${actual}" != "" ]
+}
+
+@test "client/DaemonSet: no volumes mounted when TLS is disabled" {
```
lkysow: see same comment above

test/unit/client-daemonset.bats
```diff
@@ -418,9 +418,205 @@ load _helpers
...
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/client-daemonset.yaml  \
+      --set 'global.tls.enabled=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].volumeMounts[] | select(.name == "tls-client-cert")' | tee /dev/stderr)
+  [ "${actual}" != "" ]
+}
+
+@test "client/DaemonSet: port 8500 is exposed when TLS is disabled" {
```
lkysow: this should just be "by default"

test/unit/client-daemonset.bats
```diff
@@ -418,9 +418,205 @@ load _helpers
...
+
+@test "client/DaemonSet: init container is created when global.tls.enabled=true" {
+  cd `chart_dir`
+  local object=$(helm template \
+      -x templates/client-daemonset.yaml  \
+      --set 'global.tls.enabled=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.initContainers[] | select(.name == "client-tls-init")' | tee /dev/stderr)
+
+  # test that it generates client certificate
```
lkysow: seems like overkill, e.g. why not test the exact command? i think testing the init container exists is enough

values.yaml
```diff
@@ -52,6 +52,35 @@ global:
...
+  # See https://learn.hashicorp.com/consul/security-networking/agent-encryption
+  #
+  # Note: this relies on functionality introduced with Consul 1.4.1. Make sure
+  # your global.image value is at least version 1.4.1.
+  tls:
+    enabled: false
+
+    # Set serverAdditionalDNSSANs is a list of additional DNS names to
```
lkysow: remove the word `Set` from these

values.yaml
```diff
@@ -52,6 +52,35 @@ global:
...
+    # for example, if you're using the UI.
+    serverAdditionalDNSSANs: []
+
+    # Set serverAdditionalIPSANs is a list of additional IP addresses to
+    # set as Subject Alternative Names (SANs) in the server certificate.
+    # This is useful when you need to access Consul server(s) externally,
+    # for example, if you're using the UI.
+    serverAdditionalIPSANs: []
+
+    # If httpsOnly is true, Consul will disable the HTTP port and only accept
```
lkysow: not clear if this is across server, clients or both

values.yaml
```diff
@@ -52,6 +52,35 @@ global:
...
   bootstrapACLs: false
 
+  # Enables TLS encryption across the cluster to verify authenticity of the
+  # servers and clients that connect. Note: It is HIGHLY recommended that you also
+  # enable Gossip encryption.
+  # See https://learn.hashicorp.com/consul/security-networking/agent-encryption
+  #
+  # Note: this relies on functionality introduced with Consul 1.4.1. Make sure
+  # your global.image value is at least version 1.4.1.
+  tls:
```
lkysow: https://www.youtube.com/watch?v=dAVyEFK1Cs4&t=33s

lkysow (commented):
I still need to look at
* UI
* how it interacts with ACLs
* how we might update the certs

but thought it best to add this feedback so far

lkysow:
Outstanding question from me: How do users rotate their certificates?

values.yaml
ishustava: 🤣

templates/client-daemonset.yaml
```diff
@@ -126,8 +133,20 @@ spec:
                 -advertise="${ADVERTISE_IP}" \
                 -bind=0.0.0.0 \
                 -client=0.0.0.0 \
+                {{- if .Values.global.tls.enabled }}
```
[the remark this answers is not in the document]
ishustava: Yeah, I think it would be nice. My only reservation against it is that it currently requires a manual restart. So disabling/enabling through the Helm chart on an existing cluster won't do anything.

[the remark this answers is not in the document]
lkysow: Oh good catch. Okay.

templates/tls-init-job.yaml
```diff
@@ -0,0 +1,65 @@
...
+              consul tls cert create -server \
+                {{- range .Values.global.tls.serverAdditionalIPSANs }}
+                -additional-ipaddress={{ . }} \
+                {{- end }}
+                {{- range .Values.global.tls.serverAdditionalDNSSANs }}
+                 -additional-dnsname={{ . }} \
+                 {{- end }}
+                -dc={{ .Values.global.datacenter }} \
+                -domain={{ .Values.global.domain }}
+              curl -s -X POST --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
```
lkysow: does it ever make sense for the CA to be a separate secret from the key? So I can give access to the CA for validation without also giving the key, i.e. in rbac `resourceNames`

templates/client-daemonset.yaml
```diff
@@ -177,9 +205,16 @@ spec:
                 - -c
                 - consul leave
```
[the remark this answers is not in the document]
ishustava: I'm so glad you asked this! After some testing, I've found that consul clients already have `leave_on_terminate` set to true by default (as of consul 0.7) according to the [docs](https://www.consul.io/docs/agent/options.html#leave_on_terminate). I'll set that value to true for clients, so that it works even for an earlier version. That way we can delete the preStop hook.

For the servers though, the `preStop` hook doesn't work currently even without TLS. If I delete server pods, I'm seeing in the logs that it exits 1 instead of 0. I think it's because servers set `terminationGracePeriodSeconds` to 10s and so there's not enough time for it to properly exit. Unfortunately, leave_on_terminate has a hard-coded grace period of 15seconds which isn't enough for the servers. I'll keep the `preStop` hook there and increase `terminationGracePeriodSeconds` to 30sec (default).

CHANGELOG.md
```diff
@@ -11,6 +11,15 @@ IMPROVEMENTS:
      false in your local values file. NOTE: If `connectInject.enabled` is false,
      then central config is not enabled so this change will not affect you. 
 
+  * Optionally allow enabling TLS for servers and clients [[GH-313](https://github.com/hashicorp/consul-helm/pull/313/files#)].
+
+    Note that consul-k8s components don't currently work with HTTPS
```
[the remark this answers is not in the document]
ishustava: I'll fix the injector before this PR gets merged.

templates/tls-init-job.yaml
```diff
@@ -0,0 +1,65 @@
...
+                fieldRef:
+                  fieldPath: metadata.namespace
+          command:
+            - "/bin/sh"
+            - "-ec"
+            - |
+              consul tls ca create
+              consul tls cert create -server \
+                {{- range .Values.global.tls.serverAdditionalIPSANs }}
+                -additional-ipaddress={{ . }} \
```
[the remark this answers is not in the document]
ishustava: Yeah I think that's a good idea.

templates/tls-init-cleanup-job.yaml
```diff
@@ -0,0 +1,48 @@
...
+kind: Job
+metadata:
+  name: {{ template "consul.fullname" . }}-tls-init-cleanup
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  annotations:
+    "helm.sh/hook": pre-delete
```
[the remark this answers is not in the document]
ishustava: No, because I want to make sure the resources it needs, like service account, cluster role etc, are still there. Post-delete executes after helm deletes all the resources.

templates/tls-init-clusterrole.yaml
```diff
@@ -0,0 +1,23 @@
...
+metadata:
+  name: {{ template "consul.fullname" . }}-tls-init
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  annotations:
+    "helm.sh/hook": pre-install
```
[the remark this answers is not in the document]
ishustava: You didn't need to because acl-init is a post-install,post-upgrade hook, and so it's executed after all resources are created. This is not the case for pre-install hooks. They are executed before any resources are created by Helm, so we need to mark these too to make sure they are there.

templates/tls-init-clusterrole.yaml
```diff
@@ -0,0 +1,23 @@
...
+    "helm.sh/hook-delete-policy": before-hook-creation
+rules:
+- apiGroups: [""]
+  resources:
```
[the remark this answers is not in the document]
ishustava: Unfortunately, I can't. Kubernetes [doesn't support](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources) `resourceNames` for `create` requests.

templates/client-daemonset.yaml
```diff
@@ -235,6 +277,33 @@ spec:
...
+          - "/bin/sh"
+          - "-ec"
+          - |
+            cd /consul/tls/client
+            consul tls cert create -client \
+              -dc={{ .Values.global.datacenter }} \
+              -domain={{ .Values.global.domain }} \
+              -additional-ipaddress=${HOST_IP} \
+              -ca=/consul/tls/ca/tls.crt \
+              -key=/consul/tls/ca/tls.key
```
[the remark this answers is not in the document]
ishustava: Hmm, that's an interesting question. For clients, these will be regenerated every time clients are restarted. So hopefully you restart your clients at least once in a year 🤞 

For servers, it might make sense to extend the validity period since those secrets are generated only once. We can extend those to two years.

Yeah, I could definitely rename the files :)

templates/client-daemonset.yaml
```diff
@@ -235,6 +277,33 @@ spec:
...
+              fieldPath: status.hostIP
+        command:
+          - "/bin/sh"
+          - "-ec"
+          - |
+            cd /consul/tls/client
+            consul tls cert create -client \
+              -dc={{ .Values.global.datacenter }} \
+              -domain={{ .Values.global.domain }} \
+              -additional-ipaddress=${HOST_IP} \
```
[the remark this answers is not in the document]
ishustava: What is the use case for it? If we need it later for some reason, we could just add it to the cert, and it will get rotated on restart/upgrade. Eventually, I think we want to switch away from host IPs in favor of host-local traffic whenever kube supports it. In that case, we'd need to add a service name I think, but not sure what the actual implementation of it will be.

templates/server-statefulset.yaml
```diff
@@ -99,6 +107,18 @@ spec:
                 -advertise="${POD_IP}" \
                 -bind=0.0.0.0 \
                 -bootstrap-expect={{ .Values.server.bootstrapExpect }} \
+                {{- if .Values.global.tls.enabled }}
```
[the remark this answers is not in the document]
ishustava: I think we already discussed this for the clients. Will resolve this.

templates/tls-init-cleanup-clusterrole.yaml
```diff
@@ -0,0 +1,22 @@
...
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+rules:
+- apiGroups: [""]
+  resources:
+    - secrets
+  resourceNames:
+    - {{ template "consul.fullname" . }}-server-cert
```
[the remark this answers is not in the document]
ishustava: It should be here! Will update

templates/client-daemonset.yaml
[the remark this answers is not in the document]
lkysow: One use-case would be some other type of tooling. Mostly it's just for completeness.

templates/client-daemonset.yaml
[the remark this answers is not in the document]
lkysow: > For clients, these will be regenerated every time clients are restarted. So hopefully you restart your clients at least once in a year

Nice, nvm then.

templates/tls-init-job.yaml
ishustava: Yeah, I think it does. I'll split them up into two separate secrets.

templates/client-daemonset.yaml
```diff
@@ -235,6 +277,33 @@ spec:
...
+            fieldRef:
+              fieldPath: status.hostIP
+        command:
+          - "/bin/sh"
+          - "-ec"
+          - |
+            cd /consul/tls/client
+            consul tls cert create -client \
+              -dc={{ .Values.global.datacenter }} \
+              -domain={{ .Values.global.domain }} \
```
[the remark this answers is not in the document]
ishustava: You're right they aren't strictly required, but if we don't add them, consul CLI assumes `dc1` and `.consul` domain and adds that as a common name for the client certificate. It doesn't look like it verifies it on RPC connections. This also might be useful if you are connecting to it from somewhere else and want to set that name as the SNI header.

templates/client-daemonset.yaml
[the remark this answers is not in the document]
ishustava: I'm not sure this would be intended and have to think a bit more about security implications of this. I'll keep this out for now, but it should be easy to change if someone requests it.

values.yaml
ishustava: Updated!

test/unit/client-daemonset.bats
ishustava: Yeah I think you're right. Deleted the test.

test/unit/client-daemonset.bats
ishustava: Deleted this test.

test/unit/client-daemonset.bats
ishustava: Yep, deleted this one too.

test/unit/client-daemonset.bats
ishustava: Hmm yeah, I suppose that's true. I've removed these assertions and replaced them with an assertion that the init container exists, as you suggested.

values.yaml
ishustava: Done!

ishustava:
@lkysow I've made the updates from your review and also added support for some consul-k8s components.
Changes made since your last review:
- Added TLS support for Connect inject (requires a consul-k8s docker image containing hashicorp/consul-k8s#181)
- Added TLS support to the Sync catalog deployment
- Added TLS support to the ACL init job ((requires a consul-k8s docker image containing hashicorp/consul-k8s#183)
- Now that consul-k8s supports TLS, switched `httpsOnly` to true by default. 

Outstanding work that will be submitted separately:
- TLS support for snapshot agents
- TLS support for mesh gateways

tpetrychyn:
Testing this branch in one of my clusters and I ran in to an issue that may be related to these changes.

Upgrading from consul-helm version 0.8.1 results in: 
```Error creating: pods "consul-consul-tls-init-" is forbidden: error looking up service account sd/consul-consul-tls-init: serviceaccount "consul-consul-tls-init" not found```

I opted to manually create a service account which allowed `consul-consul-tls-init` to report complete so I deleted it and re-ran helm upgrade but then consul-consul-server-2 would not initialize, giving the error: ```MountVolume.SetUp failed for volume "tls-server-cert" : secret "consul-consul-server-cert" not found```

I finally just blew away my existing consul installation and ran this chart clean.

I will say that running it on a clean environment worked great and the Golang Hashicorp Consul api with TlsConfig plugged in seamlessly so thank you for the work on this! Will be watching closely for using my own CA.

**Comment by ishustava** [comment]
@tpetrychyn thank you so much for trying it out and leaving feedback here. Really appreciate it!

It looks like I forgot to mark the `tls-init-{service-account,clusterrole,clusterrolebinding}` as a `pre-upgrade` hook. I've made the change, and it should work now.

**Comment by tpetrychyn** [comment]
Encountered another issue that I've been banging my head against for days..

Using commit 12ea51bb067d1c267d9f094d668f9c8eebea3d7e I am now getting the error in consul-sync-catalog:

```[GET /health/ready] Error getting leader status: Get https://10.137.132.250:8501/v1/status/leader: dial tcp 10.137.132.250:8501: i/o timeout```

I am using Digital Ocean managed k8 v1.16.2-do.1, my CNI is cilium v1.6.4. 

Using v0.8.1 of this chart works fine in this environment.

The oddest thing is 2 weeks I setup a completely clean DO cluster on 1.16.2-do.1 and this chart worked out of the box, but now even a clean cluster made today has that error..

P.S. commit ^6b659a2 breaks because consul-k8s 0.11.0 is not published on dockerhub

**Comment by ishustava** [comment]
@tpetrychyn This PR requires a new version of consul-k8s for many components, such as Connect Inject and ACL bootstrapping, to work correctly. I optimistically updated the version. This PR, however, should not be merged without that docker image being published first. Sorry if that was confusing!

Have you tried deploying it with TLS disabled? Do you see the same error? At a first glance, it doesn't seem TLS related since the TLS handshake happens after the TCP connection is established.

**Comment by tpetrychyn** [comment]
Thanks for the clarification! Non-tls also has that issue. Setting`exposeGossipPorts: true` appears to fix it.. Does that provide any insight?

File: values.yaml [hunk 27]
```diff
@@ -77,6 +77,40 @@ global:
...
+    # for example, if you're using the UI.
+    serverAdditionalIPSANs: []
+
+    # If verify is true, 'verify_outgoing', 'verify_server_hostname', and
+    # 'verify_incoming_rpc' will be set to true for Consul servers and clients.
+    # Set this to false to incrementally roll out TLS on an existing Consul cluster.
+    # Note: remember to switch it back to true once the rollout is complete.
+    # Please see this guide for more details:
+    # https://learn.hashicorp.com/consul/security-networking/certificates
+    verify: true
```
ishustava: @lkysow this flag is for incremental TLS rollout

lkysow (changes requested):
Looking good! A couple things came up. Here's what I've tested so far.

- [ ]  review new verify: setting
- [x]  components (with httpsonly)
    - [x]  ACL bootstrapping
    - [x]  connect
    - [x]  sync catalog
    - [x]  ent license
    - [x]  snapshot
    - [x]  mesh gateway (just that they came up)
    - [x]  consul UI
    - [x]  consul DNS
- [x]  pod security polices (does not work)
- [ ]  gradual deployment
- [x]  upgrades

templates/client-daemonset.yaml
```diff
@@ -166,17 +207,17 @@ spec:
             - name: aclconfig
               mountPath: /consul/aclconfig
             {{- end }}
-          lifecycle:
```
lkysow: not something that needs to be changed but we might be adding this back as `consul force-leave -prune` due to https://github.com/hashicorp/consul/issues/6897#issuecomment-571153009

values.yaml
ishustava: @adilyse ^^

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
+IMPROVEMENTS:
+
+  * Optionally allow enabling TLS [[GH-313](https://github.com/hashicorp/consul-helm/pull/313/files#)].
+    If `global.tls.enabled` is set to `true`, the Helm chart will generate a CA and necessary certificates and
+    enable TLS for servers, clients, Connect Inject, Mesh gateways, Sync Catalog, ACL bootstrapping, and snapshot agents.
+
+    Note that this feature is only supported if both servers and clients are running
+    on Kubernetes. We will have better support for other deployment architectures,
+    as well as bringing your own CA, in the future.
+
```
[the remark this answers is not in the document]
ishustava: done.

templates/client-daemonset.yaml
```diff
@@ -232,6 +280,38 @@ spec:
           - name: aclconfig
             mountPath: /consul/aclconfig
       {{- end }}
+      {{- if .Values.global.tls.enabled }}
+      - name: client-tls-init
+        image: {{ .Values.global.image }}
```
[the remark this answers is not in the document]
ishustava: good point! thanks for calling this out. I'll update.

templates/server-statefulset.yaml
```diff
@@ -50,14 +50,22 @@ spec:
       tolerations:
         {{ tpl .Values.server.tolerations . | nindent 8 | trim }}
     {{- end }}
-      terminationGracePeriodSeconds: 10
+      terminationGracePeriodSeconds: 30
```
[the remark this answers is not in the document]
ishustava: Sometimes servers were taking longer than 10 seconds to greacefully leave the cluster. In those cases, kubernetes will just force kill the process, which in turn causes servers to exit 1 and not finish the graceful termination.

templates/server-statefulset.yaml
```diff
@@ -163,8 +205,14 @@ spec:
...
-                  curl http://127.0.0.1:8500/v1/status/leader 2>/dev/null | \
-                  grep -E '".+"'
+                  {{- if .Values.global.tls.enabled }}
+                  curl \
+                    --cacert /consul/tls/ca/tls.crt \
+                    https://127.0.0.1:8501/v1/status/leader \
+                  {{- else }}
+                  curl http://127.0.0.1:8500/v1/status/leader \
+                  {{- end }}
+                  2>&1 | grep -E '".+"'
```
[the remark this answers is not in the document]
ishustava: Good catch. I carried this from #137. Not sure why it was changed there, but I'll switch it back.

templates/server-acl-init-job.yaml
```diff
@@ -32,6 +32,12 @@ spec:
     spec:
       restartPolicy: Never
       serviceAccountName: {{ template "consul.fullname" . }}-server-acl-init
+      {{- if .Values.global.tls.enabled }}
```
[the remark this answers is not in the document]
ishustava: Looking at the PSP for this job you've created in #326, I don't think anything will need to change once that's merged. This PR only adds secrets, and the PSP already allows attaching secret volumes. But let me know if I misunderstood you.

templates/tls-init-cleanup-job.yaml
```diff
@@ -0,0 +1,52 @@
+# tls-init-cleanup job deletes Kubernetes secrets created by tls-init
```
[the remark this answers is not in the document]
ishustava: done!

templates/tls-init-job.yaml
```diff
@@ -0,0 +1,79 @@
...
+                -additional-dnsname='*.{{ template "consul.fullname" . }}-server.{{ .Release.Namespace }}.svc.local' \
+                {{- range .Values.global.tls.serverAdditionalIPSANs }}
+                -additional-ipaddress={{ . }} \
+                {{- end }}
+                {{- range .Values.global.tls.serverAdditionalDNSSANs }}
+                 -additional-dnsname={{ . }} \
+                 {{- end }}
+                -dc={{ .Values.global.datacenter }} \
+                -domain={{ .Values.global.domain }}
+              curl -s -X POST --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
```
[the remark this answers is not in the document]
ishustava: done.

ishustava:
@lkysow I've made the changes you requested.

templates/server-acl-init-job.yaml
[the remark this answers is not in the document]
lkysow: Yeah we're good now with my PR. That was just a reminder that we needed to do it for acl init.

lkysow (approved):
🔐 🚀 
Be sure to do a release of consul-k8s first

templates/connect-inject-deployment.yaml
```diff
@@ -60,19 +60,22 @@ spec:
...
+                -consul-ca-cert=/consul/tls/ca/tls.crt \
+                {{- end }}
                 {{- if .Values.connectInject.centralConfig.enabled }}
                 -enable-central-config=true \
                 {{- end }}
                 {{- if (and .Values.connectInject.centralConfig.enabled .Values.connectInject.centralConfig.defaultProtocol) }}
                 -default-protocol="{{ .Values.connectInject.centralConfig.defaultProtocol }}" \
                 {{- end }}
-{{- if .Values.connectInject.certs.secretName }}
+                {{- if .Values.connectInject.certs.secretName }}
```
fouadsemaan, line 72: @ishustava does the generated envoy config file include the TLS context (cert/key)? I see that the CA is added inline.

ishustava: The envoy TLS context for the consul client (aka `local_agent`) doesn't need cert/key since it's not using mTLS client auth to talk to Consul. That's why it only has the CA cert. 

If your question is about authentication, we recommend using ACLs for production environments for that purpose, and when they are enabled in the Helm chart, Envoy will include an ACL token when talking to Consul.

fouadsemaan: Thank you. I was more curious about mTLS for service to service (via envoy) communication.

ishustava: Ah, I see. mTLS for service to service is configured dynamically as you add and remove services by the Consul client agent. [These docs](https://www.consul.io/docs/connect/proxies/envoy#dynamic-configuration) talk a bit about that.

fouadsemaan: Are you referring to the dynamically generated envoy-bootstrap.yaml file via the consul connect envoy -bootrap command? Is that where the cert/key should appear?

ishustava: No, that config is used only for bootstrapping envoy itself. Consul updates envoy config for services dynamically through the API, so you won't find them in a file. You could look at the config though through the envoy admin endpoint. All you need to do is this:

```
kubectl port-forward po/<your pod that has envoy sidecar container> 19000
```

Then go to the browser at http://localhost:19000. There you will see an option to dump the config. Alternatively, you could just curl the endpoint to get the config `curl http://localhost:19000/config_dump`

You could do other things through that endpoint too, like change log levels. Here are the [docs](https://www.envoyproxy.io/docs/envoy/latest/operations/admin) for the admin endpoint in case you'll find them useful.

fouadsemaan: Fantastic! I appreciate your help. Thank you! This is awesome!

PR #317, opened 2019-12-19
Add namespace support to the connect injector

adilyse:
Add the global namespace variables but only pipes them through the
connect injector. Adds connect injector allow/deny namespace lists.
Adds an ACL token for the connect injector that is necessary for
it to create namespaces/binding rules.

### DIFF
--- a/templates/connect-inject-deployment.yaml
+++ b/templates/connect-inject-deployment.yaml
@@ -41,6 +41,35 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            {{- /* An ACL token is only necessary for the connect injector if namespaces are enabled */}}
+            {{- if .Values.global.consulNamespacesEnabled }}
+            - name: HOST_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
+            {{- if (and .Values.connectInject.aclInjectToken.secretName .Values.connectInject.aclInjectToken.secretKey) }}
+            - name: CONSUL_HTTP_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.connectInject.aclInjectToken.secretName }}
+                  key: {{ .Values.connectInject.aclInjectToken.secretKey }}
+            {{- else if .Values.global.bootstrapACLs }}
+            - name: CONSUL_HTTP_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: "{{ template "consul.fullname" . }}-connect-injector-acl-token"
+                  key: "token"
+            {{- end }}
+            {{- if .Values.global.tls.enabled }}
+            - name: CONSUL_HTTP_ADDR
+              value: https://$(HOST_IP):8501
+            - name: CONSUL_CACERT
+              value: /consul/tls/ca/tls.crt
+            {{- else }}
+            - name: CONSUL_HTTP_ADDR
+              value: http://$(HOST_IP):8500
+            {{- end }}
+            {{- end }}
           command:
             - "/bin/sh"
             - "-ec"
@@ -68,6 +97,24 @@ spec:
                 {{- end }}
                 {{- if (and .Values.connectInject.centralConfig.enabled .Values.connectInject.centralConfig.defaultProtocol) }}
                 -default-protocol="{{ .Values.connectInject.centralConfig.defaultProtocol }}" \
+                {{- end }}
+                {{- range $value := .Values.connectInject.k8sAllowNamespaces }}
+                -allow-namespace="{{ $value }}" \
+                {{- end }}
+                {{- range $value := .Values.connectInject.k8sDenyNamespaces }}
+                -deny-namespace="{{ $value }}" \
+                {{- end }}
+                {{- if .Values.global.consulNamespacesEnabled }}
+                -enable-namespaces=true \
+                {{- if .Values.connectInject.consulNamespaces.consulDestinationNamespace }}
+                -consul-namespace={{ .Values.connectInject.consulNamespaces.consulDestinationNamespace }} \
+                {{- end }}
+                {{- if .Values.connectInject.consulNamespaces.mirrorK8S }}
+                -enable-namespace-mirroring=true \
+                {{- if .Values.connectInject.consulNamespaces.mirroringPrefix }}
+                -mirroring-prefix={{ .Values.connectInject.consulNamespaces.mirroringPrefix }} \
+                {{- end }}
+                {{- end }}
                 {{- end }}
                 {{- if .Values.connectInject.certs.secretName }}
                 -tls-cert-file=/etc/connect-injector/certs/{{ .Values.connectInject.certs.certName }} \
@@ -121,6 +168,19 @@ spec:
           secret:
             secretName: {{ template "consul.fullname" . }}-ca-cert
         {{- end }}
+      {{- end }}
+      {{- if and .Values.global.bootstrapACLs .Values.global.consulNamespacesEnabled }}
+      initContainers:
+      - name: injector-acl-init
+        image: {{ .Values.global.imageK8S }}
+        command:
+          - "/bin/sh"
+          - "-ec"
+          - |
+            consul-k8s acl-init \
+              -secret-name="{{ template "consul.fullname" . }}-connect-inject-acl-token" \
+              -k8s-namespace={{ .Release.Namespace }} \
+              -init-type="sync"
       {{- end }}
       {{- if .Values.connectInject.nodeSelector }}
       nodeSelector:
--- a/test/unit/connect-inject-deployment.bats
+++ b/test/unit/connect-inject-deployment.bats
@@ -409,3 +409,319 @@ load _helpers
       yq '.spec.template.spec.containers[0].volumeMounts | length' | tee /dev/stderr)
   [ "${actual}" = "2" ]
 }
+
+#--------------------------------------------------------------------
+# k8sAllowNamespaces & k8sDenyNamespaces
+
+@test "connectInject/Deployment: default is allow `*`, deny nothing" {
+  cd `chart_dir`
+  local object=$(helm template \
+      -x templates/connect-inject-deployment.yaml  \
+      --set 'connectInject.enabled=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command' | tee /dev/stderr)
+
+  local actual=$(echo $object |
+    yq 'map(select(test("allow-namespace"))) | length' | tee /dev/stderr)
+  [ "${actual}" = "1" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("allow-namespace=\"*\""))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+
+  local actual=$(echo $object |
+    yq 'map(select(test("deny-namespace"))) | length' | tee /dev/stderr)
+  [ "${actual}" = "0" ]
+}
+
+@test "connectInject/Deployment: can set allow and deny" {
+  cd `chart_dir`
+  local object=$(helm template \
+      -x templates/connect-inject-deployment.yaml  \
+      --set 'connectInject.enabled=true' \
+      --set 'connectInject.k8sAllowNamespaces[0]=allowNamespace' \
+      --set 'connectInject.k8sDenyNamespaces[0]=denyNamespace' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command' | tee /dev/stderr)
+
+  local actual=$(echo $object |
+    yq 'map(select(test("allow-namespace"))) | length' | tee /dev/stderr)
+  [ "${actual}" = "1" ]
+
+  local actual=$(echo $object |
+    yq 'map(select(test("deny-namespace"))) | length' | tee /dev/stderr)
+  [ "${actual}" = "1" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("allow-namespace=\"allowNamespace\""))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("deny-namespace=\"denyNamespace\""))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
+#--------------------------------------------------------------------
+# namespaces
+
+@test "connectInject/Deployment: namespace options disabled by default" {
+  cd `chart_dir`
+  local object=$(helm template \
+      -x templates/connect-inject-deployment.yaml  \
+      --set 'connectInject.enabled=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command' | tee /dev/stderr)
+
+  local actual=$(echo $object |
+    yq 'any(contains("enable-namespaces"))' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("consul-namespace"))' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("enable-namespace-mirroring"))' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("mirroring-prefix"))' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+}
+
+@test "connectInject/Deployment: namespace options set with .global.consulNamespacesEnabled=true" {
+  cd `chart_dir`
+  local object=$(helm template \
+      -x templates/connect-inject-deployment.yaml  \
+      --set 'connectInject.enabled=true' \
+      --set 'global.consulNamespacesEnabled=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command' | tee /dev/stderr)
+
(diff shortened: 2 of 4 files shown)


values.yaml
```diff
@@ -77,6 +77,29 @@ global:
   # Additionally requires Consul >= 1.4 and consul-k8s >= 0.10.1.
   bootstrapACLs: false
 
+  # consulNamespaces are an Enterprise only feature available in Consul v1.7+.
+  # A valid Consul Enterprise license is required to make use of this feature.
+  # Requires consul-k8s v0.x+
+  consulNamespaces:
```
lkysow: I thought this was going under `syncCatalog`?

values.yaml
```diff
@@ -77,6 +77,29 @@ global:
...
+    # consulNamespaceName is the name of the Consul namespace to register all
+    # k8s services into. If the Consul namespace does not already exist,
+    # it will be created. This will be ignored if `mirrorK8s` is true.
+    consulNamespaceName: "default"
+
+    # mirrorK8s causes k8s services to be registered into a Consul namespace
+    # equivalent to their k8s namespace. If the Consul namespace does not
+    # already exist, it will be created. Turning this on overrides the
+    # `consulNamespaceName` setting.
+    mirrorK8s: false
```
lkysow: this always bites me but we use `K8S` everwhere else so should probably match that here

values.yaml
adilyse: What about the Connect injector then? It needs this same info. Are you suggesting we duplicate it both places?

lkysow: I thought that's what we had discussed on the RFC but I see now that when you updated it you only split up the allow/deny lists.

I'm trying to think through the use-cases:
1. Only using catalog sync - doesn't matter in global or syncCatalog
1. Only using connect inject - doesn't matter in global or connectInject
1. Using both catalog sync and connect inject - I can think of some use-cases where it might be nice to manage them separately: I'm using connect inject with namespace mirroring but have a couple of services that I can't put sidecars on and I want them still registered in Consul but all in the same Consul namespace called "k8s-sync".  Why? I don't know 😄.

I think on the tradeoff, we enable some use-cases if we split them up at the cost of having more config. I don't think it significantly complicates the chart though because in the end, the configs are localized to their own components: sync and inject. Where it might complicate things is in the acl bootstrap. I'm just looking at that PR now so don't have the context there yet. Would it make things a lot more difficult?

values.yaml
```diff
@@ -465,6 +488,24 @@ connectInject:
...
+  # injection in. If a k8s namespace is not included or is listed in `denyNamespacesList`,
+  # pods in that k8s namespace will not be injected even if they are explicitly
+  # annotated. Use "*" to automatically allow all k8s namespaces.
+  # For example, ["namespace1", "namespace2"] will only allow pods in the k8s
+  # namespaces `namespace1` and `namespace2` to have Connect sidecars injected
+  # and registered with Consul. All other k8s namespaces will be ignored.
+  # Note: `denyNamespacesList` takes precedence over values defined here and
+  # `namespaceSelector` takes precedence over both since it is applied first.
+  # Requires consul-k8s v0.x+
+  allowNamespacesList: ["*"]
```
```suggestion
  k8sAllowNamespaces: ["*"]
```
k8s to make it clear we're talking k8s namespaces (in inject that doesn't matter but it will match sync)
removing List because that's inferred from the datatype

values.yaml
```diff
@@ -465,6 +488,24 @@ connectInject:
...
+
+  # denyNamespacesList is a list of k8s namespaces that should not allow Connect
+  # sidecar injection. This list takes precedence over `allowNamespacesList`.
+  # Note: `namespaceSelector` takes precedence over this since it is applied first
+  # Requires consul-k8s v0.x+
+  denyNamespacesList: []
```
```suggestion
  k8sDenyNamespaces: []
```

values.yaml
```diff
@@ -77,6 +77,29 @@ global:
   # Additionally requires Consul >= 1.4 and consul-k8s >= 0.10.1.
   bootstrapACLs: false
 
+  # consulNamespaces are an Enterprise only feature available in Consul v1.7+.
+  # A valid Consul Enterprise license is required to make use of this feature.
+  # Requires consul-k8s v0.x+
```
lkysow: TODO

values.yaml
```diff
@@ -77,6 +77,29 @@ global:
...
+  consulNamespaces:
+    enabled: false
```
lkysow: Needs docs

values.yaml
```diff
@@ -77,6 +77,29 @@ global:
...
+
+    # consulNamespaceName is the name of the Consul namespace to register all
+    # k8s services into. If the Consul namespace does not already exist,
+    # it will be created. This will be ignored if `mirrorK8s` is true.
+    consulNamespaceName: "default"
```
lkysow: do we need Name?

values.yaml
```diff
@@ -77,6 +77,29 @@ global:
...
   bootstrapACLs: false
 
+  # consulNamespaces are an Enterprise only feature available in Consul v1.7+.
+  # A valid Consul Enterprise license is required to make use of this feature.
+  # Requires consul-k8s v0.x+
+  consulNamespaces:
+    enabled: false
+
+    # consulNamespaceName is the name of the Consul namespace to register all
+    # k8s services into. If the Consul namespace does not already exist,
```
lkysow: pending decision on splitting it up but if we don't split it then this should say "k8s services into for both Connect and Sync Catalog"

values.yaml
```diff
@@ -77,6 +77,29 @@ global:
...
+    # it will be created. This will be ignored if `mirrorK8s` is true.
+    consulNamespaceName: "default"
+
+    # mirrorK8s causes k8s services to be registered into a Consul namespace
+    # equivalent to their k8s namespace. If the Consul namespace does not
```
lkysow: equivalent might be confusing here. Can we say "into a Consul namespace with the same name plus the prefix specified by `mirroringPrefix`?

values.yaml
```diff
@@ -77,6 +77,29 @@ global:
...
+    # already exist, it will be created. Turning this on overrides the
+    # `consulNamespaceName` setting.
+    mirrorK8s: false
+
+    # If `mirrorK8s` is set to true, `mirroringPrefix` allows each Consul namespace
+    # to be given a prefix. For example, if `mirroringPrefix` is set to "k8s-", a
+    # service in the k8s `staging` namespace will be registered into the
+    # `k8s-staging` Consul namespace.
```
lkysow: 👍

values.yaml
```diff
@@ -465,6 +488,24 @@ connectInject:
   #     namespace-label: label-value
   namespaceSelector: null
 
+  # allowNamespacesList is a list of k8s namespaces to allow Connect sidecar
+  # injection in. If a k8s namespace is not included or is listed in `denyNamespacesList`,
+  # pods in that k8s namespace will not be injected even if they are explicitly
+  # annotated. Use "*" to automatically allow all k8s namespaces.
```
lkysow: > allow all k8s namespaces

What about `kube-system` and `kube-public`?

```suggestion
  # annotated. Use ["*"] to automatically allow all k8s namespaces.
```

Might be a good idea to show them it needs to be in an array so they don't set it to the string

values.yaml
```diff
@@ -465,6 +488,24 @@ connectInject:
...
+  # For example, ["namespace1", "namespace2"] will only allow pods in the k8s
+  # namespaces `namespace1` and `namespace2` to have Connect sidecars injected
+  # and registered with Consul. All other k8s namespaces will be ignored.
+  # Note: `denyNamespacesList` takes precedence over values defined here and
+  # `namespaceSelector` takes precedence over both since it is applied first.
```
lkysow: 👍

values.yaml
```diff
@@ -465,6 +488,24 @@ connectInject:
...
+  # Requires consul-k8s v0.x+
```
lkysow: TODO

values.yaml
```diff
@@ -465,6 +488,24 @@ connectInject:
...
+  allowNamespacesList: ["*"]
+
+  # denyNamespacesList is a list of k8s namespaces that should not allow Connect
+  # sidecar injection. This list takes precedence over `allowNamespacesList`.
+  # Note: `namespaceSelector` takes precedence over this since it is applied first
+  # Requires consul-k8s v0.x+
```
lkysow: TODO

values.yaml
```diff
@@ -513,6 +554,14 @@ connectInject:
   # method for Connect inject, set this to the name of your auth method.
   overrideAuthMethodName: ""
 
+  # aclInjectToken refers to a Kubernetes secret that you have created that contains
```
lkysow: is this required for namespaces? Seems like a good idea, just wondering why it's in this PR.

values.yaml
```diff
@@ -513,6 +554,14 @@ connectInject:
...
+  # an ACL token for your Consul cluster which allows the Connect injector the correct
+  # permissions. This is only needed if Consul namespaces and ACLs are enabled on the
```
lkysow: Can we list the exact permissions? Or link to where it's documented.

values.yaml
adilyse: I could see putting the Consul namespace under both the sync and injector. That should allow independent management of namespaces for those two processes. 

I don't think it makes sense to have the enable namespace flag in both though. Either you can run namespaces or not. It gets hard to reason about that value if it's turned on one place and off in another.

lkysow: > I don't think it makes sense to have the enable namespace flag in both though. Either you can run namespaces or not. It gets hard to reason about that value if it's turned on one place and off in another.

Yeah that makes sense as a global setting. That flag is really asking the user if they're running Consul Enterprise right? Maybe that's a better name? Or `namespacesEnabled` or something. Because it seems like it's something that the user can turn on but really they can't, it's more telling us whether they have that capability or not.

adilyse: Left the namespace enabling as a global flag, moved the destination namespace and mirroring options under both sync catalog and the connect injector.

values.yaml

values.yaml
adilyse: These will be in `k8sDenyNamespaces` by default.

values.yaml

values.yaml
adilyse: This is required for namespaces. Previously, the connect injector didn't need an ACL token as all of the operations that required it for that process were done in the init container during injection with an acl provided by the auth method. With namespaces, however, the connect injector may need to create a Consul namespace which requires privileges if acls are enabled. This allows folks to provide that token without having run the bootstrapping, as is also available as an option for sync catalog.

values.yaml
adilyse: Added info about the required permissions.

values.yaml

templates/connect-inject-deployment.yaml
```diff
@@ -41,6 +41,23 @@ spec:
...
+            {{- /* An ACL token is only necessary for the connect injector if namespaces are enabled */}}
+            {{- if .Values.global.consulNamespacesEnabled }}
+            {{- if (and .Values.connectInject.aclInjectToken.secretName .Values.connectInject.aclInjectToken.secretKey) }}
+            - name: CONSUL_HTTP_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.connectInject.aclInjectToken.secretName }}
+                  key: {{ .Values.connectInject.aclInjectToken.secretKey }}
+            {{- end }}
+            {{- if .Values.global.bootstrapACLs }}
```
lkysow: should this be an else if? What happens if the env var is set twice?

values.yaml
```diff
@@ -499,6 +508,53 @@ connectInject:
...
+  k8sDenyNamespaces: ["kube-system", "kube-public"]
+
+  # [Enterprise Only] These settings manage the connect injector's interaction with
+  # Consul namespaces (requires consul-ent v1.7+ and consul-k8s v0.12+).
+  # Also, `global.consulNamespacesEnabled` must be true.
+  consulNamespaces:
+    # consulDestinationNamespace is the name of the Consul namespace to register all
+    # k8s services into. If the Consul namespace does not already exist,
+    # it will be created. This will be ignored if `mirrorK8S` is true.
+    consulDestinationNamespace: "default"
```
lkysow: We should match the naming of the sync catalog. I suggested some changes there so pending what we do there we may want to update these.

values.yaml
```diff
@@ -499,6 +508,53 @@ connectInject:
...
+  # k8sDenyNamespaces is a list of k8s namespaces that should not allow Connect
+  # sidecar injection. This list takes precedence over `k8sAllowNamespaces`.
+  # `*` is not supported because then nothing would be allowed to be injected.
+  #
+  # For example, if `k8sAllowNamespaces is `["*"]` and k8sDenyNamespaces is
+  # `["namespace1", "namespace2"]`, then all k8s namespaces besides "namespace1"
+  # and "namespace2" will be injected.
+  # Note: `namespaceSelector` takes precedence over this since it is applied first.
+  # Requires consul-k8s v0.12+.
+  k8sDenyNamespaces: ["kube-system", "kube-public"]
```
lkysow: I think unlike in the catalog sync that this should never be injected into and that this should also be hard coded in the injector. This would totally break kube and I wonder how you'd fix it.

values.yaml
adilyse: I switched all of them to `consulDestinationNamespace` per your recommendation.

values.yaml
adilyse: These are never injected:

values.yaml
lkysow: Yep but I realized that those might need to change: https://github.com/hashicorp/consul-helm/pull/337#discussion_r369666623.

values.yaml
lkysow: I think these should be empty but in the comment you can say that kube-system and kube-public are never injected.

test/unit/connect-inject-deployment.bats
```diff
@@ -409,3 +409,142 @@ load _helpers
...
+  local actual=$(echo $object |
+    yq 'any(contains("consul-namespace=default"))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+  local actual=$(echo $object |
+    yq 'any(contains("enable-namespace-mirroring=true"))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+  local actual=$(echo $object |
+    yq 'any(contains("mirroring-prefix=k8s-"))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
```
ishustava: Needs tests for logic around `CONSUL_HTTP_TOKEN` and `injector-acl-init` init container.

ishustava (changes requested):
Hey @adilyse, I left a couple of comments and suggestions. I wasn't able to test it though because the webhook deployment needs to be provided with Consul's HTTP configuration.

templates/connect-inject-deployment.yaml
adilyse: Updated.

values.yaml
adilyse: Removed. Added comment that they will never be injected.

values.yaml
```diff
@@ -499,6 +508,53 @@ connectInject:
...
+    # k8s services into. If the Consul namespace does not already exist,
+    # it will be created. This will be ignored if `mirrorK8S` is true.
+    consulDestinationNamespace: "default"
+
+    # mirrorK8S causes k8s services to be registered into a Consul namespace
+    # of the same name as their k8s namespace, optionally prefixed if
+    # `mirroringPrefix` is set below. If the Consul namespace does not
+    # already exist, it will be created. Turning this on overrides the
+    # `consulDestinationNamespace` setting.
+    # `addK8SNamespaceSuffix` may no longer be needed if enabling this option.
```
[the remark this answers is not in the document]
adilyse: Removed.

templates/connect-inject-deployment.yaml
```diff
@@ -41,6 +41,23 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            {{- /* An ACL token is only necessary for the connect injector if namespaces are enabled */}}
+            {{- if .Values.global.consulNamespacesEnabled }}
```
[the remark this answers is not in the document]
adilyse: Added.

test/unit/connect-inject-deployment.bats
```diff
@@ -409,3 +409,142 @@ load _helpers
...
+  local actual=$(echo $object |
+    yq 'any(contains("deny-namespace=\"kube-system\""))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("deny-namespace=\"kube-public\""))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
+@test "connectInject/Deployment: can set allow and deny {
```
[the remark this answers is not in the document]
adilyse: Fixed.

values.yaml
```diff
@@ -547,6 +603,15 @@ connectInject:
   # method for Connect inject, set this to the name of your auth method.
   overrideAuthMethodName: ""
 
+  # aclInjectToken refers to a Kubernetes secret that you have created that contains
+  # an ACL token for your Consul cluster which allows the Connect injector the correct
+  # permissions. This is only needed if Consul namespaces and ACLs are enabled on the
+  # Consul cluster and you are not using `global.bootstrapACLs`.
```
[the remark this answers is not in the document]
adilyse: Changed.

test/unit/connect-inject-deployment.bats
adilyse: Added.

PR #320, opened 2019-12-23
Update test-runner.yaml

tehmoon:
Add namespace to the tester and force restart it if failing.

### DIFF
--- a/templates/tests/test-runner.yaml
+++ b/templates/tests/test-runner.yaml
@@ -3,6 +3,7 @@ apiVersion: v1
 kind: Pod
 metadata:
   name: "{{ template "consul.fullname" . }}-test"
+  namespace: {{ .Release.Namespace }}
   labels:
     app: {{ template "consul.name" . }}
     chart: {{ template "consul.chart" . }}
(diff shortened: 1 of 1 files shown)


tehmoon:
Hi @ishustava.
Thank you for following up!

I also changed the `restartPolicy` because using this chart with `helm template` and applying the result to kubernetes directly, I saw that the tester pod was actually being scheduled right away, before the consul pods, making in fail right away.

If that blocks your pipeline, let me know and I'll remove it.

lkysow:
If using helm template, we recommend running with

```
tests:
  enabled: false
```

We added that just for helm template.

tehmoon:
Hi @lkysow,

Yes that's what I'm doing but if one wants to starts the tests, it does not work if using a namespace.
Alternatively, it works if kustomize sets the namespace on top of `helm3 template`.

ishustava:
Hey @tehmoon, if you don't mind removing the `restartPolicy` change (based on what @lkysow said), I can approve your PR. The namespace change looks good.

tehmoon:
@ishustava Done!! Thank you so much for the follow up.

ishustava:
Thank you for making the change @tehmoon and for the PR!

PR #325, opened 2020-01-06
Enterprise License Job PodSecurityPolicy

lkysow:
- add podsecuritypolicy for enterprise license job following the pattern
for other psps
- set same requirements across resources for rendering: servers enabled
and enterprise license secret fully specified. The different resources
had different requirements before, for example that clients are running
or that bootstrapACLs are true
- always create a separate service account for the job, even if acls are
disabled. This is to match the pattern across the rest of the templates
where we always create a separate service account
- remove comment across our PSP's that `RunAsAny` prevented running as root. This is not correct.

### DIFF
--- a/templates/client-podsecuritypolicy.yaml
+++ b/templates/client-podsecuritypolicy.yaml
@@ -41,7 +41,6 @@ spec:
   hostIPC: false
   hostPID: false
   runAsUser:
-    # Require the container to run without root privileges.
     rule: 'RunAsAny'
   seLinux:
     rule: 'RunAsAny'
--- a/templates/enterprise-license-clusterrole.yaml
+++ b/templates/enterprise-license-clusterrole.yaml
@@ -1,6 +1,4 @@
 {{- if (or (and (ne (.Values.server.enabled | toString) "-") .Values.server.enabled) (and (eq (.Values.server.enabled | toString) "-") .Values.global.enabled)) }}
-{{- if (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}
-{{- if .Values.global.bootstrapACLs }}
 {{- if (and .Values.server.enterpriseLicense.secretName .Values.server.enterpriseLicense.secretKey) }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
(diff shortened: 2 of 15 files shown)


lkysow:
@adilyse would appreciate a look at this from you. In https://github.com/hashicorp/consul-helm/pull/172 you made using a separate service account for the ent-license job dependent on whether ACLs were enabled. This was because the clusterrole was only needed if acls were enabled.

Now the clusterrole is needed if ACLs are enabled *or* if psp's are enabled. I could add that conditional to the service account but I looked at the rest of our templates and we're always creating a service account, even if we don't need to target it via a clusterrole rule. I think this pattern makes sense (to always have a service account per component) but would like your review.

templates/enterprise-license-clusterrole.yaml
```diff
@@ -20,6 +20,14 @@ rules:
...
       - get
 {{- end }}
+{{- if .Values.global.enablePodSecurityPolicies }}
+  - apiGroups: ["policy"]
+    resources: ["podsecuritypolicies"]
+    resourceNames:
+      - {{ template "consul.fullname" . }}-enterprise-license
+    verbs:
+      - use
+{{- end }}
```
ishustava: If `bootstrapACLs` and `enablePodSecurityPolicies` are both `false`, we'll have a cluster role without the `rules` key. I don't think it's wrong, and Kube won't complain about it. Just something to be aware of because in other places, for example in [`client-clusterrole.yaml`](https://github.com/hashicorp/consul-helm/blob/9cc2e9d58a632eecb9ec419a151fcf0605fed57b/templates/client-clusterrole.yaml#L31), we're setting it an empty array rather than leaving it out.

templates/client-podsecuritypolicy.yaml
```diff
@@ -41,7 +41,6 @@ spec:
   hostIPC: false
   hostPID: false
   runAsUser:
-    # Require the container to run without root privileges.
```
ishustava: I'm curious, are you deleting these comments because of duplication?

ishustava (approved):
Looks good to me! I left a couple of comments, but nothing blocking.

templates/client-podsecuritypolicy.yaml
lkysow: I'm deleting them because they're incorrect. `RunAsAny` means the container can run as anything. It doesn't prevent root. `MustRunAs` or `MayRunAs` can be used to prevent root.

templates/enterprise-license-clusterrole.yaml
lkysow: fixed

PR #329, opened 2020-01-13
Update tls-init-job.yaml

tehmoon:
Missing domain for the tls-init-job

### DIFF
--- a/templates/tls-init-job.yaml
+++ b/templates/tls-init-job.yaml
@@ -45,7 +45,8 @@ spec:
             - "/bin/sh"
             - "-ec"
             - |
-              consul tls ca create
+              consul tls ca create \
+                -domain={{ .Values.global.domain }}
               consul tls cert create -server \
                 -days=730 \
                 -additional-dnsname='{{ template "consul.fullname" . }}-server' \


templates/tls-init-job.yaml
```diff
@@ -45,7 +45,8 @@ spec:
             - "/bin/sh"
             - "-ec"
             - |
-              consul tls ca create
+              consul tls ca create \
+                -domain={{ .Values.global.domain }}
```
ishustava: I don't think the CA needs a domain. Are you seeing errors when you don't provide it?

tehmoon: The CA per se doesn't need one, the error lies in the fact that if a domain is set, the following command `consul tls cert create -server` which has `domain` set, will not be able to find the right file generated by `consul tls ca create`.
IE:
```
creator@infrastructure:/tmp/tmp.NtWBXsJ0Gc$ consul tls ca create
creator@infrastructure:/tmp/tmp.NtWBXsJ0Gc$ ls
consul-agent-ca-key.pem  consul-agent-ca.pem
creator@infrastructure:/tmp/tmp.NtWBXsJ0Gc$ consul tls cert create -server -domain test.tld
Error reading CA: open test.tld-agent-ca.pem: no such file or directory
```

If the `domain` flag is set for the CA:
```
creator@infrastructure:/tmp/tmp.NtWBXsJ0Gc$ consul tls ca create -domain test.tld
==> Saved test.tld-agent-ca.pem
==> Saved test.tld-agent-ca-key.pem
```

And thinking more about it, it could actually be solved with `-ca`:
```
  -ca=<string>
     Provide path to the ca. Defaults to #DOMAIN#-agent-ca.pem.
```

Like this:
```
creator@infrastructure:/tmp/tmp.NtWBXsJ0Gc$ consul tls cert create -server -ca consul-agent-ca.pem
==> WARNING: Server Certificates grants authority to become a
    server and access all state in the cluster including root keys
    and all ACL tokens. Do not distribute them to production hosts
    that are not server nodes. Store them as securely as CA keys.
==> Using consul-agent-ca.pem and consul-agent-ca-key.pem
==> Saved dc1-server-consul-0.pem
==> Saved dc1-server-consul-0-key.pem
```

ishustava: Great catch, thank you! I think it makes sense to provide the domain for consistency, even though this file will only exist during the duration of the job run.

tehmoon:
Thank you @ishustava great work for that #313 PR!!

PR #331, opened 2020-01-13
Update CHANGELOG with a note about no-downtime TLS rollout

ishustava:


ishustava:
Going to skip values.yaml update since it requires cutting a new patch release of the Helm chart, and I don't think it's worth it.

PR #332, opened 2020-01-14
Add server.exposeWANPorts option for federation

lkysow:
Will expose the server's WAN ports (serf and rpc) as host ports and will
advertise the host IP as the wan address. This is needed for federation
if users can make their host ips routable.

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -89,6 +89,12 @@ spec:
                   name: {{ .Values.global.gossipEncryption.secretName }}
                   key: {{ .Values.global.gossipEncryption.secretKey }}
             {{- end }}
+            {{ if .Values.server.exposeWANPorts }}
+            - name: NODE_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
+            {{- end }}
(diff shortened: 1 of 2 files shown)


lkysow:
Would close https://github.com/hashicorp/consul-helm/pull/84

lkysow:
Yeah we need to figure out if it makes sense to have this or wait for mesh gateway fed. There's still TLS and acl replication to figure out.

templates/server-statefulset.yaml
```diff
@@ -98,6 +104,9 @@ spec:
 
               exec /bin/consul agent \
                 -advertise="${POD_IP}" \
+                {{- if .Values.server.exposeWANPorts }}
+                -advertise-wan="${NODE_IP}" \
```
lkysow: NOTE: needs to also advertise the node_ip on the lan for folks joining external clients (https://github.com/hashicorp/consul-helm/issues/358#issuecomment-594960342)

lkysow:
This is no longer necessary for federation because instead mesh gateway should be used: https://www.consul.io/docs/k8s/installation/multi-cluster.

For joining an external consul client to a kubernetes consul datacenter https://github.com/hashicorp/consul-helm/pull/740 will be used.

PR #335, opened 2020-01-15
Add Affinity and Tolerations to Inject and Sync

lawliet89:


lawliet89:
@ishustava @adilyse @lkysow can I get this PR looked at please? It's almost half a year old and it's the only two remaining services that are missing affinity/tolerations.

PR #336, opened 2020-01-15
Use namespace environment variable instead of helm

tehmoon:


tehmoon:
This is the result of the certificate alternate names:
```
 DNS:preveil-consul-server, DNS:*.preveil-consul-server, DNS:'*.preveil-consul-server.consul-vault', DNS:'*.preveil-consul-server.consul-vault.svc', DNS:server.vault-local.prod.preveil.com, DNS:localhost, IP Address:127.0.0.1
```
Works as expected.

lkysow:
Thanks for the PR! Can you explain what difference this will make?

tehmoon:
Thank you for taking the time @lkysow. It adds better consistency, see:

- https://github.com/hashicorp/consul-helm/blob/fafa867e074e296a06de7ed39c8e6db6bdfb7a18/templates/tls-init-job.yaml#L54
- https://github.com/hashicorp/consul-helm/blob/fafa867e074e296a06de7ed39c8e6db6bdfb7a18/templates/tls-init-job.yaml#L55

It's also based on the `metadata.namespace` and not the helm template which works better when using `kustomize` instead. It does not change much if the environment is `helm` though.

ishustava:
@tehmoon Once the conflicts are resolved, we can merge it!

tehmoon:
@ishustava resolved! Thank you so much for the time to look into this.

ishustava:
Thanks for the PR @tehmoon!

PR #337, opened 2020-01-16
Add sync catalog Consul namespace support to the Helm chart

adilyse:
This adds settings to configure catalog sync's interaction with
Consul namespaces. It adds a global flag to indicate that Consul
namespaces should be used and sync options to either sync into
a single namespace or mirror k8s namespaces with an optional prefix.

Additionally, it adds allow and deny lists as a way to manage which
k8s namespaces to sync. These do not require Enterprise or Consul
namespaces.

To do:
- [x] Add bats tests for the new options

### DIFF
--- a/templates/sync-catalog-deployment.yaml
+++ b/templates/sync-catalog-deployment.yaml
@@ -95,6 +95,12 @@ spec:
                 {{- if .Values.syncCatalog.k8sSourceNamespace }}
                 -k8s-source-namespace="{{ .Values.syncCatalog.k8sSourceNamespace}}" \
                 {{- end }}
+                {{- range $value := .Values.syncCatalog.k8sAllowNamespaces }}
+                -allow-namespace="{{ $value }}" \
+                {{- end }}
+                {{- range $value := .Values.syncCatalog.k8sDenyNamespaces }}
+                -deny-namespace="{{ $value }}" \
+                {{- end }}
                 -k8s-write-namespace=${NAMESPACE} \
                 {{- if (not .Values.syncCatalog.syncClusterIPServices) }}
                 -sync-clusterip-services=false \
@@ -117,6 +123,18 @@ spec:
                 {{- if .Values.syncCatalog.addK8SNamespaceSuffix}}
                 -add-k8s-namespace-suffix \
                 {{- end}}
+                {{- if .Values.global.consulNamespacesEnabled }}
+                -enable-namespaces=true \
+                {{- if .Values.syncCatalog.consulNamespaces.consulDestinationNamespace }}
+                -consul-namespace={{ .Values.syncCatalog.consulNamespaces.consulDestinationNamespace }} \
+                {{- end }}
+                {{- if .Values.syncCatalog.consulNamespaces.mirrorK8S }}
+                -enable-namespace-mirroring=true \
+                {{- if .Values.syncCatalog.consulNamespaces.mirroringPrefix }}
+                -mirroring-prefix={{ .Values.syncCatalog.consulNamespaces.mirroringPrefix }} \
+                {{- end }}
+                {{- end }}
+                {{- end }}
           livenessProbe:
             httpGet:
               path: /health/ready
--- a/test/unit/sync-catalog-deployment.bats
+++ b/test/unit/sync-catalog-deployment.bats
@@ -415,3 +415,167 @@ load _helpers
   actual=$(echo $env | jq -r '. | select(.name == "CONSUL_CACERT") | .value' | tee /dev/stderr)
     [ "${actual}" = "/consul/tls/ca/tls.crt" ]
 }
+
+#--------------------------------------------------------------------
+# k8sAllowNamespaces & k8sDenyNamespaces
+
+@test "syncCatalog/Deployment: default is allow `*`, deny kube-system and kube-public" {
+  cd `chart_dir`
+  local object=$(helm template \
+      -x templates/sync-catalog-deployment.yaml  \
+      --set 'syncCatalog.enabled=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command' | tee /dev/stderr)
+
+  local actual=$(echo $object |
+    yq 'map(select(test("allow-namespace"))) | length' | tee /dev/stderr)
+  [ "${actual}" = "1" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("allow-namespace=\"*\""))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("deny-namespace=\"kube-system\""))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("deny-namespace=\"kube-public\""))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
+@test "syncCatalog/Deployment: can set allow and deny {
+  cd `chart_dir`
+  local object=$(helm template \
+      -x templates/sync-catalog-deployment.yaml  \
+      --set 'syncCatalog.enabled=true' \
+      --set 'syncCatalog.k8sAllowNamespaces[0]=allowNamespace' \
+      --set 'syncCatalog.k8sDenyNamespaces[0]=denyNamespace' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command' | tee /dev/stderr)
+
+  local actual=$(echo $object |
+    yq 'map(select(test("allow-namespace"))) | length' | tee /dev/stderr)
+  [ "${actual}" = "1" ]
+
+  local actual=$(echo $object |
+    yq 'map(select(test("deny-namespace"))) | length' | tee /dev/stderr)
+  [ "${actual}" = "1" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("allow-namespace=\"allowNamespace\""))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("deny-namespace=\"denyNamespace\""))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
+#--------------------------------------------------------------------
+# namespaces
+
+@test "syncCatalog/Deployment: namespace options disabled by default" {
+  cd `chart_dir`
+  local object=$(helm template \
+      -x templates/sync-catalog-deployment.yaml  \
+      --set 'syncCatalog.enabled=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command' | tee /dev/stderr)
+
+  local actual=$(echo $object |
+    yq 'any(contains("enable-namespaces"))' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("consul-namespace"))' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("enable-namespace-mirroring"))' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+
+  local actual=$(echo $object |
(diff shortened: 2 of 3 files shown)


values.yaml
```diff
@@ -40,6 +40,7 @@ global:
   # remove these checks to make the sync work.
   # If using bootstrapACLs then must be >= 0.10.1.
   # If using connect inject then must be >= 0.10.1.
+  # If using Consul namespaces, must be >= 0.12.
```
```suggestion
  # If using Consul Enterprise namespaces, must be >= 0.12.
```

In case people get confused?

values.yaml
```diff
@@ -111,6 +112,14 @@ global:
     # clients and servers and only accept HTTPS connections.
     httpsOnly: true
 
+  # [Enterprise Only] consulNamespacesEnabled indicates that you are running
+  # Consul v1.7+ with a valid Consul Enterprise license and would like to make
```
```suggestion
  # Consul Enterprise v1.7+ with a valid Consul Enterprise license and would like to make
```

You have to run the consul enterprise binary so make that clear

values.yaml
```diff
@@ -412,11 +421,52 @@ syncCatalog:
...
+  # namespaces `namespace1` and `namespace2` to be synced and registered
+  # with Consul. All other k8s namespaces will be ignored.
+  # Note: `k8sDenyNamespaces` takes precedence over values defined here.
+  # Requires consul-k8s v0.12+
+  k8sAllowNamespaces: ["*"]
+
+  # k8sDenyNamespaces is a list of k8s namespaces that should not have their
+  # service synced. This list takes precedence over `k8sAllowNamespaces`.
+  # `*` is not supported because then nothing would be allowed to sync.
+  # Requires consul-k8s v0.12+.
```
lkysow: I think an example like you gave above would be great here. Like "If k8sAllowNamespaces is ["*"] and k8sDenyNamespaces is ["namespace1", "namespace2"] then all namespaces except namespace1, namespace2 will be synced.

values.yaml
```diff
@@ -412,11 +421,52 @@ syncCatalog:
...
+  k8sDenyNamespaces: ["kube-system", "kube-public"]
+
+  # [DEPRECATED] Use the k8sAllowNamespaces and k8sDenyNamespaces instead.
```
```suggestion
  # [DEPRECATED] Use k8sAllowNamespaces and k8sDenyNamespaces instead.
```

values.yaml
```diff
@@ -412,11 +421,52 @@ syncCatalog:
...
+  # sourceNamespace is the Kubernetes namespace to watch for service
   # changes and sync to Consul. If this is not set then it will default
   # to all namespaces.
   k8sSourceNamespace: null
 
+  # [Enterprise Only] These settings manage the catalog sync's interaction with
+  # Consul namespaces (requires consul-ent v1.7+ and consul-k8s v0.12+).
+  # Also, the `global.consulNamespacesEnabled` setting must be set to true.
```
```suggestion
  # Also, `global.consulNamespacesEnabled` must be true.
```

values.yaml
```diff
@@ -412,11 +421,52 @@ syncCatalog:
...
+  consulNamespaces:
+    # consulNamespaceName is the name of the Consul namespace to register all
+    # k8s services into. If the Consul namespace does not already exist,
+    # it will be created. This will be ignored if `mirrorK8S` is true.
+    consulNamespaceName: "default"
```
lkysow: What about `consulDestinationNamespace`?

lkysow (commented):
This looks great! Really good documentation on the values. I haven't tested anything out yet so can't approve atm.

values.yaml
```diff
@@ -412,11 +421,52 @@ syncCatalog:
   # prepended with "consul-". (Consul -> Kubernetes sync)
   k8sPrefix: null
 
-  # k8sSourceNamespace is the Kubernetes namespace to watch for service
+  # k8sAllowNamespaces is a list of k8s namespaces to sync the k8s services from.
+  # If a k8s namespace is not included or is listed in `k8sDenyNamespaces`,
+  # pods in that k8s namespace will not be synced even if they are explicitly
```
```suggestion
  # services in that k8s namespace will not be synced even if they are explicitly
```

values.yaml
```diff
@@ -412,11 +421,52 @@ syncCatalog:
...
+  # annotated. Use ["*"] to automatically allow all k8s namespaces.
+  #
+  # For example, ["namespace1", "namespace2"] will only allow pods in the k8s
```
```suggestion
  # For example, ["namespace1", "namespace2"] will only allow services in the k8s
```

values.yaml
ishustava: +1 to an example. Also, it'd be good to say what should this value be set to if you don't want to deny any namespaces. Don't know why would people want to sync services in kube-system ns, but just for completeness.

templates/sync-catalog-deployment.yaml
```diff
@@ -117,6 +123,16 @@ spec:
                 {{- if .Values.syncCatalog.addK8SNamespaceSuffix}}
                 -add-k8s-namespace-suffix \
                 {{- end}}
+                {{- if .Values.global.consulNamespacesEnabled }}
+                -enable-namespaces=true \
```
ishustava: Could this also be just `-enable-namespaces` since it's a boolean flag?

values.yaml
```diff
@@ -412,11 +421,52 @@ syncCatalog:
...
+    # consulNamespaceName is the name of the Consul namespace to register all
+    # k8s services into. If the Consul namespace does not already exist,
+    # it will be created. This will be ignored if `mirrorK8S` is true.
+    consulNamespaceName: "default"
+
+    # mirrorK8S causes k8s services to be registered into a Consul namespace
+    # of the same name as their k8s namespace, optionally prefixed if
+    # `mirroringPrefix` is set below. If the Consul namespace does not
+    # already exist, it will be created. Turning this on overrides the
+    # `consulNamespaceName` setting.
```
lkysow: We should also mention that they may want to set `addK8SNamespaceSuffix: false` since they don't need the prefixes anymore

templates/sync-catalog-deployment.yaml
```diff
@@ -117,6 +123,16 @@ spec:
...
+                {{- if .Values.syncCatalog.consulNamespaces.consulNamespaceName }}
+                -consul-namespace={{ .Values.syncCatalog.consulNamespaces.consulNamespaceName }} \
+                {{- end }}
+                {{- if .Values.syncCatalog.consulNamespaces.mirrorK8s }}
```
lkysow: Bug here, this should be `mirrorK8S`

values.yaml

values.yaml

values.yaml
adilyse: An example is a great idea here. Updated.

values.yaml

values.yaml

values.yaml
adilyse: Sure, that sounds good. Updated.

templates/sync-catalog-deployment.yaml
adilyse: Updated.

templates/sync-catalog-deployment.yaml
adilyse: It could, but I prefer to be explicit. This also follows the pattern set elsewhere in the Helm chart.

values.yaml

values.yaml

values.yaml
adilyse: Added a note.

templates/sync-catalog-deployment.yaml
```diff
@@ -117,6 +123,18 @@ spec:
                 {{- if .Values.syncCatalog.addK8SNamespaceSuffix}}
                 -add-k8s-namespace-suffix \
                 {{- end}}
+                {{- if .Values.global.consulNamespacesEnabled }}
+                -enable-namespaces=true \
+                {{- if .Values.syncCatalog.consulNamespaces.consulDestinationNamespace }}
+                -consul-namespace={{ .Values.syncCatalog.consulNamespaces.consulDestinationNamespace }} \
```
lkysow: Looking at the old options, we had:
* `-k8s-service-prefix` - `syncCatalog.k8sPrefix`
* `-k8s-source-namespace` - `syncCatalog.k8sSourceNamespace`
* `-k8s-write-namespace` - `${NAMESPACE}` (comes from the release namespace)
* `-consul-service-prefix`- `syncCatalog.consulPrefix`

We're now adding:
* `-consul-namespace` - `consulNamespaces.consulDestinationNamespace`
* `-enable-namespace-mirroring` - `consulNamespaces.mirrorK8S`
* `-mirroring-prefix` - `consulNamespaces.mirroringPrefix`

Based on the old naming conventions, these should actually be:
* `-consul-write-namespace` - `consulNamespaces.consulWriteNamespace`
* `-enable-k8s-namespace-mirroring` - `consulNamespaces.mirrorK8S` (Specifying this is k8s to consul in the flag is important if we ever want to support consul to k8s in the future)
* `-k8s-namespace-mirroring-prefix` - `consulNamespaces.mirroringPrefix`

So to sum up the changes:
* `-consul-namespace` => `-consul-write-namespace`
* `consulNamespaces.consulDestinationNamespace` => `consulNamespaces.consulWriteNamespace`
* `-enable-namespace-mirroring` => `-enable-k8s-namespace-mirroring`
* `-mirroring-prefix` => `-k8s-namespace-mirroring-prefix`

What do you think?

test/unit/sync-catalog-deployment.bats
```diff
@@ -415,3 +415,167 @@ load _helpers
...
+      --set 'syncCatalog.enabled=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command' | tee /dev/stderr)
+
+  local actual=$(echo $object |
+    yq 'map(select(test("allow-namespace"))) | length' | tee /dev/stderr)
+  [ "${actual}" = "1" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("allow-namespace=\"*\""))' | tee /dev/stderr)
```
lkysow: These should all test for the leading `-` on the flag right? i.e. `'any(contains("-allow-namespace=\"*\""))'`

templates/sync-catalog-deployment.yaml
```diff
@@ -117,6 +123,18 @@ spec:
                 {{- if .Values.syncCatalog.addK8SNamespaceSuffix}}
                 -add-k8s-namespace-suffix \
                 {{- end}}
+                {{- if .Values.global.consulNamespacesEnabled }}
+                -enable-namespaces=true \
+                {{- if .Values.syncCatalog.consulNamespaces.consulDestinationNamespace }}
```
lkysow: What do you think about only setting the `-consul-namespace` flag *iff* mirrorK8S is `false`. This will have the same effect (mirroring overrides the flag) but if users look at the generated pod spec it will make a bit more sense as to what's going to occur.

test/unit/sync-catalog-deployment.bats
```diff
@@ -415,3 +415,167 @@ load _helpers
...
+  [ "${actual}" = "true" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("enable-namespace-mirroring=true"))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+
+  local actual=$(echo $object |
+    yq 'any(contains("mirroring-prefix=k8s-"))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
```
lkysow: shouldn't all these tests have `-` for the flags?

lkysow (commented):
Looking good. After looking at the flags again I noticed some discrepancies with our old flags. Wondering what you think about that.

Other than that the tests are good, just a question about why `-` isn't used.

values.yaml
```diff
@@ -412,11 +421,57 @@ syncCatalog:
...
+  # Note: `k8sDenyNamespaces` takes precedence over values defined here.
+  # Requires consul-k8s v0.12+
+  k8sAllowNamespaces: ["*"]
+
+  # k8sDenyNamespaces is a list of k8s namespaces that should not have their
+  # service synced. This list takes precedence over `k8sAllowNamespaces`.
+  # `*` is not supported because then nothing would be allowed to sync.
+  # Requires consul-k8s v0.12+.
+  #
+  # For example, if `k8sAllowNamespaces is `["*"]` and k8sDenyNamespaces is
```
```suggestion
  # For example, if k8sAllowNamespaces is `["*"]` and k8sDenyNamespaces is
```

PR #338, opened 2020-01-17
Add replicas and disruptionBudget for connect-inject

rrondeau:
This PR aims to make the connect-inject service highly available by default.
Only one pod to deliver pod injection for a service mesh is not enought for us.
We encountered injection problems when lots of pod are restarting during application upgrades in ours clusters.
I understand this PR change a default behavior.
If you think this is a breaking change, i can revert back the default replicas value.
Thanks

### DIFF
--- a/templates/connect-inject-deployment.yaml
+++ b/templates/connect-inject-deployment.yaml
@@ -13,7 +13,7 @@ metadata:
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
 spec:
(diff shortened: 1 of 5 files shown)


david-yu:
Hi @rrondeau thanks for the PR, this is also now defaulted 2 in the connect inject webhook deployment as of Helm 0.32.0+. I'll go ahead and close this PR! Please re-open if you feel that is not the case.

PR #345, opened 2020-01-31
feat: snapshot to persistentVolume and s3 bucket

### LINKED ISSUE #257: Backing and Restoring Consul Snapshot on S3 from Kubernetes
Can someone provide a instructions creating backup on S3 when Consul is hosted on K8s.

### LINKED ISSUE #344: Support for snapshot cronjob
It would be great if this chart could also install a cronjob that ran `consul snapshot` and uploaded it to s3 storage.

tomwganem:
This PR adds a cronjob that creates a snapshot using the consul http api. The snapshot can be saved to a pvc, or uploaded to an s3 bucket.

Tested with global.bootstrapACLs=true and global.tls.enabled=true. Since snapshotting requires a management acl token, I decided to use the bootstrap token that is generated.

Unit tests have been provided

Closes #344 
Closes #257

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -61,3 +61,18 @@ Inject extra environment vars in the format key:value, if populated
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Inject extra environment populated by secrets, if populated
+*/}}
+{{- define "consul.extraSecretEnvironmentVars" -}}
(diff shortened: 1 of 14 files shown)


adilyse:
Thank you for the time and attention evident in this PR, but per https://github.com/hashicorp/consul-helm/issues/344#issuecomment-608070702, this functionality is not currently something we can support in OSS.

PR #346, opened 2020-02-01
Allow setting your own CA as a kube secret

ishustava:
This PR adds the following values to the Helm chart to allow bringing your CA as a Kubernetes secret:
```yaml
tls:    
  caCert:
    secretName: ""
    secretKey: tls.crt
  caKey:
    secretName: ""
    secretKey: tls.key
```

It assumes the defaults for secret keys to follow the pattern of the Kubernetes TLS secrets. This also significantly simplifies the underlying implementation in that we don't need to check if the key is set and can instead assume it's always set.

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -68,12 +68,26 @@ spec:
           configMap:
             name: {{ template "consul.fullname" . }}-client-config
         {{- if .Values.global.tls.enabled }}
-        - name: tls-ca-cert
+        - name: consul-ca-cert
           secret:
+            {{- if .Values.global.tls.caCert.secretName }}
+            secretName: {{ .Values.global.tls.caCert.secretName }}
+            {{- else }}
             secretName: {{ template "consul.fullname" . }}-ca-cert
-        - name: tls-ca-key
+            {{- end }}
+            items:
+            - key: {{ default "tls.crt" .Values.global.tls.caCert.secretKey }}
+              path: tls.crt
+        - name: consul-ca-key
           secret:
+            {{- if .Values.global.tls.caKey.secretName }}
+            secretName: {{ .Values.global.tls.caKey.secretName }}
+            {{- else }}
             secretName: {{ template "consul.fullname" . }}-ca-key
+            {{- end }}
+            items:
+            - key: {{ default "tls.key" .Values.global.tls.caKey.secretKey }}
+              path: tls.key
         - name: tls-client-cert
           emptyDir:
             # We're using tmpfs here so that
@@ -191,7 +205,7 @@ spec:
             - name: config
               mountPath: /consul/config
             {{- if .Values.global.tls.enabled }}
-            - name: tls-ca-cert
+            - name: consul-ca-cert
               mountPath: /consul/tls/ca
               readOnly: true
             - name: tls-client-cert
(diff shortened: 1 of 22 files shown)


templates/client-daemonset.yaml
```diff
@@ -70,10 +70,18 @@ spec:
         {{- if .Values.global.tls.enabled }}
         - name: tls-ca-cert
           secret:
+            {{- if .Values.global.tls.caCert.secretName }}
+            secretName: {{ .Values.global.tls.caCert.secretName }}
+            {{- else }}
             secretName: {{ template "consul.fullname" . }}-ca-cert
+            {{- end }}
```
lkysow: If we do
```yaml
  volumes:
    - name: tls-ca-cert
      secret:
        {{- if .Values.global.tls.caCert.secretName }}
        secretName: {{ .Values.global.tls.caCert.secretName }}
        {{- else }}
        secretName: {{ template "consul.fullname" . }}-ca-cert
        {{- end }}
        items:
        - key: {{ default "tls.crt" .Values.global.tls.caCert.secretKey }}
          path: tls.crt
```

Then the ca cert will always end up in the file `tls.crt`. What do you think about that pattern instead of defaulting the `secretKey` to `tls.crt` in `values.yaml`? I think I like it better because the file system always looks the same, regardless of how it's being initialized.

ishustava: I really like this idea, thanks! I've updated everything to use this pattern instead.

templates/client-daemonset.yaml
```diff
@@ -126,7 +140,7 @@ spec:
             - name: CONSUL_HTTP_ADDR
               value: https://localhost:8501
             - name: CONSUL_CACERT
-              value: /consul/tls/ca/tls.crt
+              value: /consul/tls/ca/{{ .Values.global.tls.caCert.secretKey }}
```
lkysow: I think this is now unnecessary with the changes you made for `path`.

templates/tests/test-runner.yaml
```diff
@@ -30,7 +34,7 @@ spec:
         - name: CONSUL_HTTP_ADDR
           value: https://$(HOST_IP):8501
         - name: CONSUL_CACERT
-          value: /consul/tls/ca/tls.crt
+          value: /consul/tls/ca/{{ .Values.global.tls.caCert.secretKey }}
```
lkysow: not needed?

values.yaml
```diff
@@ -111,6 +111,33 @@ global:
     # clients and servers and only accept HTTPS connections.
     httpsOnly: true
 
+    # caCert is a kubernetes secret containing the certificate
+    # of the CA to use for TLS communications within the Consul cluster.
```
```suggestion
    # of the CA to use for TLS communication within the Consul cluster.
```

values.yaml
```diff
@@ -111,6 +111,33 @@ global:
     # clients and servers and only accept HTTPS connections.
     httpsOnly: true
 
+    # caCert is a kubernetes secret containing the certificate
```
```suggestion
    # caCert is a Kubernetes secret containing the certificate
```

values.yaml
```diff
@@ -111,6 +111,33 @@ global:
...
+    # of the CA to use for TLS communications within the Consul cluster.
+    # If you have generated the CA yourself with the consul CLI,
+    # you could use the following command to create the secret in Kubernetes:
+    #
+    #   kubectl create secret generic consul-ca-cert \
+    #           --from-file='tls.crt=./consul-agent-ca.pem'
+    caCert:
+      secretName: ""
```
lkysow: We've been using `null` for these in other places

lkysow (approved):
Looks good. We should wait until after namespaces to merge. I've been testing this with the wan federation work.

values.yaml
ishustava: I think I looked at gossip encryption which happens to be the only one that uses empty strings 😄 . It seems like `null` is more common; I'll update.

PR #349, opened 2020-02-10
update envoy to 1.13.0

i0rek:


values.yaml
```diff
@@ -640,7 +640,7 @@ meshGateway:
     additionalSpec: null
 
   # Envoy image to use.
```
lkysow: Do we need a note saying if you're running Consul version x you need envoy xxx?

adilyse: Consul 1.7+ needs envoy 1.13+? Or 1.11+?

adilyse:
Closing in favor of #355 that doesn't have an extra commit from master.

PR #356, opened 2020-02-21
Update changelog for namespaces

adilyse:


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 ## Unreleased
 
+IMPROVEMENTS:
+
+  * Add allow/deny namespace lists to catalog sync and connect inject
+
+  * [Enterprise feature] Support Consul namespaces through the Helm chart
+
+  * Updates default version of Envoy to 1.13.0, Consul to 1.7.1 and consul-k8s to 0.12.0.
```
ishustava: Could we add the corresponding PR references?

ishustava (commented):
@adilyse I think we should add some/all of the notes you've added to the consul-k8s Changelog. Especially things that change their behavior are important to repeat here because most folks read Helm chart Changelog/release notes.

lkysow (approved):
🎉 nice catch on the namespaces deprecation

CHANGELOG.md
```diff
@@ -1,5 +1,96 @@
 ## Unreleased
 
+BREAKING CHANGES:
+
+* `consul-k8s` `v0.12.0`+ is now required. The chart is passing new flags that are only available in this version.
+  To use this version, set
```
ishustava: This is the default, right? so we don't need to tell people to set it.

ishustava (approved):
Looks good. We should also mention that we're now changing ACL policies if namespaces are enabled.

CHANGELOG.md
adilyse: Updated the phrasing to reflect that.

adilyse:
Copied the acl policy updating language from the consul-k8s changelog.

CHANGELOG.md
lkysow: Yes it's the default but I'd guess that some folks pin it in their local values.yaml so they can upgrade the chart without also upgrading consul-k8s without doing it on purpose.

PR #357, opened 2020-02-21
update server statefulset port

like-inspur:
add serflan and serfwan udp port


like-inspur:
#364

lkysow:
This isn't needed because we're executing via `sh -c`

adilyse:
Per the comment in #364, this is only an issue when using an `args` array directly. Since the Helm chart uses these variables by executing a shell command, the environment variable substitution is working as intended.

PR #365, opened 2020-03-02
Document envoy version requirements

lkysow:


values.yaml
```diff
@@ -542,11 +542,14 @@ connectInject:
...
-  # default will be used. If the Envoy image is not specified, an early
-  # version of Envoy will be used.
+  # The Docker image for Consul to use when performing Connect injection.
+  # Defaults to global.image.
   imageConsul: null
+
+  # The Docker image for envoy to use as the proxy sidecar when performing
+  # Connect injection. If using Consul 1.7+, the envoy version must be 1.13+.
+  # If not set, the image used depends on the consul-k8s version. For
+  # consul-k8s 0.12.0 the default is envoyproxy/envoy-alpine:v1.13.0.
```
adilyse: Is there a reference doc somewhere we could link to so that folks can find out what the defaults are for other versions?

Not strictly necessary, given that they can always be prescriptive about the value, but seems like something we should document.

adilyse (approved):
Looks good!

values.yaml
lkysow: There isn't a reference doc for consul-k8s. The best we can do right now is link to the line in the code where the constant is declared but that's super sketchy.

I've been meaning to do a write-up of a suggestion to have the helm chart always pass the -envoy-image flag to consul-k8s and have it be the single source of truth for the envoy image because of this issue.

PR #368, opened 2020-03-03
Support ACL replication.

lkysow:
Support creating an ACL replication token (for primary dcs) via `global.acls.createReplicationToken` and
referencing a pre-created ACL replication token (for secondary dcs) via `global.acls.replicationToken`.

One thing to review here is that we're using the existence of `global.acls.replicationToken` to determine whether we set other config flags that are required for replication. We *could* have had a separate setting, e.g. `global.acls.enableReplication`.

### DIFF
--- a/templates/server-acl-init-job.yaml
+++ b/templates/server-acl-init-job.yaml
@@ -32,18 +32,28 @@ spec:
     spec:
       restartPolicy: Never
       serviceAccountName: {{ template "consul.fullname" . }}-server-acl-init
-      {{- if .Values.global.tls.enabled }}
+      {{- if (or .Values.global.tls.enabled (and .Values.global.acls.replicationToken.secretName .Values.global.acls.replicationToken.secretKey)) }}
       volumes:
-      - name: consul-ca-cert
-        secret:
-          {{- if .Values.global.tls.caCert.secretName }}
-          secretName: {{ .Values.global.tls.caCert.secretName }}
-          {{- else }}
-          secretName: {{ template "consul.fullname" . }}-ca-cert
-          {{- end }}
-          items:
-          - key: {{ default "tls.crt" .Values.global.tls.caCert.secretKey }}
-            path: tls.crt
+        {{- if .Values.global.tls.enabled }}
+        - name: consul-ca-cert
+          secret:
+            {{- if .Values.global.tls.caCert.secretName }}
+            secretName: {{ .Values.global.tls.caCert.secretName }}
+            {{- else }}
+            secretName: {{ template "consul.fullname" . }}-ca-cert
+            {{- end }}
+            items:
+              - key: {{ default "tls.crt" .Values.global.tls.caCert.secretKey }}
+                path: tls.crt
+        {{- end }}
+        {{- if (and .Values.global.acls.replicationToken.secretName .Values.global.acls.replicationToken.secretKey) }}
+        - name: acl-replication-token
+          secret:
+            secretName: {{ .Values.global.acls.replicationToken.secretName }}
+            items:
+              - key: {{ .Values.global.acls.replicationToken.secretKey }}
+                path: acl-replication-token
+        {{- end }}
       {{- end }}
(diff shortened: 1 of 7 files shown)


templates/server-acl-init-job.yaml
```diff
@@ -96,6 +103,13 @@ spec:
                 {{- if not (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}
                 -create-client-token=false \
                 {{- end }}
+                {{- if .Values.global.acls.createReplicationToken }}
+                -create-acl-replication-token=true \
+                {{- end }}
+                {{- if (and .Values.global.acls.replicationToken.secretName .Values.global.acls.replicationToken.secretKey) }}
+                -enable-acl-replication=true \
+                -datacenter={{ .Values.global.datacenter }} \
```
lkysow: This flag is only needed if we're in replication mode. I put it here to make the helm chart backwards compatible with older versions of consul-k8s that don't have this flag. For single-dc users, this flag isn't needed.

ishustava (approved):
Luke, this looks good. A couple of thoughts:
- The datacenter flag could be retrieved from the server by the ACL bootstrapping job itself. This is optional though and definitely should not block the merge.
- When it comes to using the replication token to determine if this is the secondary datacenter, I like it. I personally have a bias towards having less configuration options.
- Right now the `bootstrapACLs` and the `acl.*` properties don't interact. I know you've mentioned that this is something you're thinking to change in the future PR, but I thought it's worth noting it here so we don't forget.

templates/server-acl-init-job.yaml
```diff
@@ -96,6 +103,13 @@ spec:
                 {{- if not (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}
                 -create-client-token=false \
                 {{- end }}
+                {{- if .Values.global.acls.createReplicationToken }}
+                -create-acl-replication-token=true \
```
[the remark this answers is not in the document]
lkysow: Yeah for sure but I think we're using `=true` on all the boolean flags here (except `-use-https` 😉 ) so was trying to be consistent.

lkysow:
Updated to remove `-datacenter`. Another PR for `bootstrapACLs` deprecation is coming. Companion PR: https://github.com/hashicorp/consul-k8s/pull/226

lkysow:
Oh, also changed to use the `-acl-replicaton-token-file` flag.

values.yaml
```diff
@@ -156,6 +156,20 @@ global:
...
+  # Configure ACLs.
+  acls:
+    # If true, an ACL token will be created that can be used in secondary
+    # datacenters for replication. This should only be set to true in the
+    # primary datacenter since the replication token must be created from that
+    # datacenter. It will then be imported into secondary datacenters.
+    createReplicationToken: false
+
+    # replicationToken references a secret containing the replication ACL token.
+    # This token will be used by secondary datacenters to perform ACL replication.
```
```suggestion
    # This token will be used by secondary datacenters to perform ACL replication
    # and create ACL tokens and policies.
```

values.yaml
```diff
@@ -156,6 +156,20 @@ global:
     # This setting must be true in both primary and secondary datacenters.
     enabled: false
 
+  # Configure ACLs.
+  acls:
+    # If true, an ACL token will be created that can be used in secondary
+    # datacenters for replication. This should only be set to true in the
+    # primary datacenter since the replication token must be created from that
+    # datacenter. It will then be imported into secondary datacenters.
```
ishustava: It'd be nice to point to `replicationToken` setting here and say that this is the value you should set in the secret for the replication token.

ishustava (approved):
Looks good! I've tested this PR in combination with hashicorp/consul-k8s#226; everything worked great!

PR #375, opened 2020-03-05
Support auto-encrypt

ishustava:
- Adds a new property `global.tls.enableAutoEncrypt`
- Setting the `global.tls.enableAutoEncrypt` will enable auto-encrypt for clients and servers
- consul-k8s components that need to talk to the clients (connect injector,
mesh gateway, sync catalog, and snapshot agent) now get the CA through the API from the Consul server before they start.

Requires hashicorp/consul-k8s#211

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -74,16 +74,30 @@ Get Consul client CA to use when auto-encrypt is enabled
     - |
       consul-k8s get-consul-client-ca \
         -output-file=/consul/tls/client/ca/tls.crt \
-        {{- if .Values.client.join }}
-        -server-addr={{ quote (first .Values.client.join) }} \
-        -tls-server-name=server.{{ .Values.global.datacenter }}.{{ .Values.global.domain }} \
+        {{- if .Values.externalServer.enabled }}
+        {{- if not (or .Values.externalServer.https.address .Values.client.join)}}{{ fail "either client.join or externalServer.https.address must be set if externalServer.enabled is true" }}{{ end -}}
+        {{- if .Values.externalServer.https.address }}
+        -server-addr={{ .Values.externalServer.https.address }} \
         {{- else }}
-        -server-addr=https://{{ template "consul.fullname" . }}-server:8501 \
+        -server-addr={{ quote (first .Values.client.join) }} \
         {{- end }}
+        -server-port={{ .Values.externalServer.https.port }} \
+        {{- if .Values.externalServer.https.tlsServerName }}
+        -tls-server-name={{ .Values.externalServer.https.tlsServerName }} \
+        {{- end }}
+        {{- if not .Values.externalServer.https.useSystemRoots }}
         -ca-file=/consul/tls/ca/tls.crt
+        {{- end }}
+        {{- else }}
+        -server-addr={{ template "consul.fullname" . }}-server \
+        -server-port=8501 \
+        -ca-file=/consul/tls/ca/tls.crt
+        {{- end }}
   volumeMounts:
+    {{- if not (and .Values.externalServer.enabled .Values.externalServer.https.useSystemRoots) }}
     - name: consul-ca-cert
       mountPath: /consul/tls/ca
+    {{- end }}
     - name: consul-auto-encrypt-ca-cert
       mountPath: /consul/tls/client/ca
 {{- end -}}
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -63,7 +63,8 @@ Inject extra environment vars in the format key:value, if populated
 {{- end -}}
 
 {{/*
-Get Consul client CA to use when auto-encrypt is enabled
+Get Consul client CA to use when auto-encrypt is enabled.
+This template is for an init container.
 */}}
 {{- define "consul.getAutoEncryptClientCA" -}}
 - name: get-auto-encrypt-client-ca
@@ -74,18 +75,18 @@ Get Consul client CA to use when auto-encrypt is enabled
     - |
       consul-k8s get-consul-client-ca \
         -output-file=/consul/tls/client/ca/tls.crt \
-        {{- if .Values.externalServer.enabled }}
-        {{- if not (or .Values.externalServer.https.address .Values.client.join)}}{{ fail "either client.join or externalServer.https.address must be set if externalServer.enabled is true" }}{{ end -}}
-        {{- if .Values.externalServer.https.address }}
-        -server-addr={{ .Values.externalServer.https.address }} \
+        {{- if .Values.externalServers.enabled }}
+        {{- if not (or .Values.externalServers.https.address .Values.client.join)}}{{ fail "either client.join or externalServers.https.address must be set if externalServers.enabled is true" }}{{ end -}}
+        {{- if .Values.externalServers.https.address }}
+        -server-addr={{ .Values.externalServers.https.address }} \
         {{- else }}
         -server-addr={{ quote (first .Values.client.join) }} \
         {{- end }}
-        -server-port={{ .Values.externalServer.https.port }} \
-        {{- if .Values.externalServer.https.tlsServerName }}
-        -tls-server-name={{ .Values.externalServer.https.tlsServerName }} \
+        -server-port={{ .Values.externalServers.https.port }} \
+        {{- if .Values.externalServers.https.tlsServerName }}
+        -tls-server-name={{ .Values.externalServers.https.tlsServerName }} \
         {{- end }}
-        {{- if not .Values.externalServer.https.useSystemRoots }}
+        {{- if not .Values.externalServers.https.useSystemRoots }}
         -ca-file=/consul/tls/ca/tls.crt
         {{- end }}
         {{- else }}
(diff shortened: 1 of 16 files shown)


lkysow (commented):
I was mostly through my review so going to leave it anyway in case it's helpful.

templates/client-daemonset.yaml
```diff
@@ -158,13 +166,20 @@ spec:
                 -hcl='leave_on_terminate = true' \
                 {{- if .Values.global.tls.enabled }}
                 -hcl='ca_file = "/consul/tls/ca/tls.crt"' \
+                {{- if .Values.global.tls.enableAutoEncrypt }}
+                -hcl='auto_encrypt = {tls = true}' \
```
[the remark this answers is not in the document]
ishustava: Good find! Yeah, we can move it to the config map in this case. Do you feel it should be done in this PR or maybe do a sweeping change in a separate PR? If separate, then it might make sense to create an issue for it.

templates/client-daemonset.yaml
```diff
@@ -267,7 +284,7 @@ spec:
                 - |
                   {{- if .Values.global.tls.enabled }}
                   curl \
-                    --cacert /consul/tls/ca/tls.crt \
+                    -k \
```
ishustava: Note that I've made the readiness check less secure. This was intentional. My thinking is that it's checking on `localhost` within the pod network security boundary, and the traffic within that boundary is trusted, so it's safe to skip validation. If we wanted to provide the CA, it'd significantly complicate client startup because they'd now need to wait for servers to start, which could increase downtime for Connect-enabled apps, for example, during upgrades.

templates/client-daemonset.yaml
```diff
@@ -316,14 +333,16 @@ spec:
             mv {{ .Values.global.datacenter }}-client-{{ .Values.global.domain }}-0.pem tls.crt
             mv {{ .Values.global.datacenter }}-client-{{ .Values.global.domain }}-0-key.pem tls.key
         volumeMounts:
-          - name: tls-client-cert
+          - name: consul-client-cert
```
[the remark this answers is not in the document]
ishustava: Yes! 🤦‍♀

templates/client-snapshot-agent-deployment.yaml
```diff
@@ -132,6 +142,29 @@ spec:
           - name: aclconfig
             mountPath: /consul/aclconfig
       {{- end }}
+      {{- if (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt) }}
+      - name: get-consul-client-ca
+        image: {{ .Values.global.imageK8S }}
+        command:
+          - "/bin/sh"
+          - "-ec"
```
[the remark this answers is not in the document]
ishustava: I thought `-e` will make the pod exit immediately if any of the commands fail. Although, it's also not strictly needed here because there's only one command.

test/unit/client-daemonset.bats
```diff
@@ -444,6 +444,17 @@ load _helpers
...
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/client-daemonset.yaml  \
+      --set 'global.tls.enabled=true' \
+      --set 'global.tls.enableAutoEncrypt=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.volumes[] | select(.name == "tls-client-cert")' | tee /dev/stderr)
+  [ "${actual}" == "" ]
+}
+
```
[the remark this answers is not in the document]
ishustava: At first, I wrote some of those tests while developing it, but then saw that we're not testing any other command options. IMO it's not super useful.

test/unit/client-snapshot-agent-deployment.bats
```diff
@@ -269,3 +269,52 @@ load _helpers
...
+  local actual=$(helm template \
+      -x templates/client-snapshot-agent-deployment.yaml  \
+      --set 'client.snapshotAgent.enabled=true' \
+      --set 'global.bootstrapACLs=true' \
+      --set 'global.tls.enabled=true' \
+      --set 'global.tls.enableAutoEncrypt=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.initContainers | length == 2' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
```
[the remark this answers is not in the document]
ishustava: Yeah, I will add them.

test/unit/client-daemonset.bats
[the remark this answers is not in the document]
ishustava: On the second look though, I see that we *do* test all command options. I'll add command option tests for clients and servers.

templates/client-daemonset.yaml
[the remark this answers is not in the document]
lkysow: Definitely not this pr! Was just thinking about it.

values.yaml
```diff
@@ -88,6 +88,12 @@ global:
   tls:
     enabled: false
 
+    # enableAutoEncrypt turns on the auto-encrypt feature on
+    # clients and servers.
+    # It also switches consul-k8s components to retrieve the CA
+    # from the servers via the API.
+    enableAutoEncrypt: false
```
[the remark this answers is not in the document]
ishustava: We decided not to because we want to get more mileage on this feature, but could change the default in the future.

templates/_helpers.tpl
```diff
@@ -61,3 +61,43 @@ Inject extra environment vars in the format key:value, if populated
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Get Consul client CA to use when auto-encrypt is enabled
```
lkysow: I'd mention this template is for an init container

values.yaml
```diff
@@ -254,6 +261,41 @@ server:
     # https_proxy: http://localhost:3128,
     # no_proxy: internal.domain.com
 
+
+# Add configuration for Consul servers running externally,
+# i.e. outside of Kubernetes.
+# This information is required if Consul servers are running
+# outside of k8s and you’re setting global.tls.enableAutoEncrypt to true.
+externalServer:
```
lkysow: I know we usually go singular but this reads weird to me. Is it referring to having a single external server? I think this should be `externalServers`

test/unit/client-daemonset.bats
```diff
@@ -561,7 +570,7 @@ load _helpers
 @test "client/DaemonSet: sets Consul environment variables when global.tls.enabled" {
   cd `chart_dir`
   local env=$(helm template \
-      -x templates/server-statefulset.yaml  \
+      -x templates/client-daemonset.yaml  \
```
lkysow: 🍺

lkysow (commented):
Small things, moving on to actually testing now.

values.yaml
ishustava: It's referring to any number of servers; it could be 1 or more. I originally proposed plural in the RFC, but changed it to singular because of other values. Thinking about this more though, these values seem to refer to "features" and in that sense, the feature here would be `externalServers`.

lkysow: 👍 I think `externalServers` makes sense then.

values.yaml
```diff
@@ -254,6 +261,41 @@ server:
     # https_proxy: http://localhost:3128,
     # no_proxy: internal.domain.com
 
+
```
lkysow: one too many newlines here

values.yaml
```diff
@@ -254,6 +261,41 @@ server:
...
+  # HTTPS configuration for external servers.
+  # Note: HTTP connections to the servers are
+  # not supported.
+  https:
+    # IP, DNS name, or Cloud auto-join string
+    # pointing to the external Consul servers.
+    # Note that if you’re providing the cloud
+    # auto-join string and multiple addresses
+    # can be returned, only the first address
+    # will be used.
```
lkysow: Should you note that if client.join is set then this isn't required?

values.yaml
```diff
@@ -254,6 +261,41 @@ server:
...
+    address: null
+
+    # The HTTPS port of the server.
+    port: 443
+
+    # tlsServerName is the server name to use as the SNI
+    # host header when connecting with HTTPS.
+    # This property is useful in case ‘externalServers.https.address’
```
lkysow: > in case ‘externalServers.https.address’

even you knew this was better 😆

templates/client-daemonset.yaml
```diff
@@ -129,6 +131,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
+            - name: HOST_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
```
lkysow: can you also set `CONSUL_HTTP_SSL_VERIFY=false` if autoencrypt is used since the ca cert doesn't exist.

lkysow (approved):
🎉 everything works in my tests:
  - [x] HCS autoencrypt
  - [x] autoencrypt with servers on k8s
  - [x] gke w/ pod security policies
  - [x] acls enabled

templates/client-daemonset.yaml
ishustava: Yeah, good idea. At first, I thought it's a bad idea because skipping SSL verification is always a bad idea, but since it's for local traffic, we're not violating any security assumptions.

PR #376, opened 2020-03-06
Add custom annotation for service

jb-abbadie:
This will allow users of [External DNS](https://github.com/kubernetes-sigs/external-dns) to create a record with the Pod IP of the servers.

### DIFF
--- a/templates/server-service.yaml
+++ b/templates/server-service.yaml
@@ -15,6 +15,9 @@ metadata:
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
   annotations:
+    {{- if .Values.server.service.annotations }}
+    {{ tpl .Values.server.service.annotations . | nindent 4 | trim }}
+    {{- end }}
     # This must be set in addition to publishNotReadyAddresses due
     # to an open issue where it may not work:
     # https://github.com/kubernetes/kubernetes/issues/58662
--- a/values.yaml
+++ b/values.yaml
@@ -245,6 +245,11 @@ server:
   # the annotations to apply to the server pods
   annotations: null
 
+  service:
+    # This should be a multi-line string mapping directly to the a map of
+    # the annotations to apply to the server service
+    annotations: null
+
   # extraEnvVars is a list of extra environment variables to set with the stateful set. These could be
   # used to include proxy settings required for cloud auto-join feature,
   # in case kubernetes cluster is behind egress http proxies. Additionally, it could be used to configure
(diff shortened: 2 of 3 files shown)


values.yaml
```diff
@@ -245,6 +245,11 @@ server:
   # the annotations to apply to the server pods
   annotations: null
 
+  # Extra annotations to attach to the server service
+  # This should be a multi-line string mapping directly to the a map of
+  # the annotations to apply to the server services
+  serviceAnnotations: null
```
lkysow: Can you make it like our other service annotation keys.
```
service:
  # Optional YAML string for additional annotations.
   # This should be a multi-line string of
  # the annotations to apply to the server service.
  annotations: null
```

templates/server-service.yaml
```diff
@@ -15,6 +15,9 @@ metadata:
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
   annotations:
+    {{- if .Values.server.serviceAnnotations }}
```
lkysow: Can you add a test to `test/unit/server-service.bats` similar to the mesh gateway test:
```bash
# annotations

@test "meshGateway/Service: no annotations by default" {
  cd `chart_dir`
      -x templates/mesh-gateway-service.yaml  \
      --set 'meshGateway.enabled=true' \
      --set 'connectInject.enabled=true' \
      --set 'client.grpc=true' \
      --set 'meshGateway.service.enabled=true' \
      . | tee /dev/stderr |
      yq -r '.metadata.annotations' | tee /dev/stderr)
  [ "${actual}" = "null" ]
}

@test "meshGateway/Service: can set annotations" {
  cd `chart_dir`
      -x templates/mesh-gateway-service.yaml  \
      --set 'meshGateway.enabled=true' \
      --set 'connectInject.enabled=true' \
      --set 'client.grpc=true' \
      --set 'meshGateway.service.enabled=true' \
      --set 'meshGateway.service.annotations=key: value' \
      . | tee /dev/stderr |
      yq -r '.metadata.annotations.key' | tee /dev/stderr)
  [ "${actual}" = "value" ]
}

```

templates/server-service.yaml
```diff
@@ -15,6 +15,9 @@ metadata:
...
+      {{- tpl .Values.server.serviceAnnotations . | nindent 4 }}
```
lkysow:
```suggestion
    {{- tpl .Values.server.serviceAnnotations . | nindent 4 | trim }}
```

lkysow:
Thanks for the PR! Just a few comments.

jb-abbadie:
Fixed + Rebased based on the comments.

PR #378, opened 2020-03-09
Update annotation docs

lkysow:
Fix some of the wording on the annotation docs and add examples since using a multi-line YAML string isn't necessarily obvious.


ishustava (approved):
Looks good!

PR #380, opened 2020-03-11
Add lifecycle sidecar to mesh gateway

lkysow:
When Consul clients in Kubernetes restart, they lose their
registrations. This causes any mesh gateways to be deregistered. To
solve this, we need to run a sidecar that ensures the service is always
registered: a lifecycle-sidecar.

For the lifecycle-sidecar to work, it needs a service.hcl service config
file. This required adding an init container that writes this file and
does the initial service registration. Since the init container is
registering the service, the consul connect envoy -mesh-gateway command
no longer needs the -register flag.

### DIFF
--- a/templates/mesh-gateway-deployment.yaml
+++ b/templates/mesh-gateway-deployment.yaml
@@ -1,6 +1,7 @@
 {{- if .Values.meshGateway.enabled }}
 {{- if not .Values.connectInject.enabled }}{{ fail "connectInject.enabled must be true" }}{{ end -}}
 {{- if not .Values.client.grpc }}{{ fail "client.grpc must be true" }}{{ end -}}
+{{- if and .Values.global.bootstrapACLs (ne .Values.meshGateway.consulServiceName "") (ne .Values.meshGateway.consulServiceName "mesh-gateway") }}{{ fail "if global.bootstrapACLs is true, meshGateway.consulServiceName cannot be set" }}{{ end -}}
 {{- /* The below test checks if clients are disabled (and if so, fails). We use the conditional from other client files and prepend 'not' */ -}}
 {{- if not (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}{{ fail "clients must be enabled" }}{{ end -}}
 apiVersion: apps/v1
@@ -48,6 +49,9 @@ spec:
       volumes:
         - name: consul-bin
           emptyDir: {}
+        - name: consul-service
+          emptyDir:
+            medium: "Memory"
         {{- if .Values.global.tls.enabled }}
         - name: consul-ca-cert
           secret:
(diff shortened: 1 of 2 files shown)


lkysow:
For testing:
* Use `consul-k8s` image `lkysow/consul-k8s-dev:mar16-2020-2` that has the new `-token-sink-file` flag for `acl-init` (https://github.com/hashicorp/consul-k8s/pull/232)
* Start the mesh gateways, then delete the consul client pod on that node. The mesh gateway should be re-registered momentarily.

templates/mesh-gateway-deployment.yaml
```diff
@@ -1,6 +1,7 @@
 {{- if .Values.meshGateway.enabled }}
 {{- if not .Values.connectInject.enabled }}{{ fail "connectInject.enabled must be true" }}{{ end -}}
 {{- if not .Values.client.grpc }}{{ fail "client.grpc must be true" }}{{ end -}}
+{{- if and .Values.global.bootstrapACLs (ne .Values.meshGateway.consulServiceName "") (ne .Values.meshGateway.consulServiceName "mesh-gateway") }}{{ fail "if global.bootstrapACLs is true, meshGateway.consulServiceName cannot be set" }}{{ end -}}
```
lkysow: Moved this up here from where it was previously inside the command definition since the -service-name flag was removed from the command definition.

templates/mesh-gateway-deployment.yaml
```diff
@@ -79,21 +82,92 @@ spec:
...
+                {{- if .Values.meshGateway.wanAddress.host }}
+                WAN_ADDR="{{ .Values.meshGateway.wanAddress.host }}"
+                {{- else if .Values.meshGateway.wanAddress.useNodeName }}
+                WAN_ADDR="${NODE_NAME}"
+                {{- else if .Values.meshGateway.wanAddress.useNodeIP }}
+                WAN_ADDR="${HOST_IP}"
+                {{- end }}
+
+                cat > /consul/service/service.hcl << EOF
+                service {
```
ishustava: Just thinking, the mgw service needs to be in the default namespace, right? Do you think it's worth saying so explicitly?

templates/mesh-gateway-deployment.yaml
```diff
@@ -79,21 +82,92 @@ spec:
...
+                  kind = "mesh-gateway"
+                  name = "{{ default "mesh-gateway" .Values.meshGateway.consulServiceName }}"
+                  {{- if .Values.global.federation.enabled }}
+                  meta {
+                    consul-wan-federation = "1"
```
ishustava: I'm a little worried about this one. It feels like we're copying Consul's internal implementation here. Do you think it'll be supported it in a backwards compatible way?

templates/mesh-gateway-deployment.yaml
```diff
@@ -79,21 +82,92 @@ spec:
...
+                  }
+                  {{- end }}
+                  port = {{ .Values.meshGateway.containerPort }}
+                  address = "${POD_IP}"
+                  tagged_addresses {
+                    lan {
+                      address = "${POD_IP}"
+                      port = {{ .Values.meshGateway.containerPort }}
+                    }
+                    wan {
```
ishustava: FYI 
It looks like Consul UI is also showing the `wan_ipv4` address in the `Addresses` tab. It sets it to the lan address.
Screen Shot 2020-03-19 at 5 04 37 PM

templates/mesh-gateway-deployment.yaml
```diff
@@ -197,6 +252,44 @@ spec:
               exec:
                 command: ["/bin/sh", "-ec", "/consul-bin/consul services deregister -id=\"{{ default "mesh-gateway" .Values.meshGateway.consulServiceName }}\""]
 
+        # lifecycle-sidecar ensures the mesh gateway is always registered with
+        # the local Consul agent, even if it loses the initial registration.
+        - name: lifecycle-sidecar
+          image: {{ .Values.global.imageK8S }}
+          volumeMounts:
+            - name: consul-service
+              mountPath: /consul/service
```
ishustava: Could you make it readOnly?

ishustava (commented):
Luke, it generally looks good. Great job on all those pesky tests! I left a few comments with suggestions and questions. Let me know what you think

templates/mesh-gateway-deployment.yaml
lkysow: It makes sense if running Consul enterprise but if running OSS you'll get
```
Error: Error parsing service.hcl: 1 error occurred:
	* invalid config key service.namespace
```

lkysow: So we'd need a conditional around if enableNamespaces. I'm not sure it's worth it since we'd need a conditional and it has no effect but happy to add it.

templates/mesh-gateway-deployment.yaml
lkysow: This is the recommended say to do it according to RB. Here's the docs he's working on: https://github.com/hashicorp/consul/pull/7430/files#diff-58cf12d8599883137a5e562eae01a2d8R127

templates/mesh-gateway-deployment.yaml
lkysow: Wow good catch. I'll add `wan_ipv4`!

templates/mesh-gateway-deployment.yaml
lkysow: done

templates/mesh-gateway-deployment.yaml
ishustava: Phew, great. Thanks for looking into it!

templates/mesh-gateway-deployment.yaml
ishustava: Got ya. It makes sense to me then to keep as is. Ty!

ishustava (approved):
Looks good to me 🎉 ! Thanks for addressing my comments 🌮

PR #381, opened 2020-03-12
Remove meshGateway.consulServiceName value

lkysow:
The value was not supported when ACLs were enabled and we didn't see the
need for it at this time.


lkysow:
Decided against this because it may be needed for HCS.

PR #382, opened 2020-03-12
Deprecate bootstrapACLs for acls.enabled

lkysow:
With the new `global.acls` setting, it was confusing to have bootstrapACLs
as a separate setting. Instead we create a new `global.acls.manageSystemACLs`
setting which can be used in place of global.bootstrapACLs.

For backwards compatibility, I've replaced all instances of `if .global.bootstrapACLs` with `if (or .global.acls.manageSystemACLs .global.bootstrapACLs)`. This means users previously setting `global.bootstrapACLs: true` will see no changes.

For testing, I've replaced all use of `global.bootstrapACLs` with `global.acls.manageSystemACLs` because creating a new test for each use would have ended up with 100 new tests.

### DIFF
--- a/templates/client-snapshot-agent-deployment.yaml
+++ b/templates/client-snapshot-agent-deployment.yaml
@@ -37,7 +37,7 @@ spec:
       {{- if .Values.client.priorityClassName }}
       priorityClassName: {{ .Values.client.priorityClassName | quote }}
       {{- end }}
-      {{- if (or .Values.global.bootstrapACLs .Values.global.tls.enabled (and .Values.client.snapshotAgent.configSecret.secretName .Values.client.snapshotAgent.configSecret.secretKey)) }}
+      {{- if (or .Values.global.acls.manageSystemACLs .Values.global.bootstrapACLs .Values.global.tls.enabled (and .Values.client.snapshotAgent.configSecret.secretName .Values.client.snapshotAgent.configSecret.secretKey)) }}
       volumes:
         {{- if (and .Values.client.snapshotAgent.configSecret.secretName .Values.client.snapshotAgent.configSecret.secretKey) }}
         - name: snapshot-config
@@ -47,7 +47,7 @@ spec:
             - key: {{ .Values.client.snapshotAgent.configSecret.secretKey }}
               path: snapshot-config.json
         {{- end }}
-        {{- if .Values.global.bootstrapACLs }}
+        {{- if (or .Values.global.acls.manageSystemACLs .Values.global.bootstrapACLs) }}
         - name: aclconfig
           emptyDir: {}
         {{- end }}
@@ -88,7 +88,7 @@ spec:
             - name: CONSUL_HTTP_ADDR
               value: http://$(HOST_IP):8500
             {{- end }}
-            {{- if .Values.global.bootstrapACLs }}
+            {{- if (or .Values.global.acls.manageSystemACLs .Values.global.bootstrapACLs) }}
             - name: CONSUL_HTTP_TOKEN
               valueFrom:
                 secretKeyRef:
@@ -103,17 +103,17 @@ spec:
                 {{- if (and .Values.client.snapshotAgent.configSecret.secretName .Values.client.snapshotAgent.configSecret.secretKey) }}
                 -config-dir=/consul/config \
                 {{- end }}
-                {{- if .Values.global.bootstrapACLs}}
+                {{- if (or .Values.global.acls.manageSystemACLs .Values.global.bootstrapACLs) }}
                 -config-dir=/consul/aclconfig \
                 {{- end }}
-          {{- if (or .Values.global.bootstrapACLs .Values.global.tls.enabled (and .Values.client.snapshotAgent.configSecret.secretName .Values.client.snapshotAgent.configSecret.secretKey) ) }}
+          {{- if (or .Values.global.acls.manageSystemACLs .Values.global.bootstrapACLs .Values.global.tls.enabled (and .Values.client.snapshotAgent.configSecret.secretName .Values.client.snapshotAgent.configSecret.secretKey) ) }}
           volumeMounts:
             {{- if (and .Values.client.snapshotAgent.configSecret.secretName .Values.client.snapshotAgent.configSecret.secretKey) }}
             - name: snapshot-config
               readOnly: true
               mountPath: /consul/config
             {{- end }}
-            {{- if .Values.global.bootstrapACLs}}
+            {{- if (or .Values.global.acls.manageSystemACLs .Values.global.bootstrapACLs) }}
             - name: aclconfig
               mountPath: /consul/aclconfig
             {{- end }}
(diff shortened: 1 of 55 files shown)


templates/client-snapshot-agent-deployment.yaml
```diff
@@ -127,9 +127,9 @@ spec:
               readOnly: true
           {{- end }}
           {{- end }}
-      {{- if (or .Values.global.bootstrapACLs (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt)) }}
+      {{- if (or (or .Values.global.acls.manageSystemACLs .Values.global.bootstrapACLs) (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt)) }}
```
adilyse: As you demonstrated elsewhere, it's possible to have three inputs to `or`.

```suggestion
      {{- if (or .Values.global.acls.manageSystemACLs .Values.global.bootstrapACLs (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt)) }}
```

templates/mesh-gateway-deployment.yaml
```diff
@@ -1,7 +1,7 @@
 {{- if .Values.meshGateway.enabled }}
 {{- if not .Values.connectInject.enabled }}{{ fail "connectInject.enabled must be true" }}{{ end -}}
 {{- if not .Values.client.grpc }}{{ fail "client.grpc must be true" }}{{ end -}}
-{{- if and .Values.global.bootstrapACLs (ne .Values.meshGateway.consulServiceName "") (ne .Values.meshGateway.consulServiceName "mesh-gateway") }}{{ fail "if global.bootstrapACLs is true, meshGateway.consulServiceName cannot be set" }}{{ end -}}
+{{- if and (or .Values.global.acls.manageSystemACLs .Values.global.bootstrapACLs) (ne .Values.meshGateway.consulServiceName "") (ne .Values.meshGateway.consulServiceName "mesh-gateway") }}{{ fail "if global.acls.manageSystemACLs is true, meshGateway.consulServiceName cannot be set" }}{{ end -}}
```
adilyse: This error message won't help someone track down what's going wrong if they're using `bootstrapACLs`. Though, this is likely an error someone would hit when first setting things up, and then they should be using the other flag. So it's probably okay.

templates/sync-catalog-deployment.yaml
```diff
@@ -176,9 +176,9 @@ spec:
             periodSeconds: 5
             successThreshold: 1
             timeoutSeconds: 5
-      {{- if or .Values.global.bootstrapACLs (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt) }}
+      {{- if or (or .Values.global.acls.manageSystemACLs .Values.global.bootstrapACLs) (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt) }}
```
adilyse: Same switch to a 3 input `or` command that's used elsewhere.

```suggestion
      {{- if or .Values.global.acls.manageSystemACLs .Values.global.bootstrapACLs (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt) }}
```

values.yaml
```diff
@@ -808,7 +813,8 @@ meshGateway:
   # If mesh gateways are enabled, a Deployment will be created that runs
   # gateways and Consul Connect will be configured to use gateways.
   # See https://www.consul.io/docs/connect/mesh_gateway.html
-  # Requirements: consul >= 1.6.0 and consul-k8s >= 0.9.0 if using global.bootstrapACLs.
+  # Requirements: consul >= 1.6.0 and consul-k8s >= 0.9.0 if using
+  # global.acls.manageSystemACLs.
```
```suggestion
  # `global.acls.manageSystemACLs.`
```

values.yaml
```diff
@@ -892,8 +898,8 @@ meshGateway:
   dnsPolicy: null
 
   # Override the default 'mesh-gateway' service name registered in Consul.
-  # Cannot be used if bootstrapACLs is true since the ACL token generated
-  # is only for the name 'mesh-gateway'.
+  # Cannot be used if global.acls.manageSystemACLs is true since the ACL token
```
```suggestion
  # Cannot be used if `global.acls.manageSystemACLs` is true since the ACL token
```

values.yaml
```diff
@@ -766,15 +770,16 @@ connectInject:
   # Requires Consul >= v1.5 and consul-k8s >= v0.8.0.
   aclBindingRuleSelector: "serviceaccount.name!=default"
 
-  # If not using global.bootstrapACLs and instead manually setting up an auth
-  # method for Connect inject, set this to the name of your auth method.
+  # If not using global.acls.manageSystemACLs and instead manually setting up an
```
```suggestion
  # If not using `global.acls.manageSystemACLs` and instead manually setting up an
```

values.yaml
```diff
@@ -163,8 +158,17 @@ global:
     # This setting must be true in both primary and secondary datacenters.
     enabled: false
 
+  # [DEPRECATED] Use acls.manageSystemACLs instead.
```
```suggestion
  # [DEPRECATED] Use `global.acls.manageSystemACLs` instead.
```

values.yaml
```diff
@@ -38,7 +38,7 @@ global:
   # Note: support for the catalog sync's liveness and readiness probes was added
   # to consul-k8s 0.6.0. If using an older consul-k8s version, you may need to
   # remove these checks to make the sync work.
-  # If using bootstrapACLs then must be >= 0.10.1.
+  # If using acls.manageSystemACLs then must be >= 0.10.1.
```
```suggestion
  # If using `global.acls.manageSystemACLs` then must be >= 0.10.1.
```

adilyse (commented):
I understand why you repurposed the bootstrapACL tests, but this leaves us with an entirely untested Helm chart option which really isn't great. Not sure where the middle ground is, but having no testing isn't good.

Otherwise, it looks good, just a few minor suggestions.

templates/mesh-gateway-deployment.yaml
lkysow: Yeah I thought putting verbiage about how it could be triggered by either bootstrapACLs or manageSystemACLs would make it more confusing. I'm hoping they'd be able to trace it through since they'd be like "I'm not setting manageSystemACLs", look at that flag and realize that bootstrapACLs was deprecated.

lkysow:
Test was added in https://github.com/hashicorp/consul-helm/pull/382/commits/1712d6ec99f0ab87f7fcad0d818f61177a41a7ba. This test looks a bit hacky but actually has better coverage than if we were to test the first instance of bootstrapacls use in each file.

test/unit/helpers.bats
```diff
@@ -257,4 +259,22 @@ load _helpers
...
+#
+# Test that every place global.bootstrapACLs is used, global.acls.manageSystemACLs
+# is also used.
+# If this test is failing, you've used either only global.bootstrapACLs or
+# only global.acls.manageSystemACLs instead of using the backwards compatible:
+#     or global.acls.manageSystemACLs global.bootstrapACLs
+@test "helper/bootstrapACLs: used alongside manageSystemACLs" {
+  cd `chart_dir`
+
+  diff=$(diff <(grep -r '\.Values\.global\.bootstrapACLs' templates/*) <(grep -r '\.Values\.global\.acls\.manageSystemACLs \.Values\.global\.bootstrapACLs' templates/*) | tee /dev/stderr)
```
adilyse, line 275: This is a great way to test this out! As extra insurance that we're using things correctly, would it make sense to make sure that these two values are associated with an `or`?

So, using `grep -r 'or \.Values\.global\.acls\.manageSystemACLs \.Values\.global\.bootstrapACLs' templates/*` instead. There are two places that would need an order change for this to work:

```
< templates/client-clusterrole.yaml:{{- if (or .Values.global.enablePodSecurityPolicies .Values.global.acls.manageSystemACLs .Values.global.bootstrapACLs) }}

< templates/client-snapshot-agent-clusterrole.yaml:{{- if (or .Values.global.enablePodSecurityPolicies .Values.global.acls.manageSystemACLs .Values.global.bootstrapACLs) }}
```

adilyse (approved):
The grep test is a great way to go! I added a suggestion to be able to include the `or` in the check. It wouldn't be strictly necessary, but adds an extra layer of insurance.

test/unit/helpers.bats
lkysow: Yeah that's safer. Done!

PR #384, opened 2020-03-16
Added liveness probe for Consul agents

ibrahimalihc:
This change adds a liveness probe for the Consul agents to ensure the Pod crashes if the Consul agent is unable to join the cluster. The [/health/node/:node](https://www.consul.io/api/health.html#list-checks-for-node) endpoint is able to be used only because the node name is set as an environment variable here:
* https://github.com/hashicorp/consul-helm/blob/v0.17.0/templates/client-daemonset.yaml#L114
* https://github.com/hashicorp/consul-helm/blob/v0.17.0/templates/client-daemonset.yaml#L139

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -257,6 +257,24 @@ spec:
             - containerPort: 8600
               name: dns-udp
               protocol: "UDP"
+          livenessProbe:
+            exec:
+              command:
+                - "/bin/sh"
+                - "-ec"
+                - |
+                  curl \
+                    -k -s --show-error --fail \
+                  {{- if .Values.global.tls.enabled }}
+                    https://127.0.0.1:8501/v1/health/node/$NODE
+                  {{- else }}
+                    http://127.0.0.1:8500/v1/health/node/$NODE
+                  {{- end }}
+            initialDelaySeconds: 15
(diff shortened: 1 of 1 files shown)


ishustava:
Hey @ibrahimalihc,

Thanks so much for this PR 🙏 👍 🎉 

As we've talked about offline, for the time being we will continue relying on just the readiness probe. My reasoning is this:
1. I could see that it can make sense when the agent is misconfigured, however, currently there is no way for us to tell apart cases when the agent is misconfigured from cases when agent can't join the cluster for other reasons, such as network connectivity. In case of a network connectivity issue, it doesn't make sense to crash the pod since the agent should eventually recover once the network connectivity is established.
1. In some cases (e.g. if not using auto-encrypt), the client agent could still be partially functioning. For example, the `/v1/agent/` API will work, although most consul-k8s features, like sync catalog and Consul connect, will not be usable. In these situations, it also doesn't make sense to crash the pod, and instead, rely on the agent to self-recover whenever it can.
1. The point that @lkysow brought up: agents will continue to work if there's no leader with caching reasonably well, but with the liveness probe in place, Kube might crash the pod in case leader election cannot be established for too long.

I hope this makes sense and let me know if I missed anything!

PR #388, opened 2020-03-17
Support LoadBalancerAddress for mesh gateways

lkysow:
Add support for using the external address of Kubernetes load balancer
for the mesh gateway wan address.

This change uses the new consul-k8s load-balancer-address command to get
the address of the load balancer.

It also removes meshGateway.wanAddress.{useNodeIP, useNodeName, host}
config values in favour of meshGateway.wanAddress.{source, static}. The
new source value allows selecting NodeIP, NodeName, LoadBalancerAddress
or Static. This is more extensible than the previous boolean values.

### DIFF
--- a/templates/mesh-gateway-deployment.yaml
+++ b/templates/mesh-gateway-deployment.yaml
@@ -116,12 +116,35 @@ spec:
                   -token-sink-file=/consul/service/acl-token
                 {{ end }}
 
-                {{- if .Values.meshGateway.wanAddress.host }}
-                WAN_ADDR="{{ .Values.meshGateway.wanAddress.host }}"
-                {{- else if .Values.meshGateway.wanAddress.useNodeName }}
-                WAN_ADDR="${NODE_NAME}"
-                {{- else if .Values.meshGateway.wanAddress.useNodeIP }}
+                {{- $source := .Values.meshGateway.wanAddress.source }}
+                {{- $serviceType := .Values.meshGateway.service.type }}
+                {{- if and (eq $source "Service") (not .Values.meshGateway.service.enabled) }}{{ fail "if meshGateway.wanAddress.source=Service then meshGateway.service.enabled must be set to true" }}{{ end }}
+                {{- if or (eq $source "NodeIP") (and (eq $source "Service") (eq $serviceType "NodePort")) }}
                 WAN_ADDR="${HOST_IP}"
+                {{- else if eq $source "NodeName" }}
+                WAN_ADDR="${NODE_NAME}"
+                {{- else if and (eq $source "Service") (or (eq $serviceType "ClusterIP") (eq $serviceType "LoadBalancer")) }}
+                consul-k8s service-address \
+                  -k8s-namespace={{ .Release.Namespace }} \
+                  -name={{ template "consul.fullname" . }}-mesh-gateway \
+                  -output-file=address.txt
+                WAN_ADDR="$(cat address.txt)"
+                {{- else if eq $source "Static" }}
+                {{- if eq .Values.meshGateway.wanAddress.static "" }}{{ fail "if meshGateway.wanAddress.source=Static then meshGateway.wanAddress.static cannot be empty" }}{{ end }}
+                WAN_ADDR="{{ .Values.meshGateway.wanAddress.static }}"
+                {{- else }}
+                {{- fail "currently set meshGateway values for wanAddress.source and service.type are not supported" }}
+                {{- end }}
+
+                {{- if eq $source "Service" }}
+                {{- if eq $serviceType "NodePort" }}
+                {{- if not .Values.meshGateway.service.nodePort }}{{ fail "if meshGateway.wanAddress.source=Service and meshGateway.service.type=NodePort, meshGateway.service.nodePort must be set" }}{{ end }}
+                WAN_PORT="{{ .Values.meshGateway.service.nodePort }}"
+                {{- else }}
+                WAN_PORT="{{ .Values.meshGateway.service.port }}"
+                {{- end }}
+                {{- else }}
+                WAN_PORT="{{ .Values.meshGateway.wanAddress.port }}"
                 {{- end }}
 
                 cat > /consul/service/service.hcl << EOF
@@ -140,17 +163,9 @@ spec:
                       address = "${POD_IP}"
                       port = {{ .Values.meshGateway.containerPort }}
                     }
-                    lan_ipv4 {
-                      address = "${POD_IP}"
-                      port = {{ .Values.meshGateway.containerPort }}
-                    }
                     wan {
                       address = "${WAN_ADDR}"
-                      port = {{ .Values.meshGateway.wanAddress.port }}
-                    }
-                    wan_ipv4 {
-                      address = "${WAN_ADDR}"
-                      port = {{ .Values.meshGateway.wanAddress.port }}
+                      port = ${WAN_PORT}
                     }
                   }
                   checks = [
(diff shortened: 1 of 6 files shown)


lkysow:
Reviewers:
* Review https://github.com/hashicorp/consul-helm/pull/380 first because this PR relies on the new structure for the mesh-gateway deployment.
* Needs consul-k8s image built from https://github.com/hashicorp/consul-k8s/pull/234

ishustava (commented):
Luke, I haven't looked at the code yet, but decided to try it out first. 

So far, it didn't work because the mesh gateway clusterrole needs to be updated to allow it to read services. The error I get is:
```
2020-03-19T22:14:46.604Z [ERROR] getting service iryna-consul-mesh-gateway: services "iryna-consul-mesh-gateway" is forbidden: User "system:serviceaccount:default:iryna-consul-mesh-gateway" cannot get resource "services" in API group "" in the namespace "default"
```


I also had a thought on the UX. I initially looked at the config and tried to figure out which values I need to set to enable the Load Balancer. I set `meshGateway.service.enabled` to `true` and `meshGateway.service.type` to `LoadBalancer`. But that didn't trigger the mesh gateway deployment to fetch the load balancer address. I then realized that I need to the `meshGateway.wanAddress.source` to `LoadBalancerAddress`. I'm wondering if my stupidity is an indicator that the UX could be a bit more obvious. It would be super nice if we could infer from the service settings what you most likely would want. 

One idea on how to do that could be enabling service by default and adding an option to `wanAddress.source` called something like `service`. This option would imply that we'll use the service for wan address, and perhaps, even the port. It might also make sense to change the default service type from ClusterIP to something else. IIRC routable cluster IPs are pretty niche, and so I'm not sure if this is a sane default for the mesh gateway. Just an idea, but I'm curious what you think (and apologies for the long story).

lkysow:
Ack sorry I missed that in my commit, I had it locally. You need
```
  - apiGroups: [""]
    resources:
      - services
    resourceNames:
      - {{ template "consul.fullname" . }}-mesh-gateway
    verbs:
      - get
```
I'll update the PR.

ishustava:
Update: my tests were successful after adding the clusterrole rules. Barring what I said about UX improvements, I think this is good to go, but would like to hear what you think first.

lkysow:
Updated to have `wanAddress.source=Service` be smarter. See `values.yaml` docs for new behaviour. Needs https://github.com/hashicorp/consul-k8s/pull/235 since we now support `ClusterIP` services via the `load-balancer-address` => `service-address` command

ishustava:
@lkysow I found a problem with my own suggestion to add `wan_ipv4` tagged address. When I tried it on EKS, I got this error in the `service-init` container:

```
Address "a143257ad85f14ba6a4e526cf2ae9760-2083317689.us-west-2.elb.amazonaws.com" written to address.txt successfully
Error registering service "mesh-gateway": Unexpected response code: 500 (Service tagged address "wan_ipv4" must be a valid ipv4 address)
```

values.yaml
```diff
@@ -766,46 +766,53 @@ meshGateway:
...
 
     # Port that the service will be exposed on.
     # The targetPort will be set to meshGateway.containerPort.
     port: 443
 
-    # Optional nodePort of the service. Can be used in conjunction with
-    # type: NodePort.
+    # Optionally hardcode the nodePort of the service if using type: NodePort.
+    # If not set and using type: NodePort, Kubernetes will automatically assign
+    # a port.
```
ishustava: Right now, the helm chart fails if I don't have this set:
```
Error: template: consul/templates/mesh-gateway-deployment.yaml:141:69: executing "consul/templates/mesh-gateway-deployment.yaml" at <fail "if meshGateway.wanAddress.source=Service and meshGateway.service.type=NodePort, meshGateway.service.nodePort must be set">: error calling fail: if meshGateway.wanAddress.source=Service and meshGateway.service.type=NodePort, meshGateway.service.nodePort must be set
```

values.yaml
```diff
@@ -766,46 +766,53 @@ meshGateway:
...
-    # useNodeIP and host must be false and "" respectively.
-    useNodeName: false
-
-    # If set, each gateway Pod will use this host as its wan address.
-    # Users must ensure that this address routes to the Gateway pods,
-    # for example via a DNS entry that routes to the Service fronting the Deployment.
-    # meshGateway.wanAddress.port will be used as the port for the wan address.
-    # useNodeIP and useNodeName must be false.
-    host: ""
+    # If source is set to "Static" then this value will be used as the wan
```
```suggestion
    # If source is set to "Static" then this value will be used as the WAN
```

values.yaml
```diff
@@ -766,46 +766,53 @@ meshGateway:
...
-    # Port that gets registered.
+    # source configures where to retrieve the WAN address (and possibly port)
+    # for the mesh gateway from.
+    # Can be set to either: Service, NodeIP, NodeName or Static.
+    #
+    # Service - Determine the address based on the service type.
+    #   If service.type=LoadBalancer use the external IP or hostname of
+    #   the service. Use the port set by service.port.
+    #   If service.type=NodePort use the Node IP. The port will be set to
+    #   service.nodePort so this cannot be null.
```
ishustava: What is "this" referring to here that cannot be null?

values.yaml
```diff
@@ -766,46 +766,53 @@ meshGateway:
...
+    #   If service.type=ClusterIP use the ClusterIP. The port will be set to
+    #   service.port.
+    #   service.type=ExternalName is not supported.
+    # NodeIP - The node ip as provided by the Kubernetes downward API.
```
```suggestion
    # NodeIP - The node IP as provided by the Kubernetes downward API.
```

ishustava (commented):
Thanks for making the changes to support the different kinds of services! I've left a couple of thoughts so far.

values.yaml
lkysow: `service.nodePort`. Will make clearer.

values.yaml
lkysow: Yes because we have to set the port in the addresses section of the service definition so we need to know what it is ahead of time

ishustava: Sure, I understand. I think we need to say that we require it in the comment. But maybe, since Kube doesn't require it, we could get it by running the `service-address` command?

lkysow: It would need to be a separate command/flag since it needs to be templated separately into the service definition.

After talking we decided to leave it required for now and we can later make it optional.

lkysow:
Removed all `_ipv4` addresses. I've also tested on EKS, AKS and GKE now. I've set the default `containerPort` to be `8443` instead of `443` because I was getting bind permission errors on GKE.

templates/mesh-gateway-deployment.yaml
```diff
@@ -116,12 +116,35 @@ spec:
...
+                consul-k8s service-address \
+                  -k8s-namespace={{ .Release.Namespace }} \
+                  -name={{ template "consul.fullname" . }}-mesh-gateway \
+                  -output-file=address.txt
+                WAN_ADDR="$(cat address.txt)"
+                {{- else if eq $source "Static" }}
+                {{- if eq .Values.meshGateway.wanAddress.static "" }}{{ fail "if meshGateway.wanAddress.source=Static then meshGateway.wanAddress.static cannot be empty" }}{{ end }}
+                WAN_ADDR="{{ .Values.meshGateway.wanAddress.static }}"
+                {{- else }}
+                {{- fail "currently set meshGateway values wanAddress.source and service.type are not supported" }}
```
ishustava:
```suggestion
                {{- fail "currently set meshGateway values for wanAddress.source and service.type are not supported" }}
```

ishustava (approved):
Looks good (except for some failing unit tests)!

PR #394, opened 2020-03-23
Remove use of -init-type=sync as it has no effect

lkysow:
As part of https://github.com/hashicorp/consul-k8s/pull/232 we removed
the docs about -init-type=sync. This does not break anything because if -init-type was set to sync it actually had no effect. Only -init-type=client had effect.


ishustava (approved):
Looks good!

Issue #270, opened 2019-10-24
Add `imagePullSecrets` to DaemonSets


errygg:
Can the helm chart be expanded to enable `imagePullSecrets` for any required image pulls from a private repository? In order to pull images from a private repo today we have to export the template and manually add `imagePullSecrets` prior to each container stanza which is not ideal.

PR #402, opened 2020-03-31
Added Image pull secret to all service Accounts

junaid18183:
This PR adds the Image pull secret to all service Accounts. So that the consul images can be pull from private registries


junaid18183:
Fixes - https://github.com/hashicorp/consul-helm/issues/270

PR #406, opened 2020-04-02
ACL replication changelog, remove federation.enabled

lkysow:
* Changelog for latest changes
* Remove `global.federation.enabled` from `values.yaml` since there's no Consul release  that currently supports it. Users that have a custom Docker image can still set it in their local config files and it will work.

In order to not break Helm since the `global.federation` key doesn't exist, I needed to wrap the check for `global.federation.enabled` with a check for `global.federation`. Otherwise we'd get an `<.Values.global.federation.enabled>: nil pointer evaluating interface {}.enabled` error. Gotemplate doesn't support short circuiting so I needed a separate `if` statement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,151 @@
 ## Unreleased
 
+BREAKING CHANGES:
+
+* Mesh Gateways:
+  * `meshGateway.wanAddress` - The following values are no longer supported:
+  
+       ```yaml
+       meshGateway:
+         wanAddress:
+           useNodeIP: true
+           useNodeName: false
+           host: ""
+       ```
+    
+    Instead, if previously setting `useNodeIP: true`, now you must set:
+       ```yaml
+       meshGateway:
+         wanAddress:
+           source: "NodeIP"
+       ```
+    
+    If previously setting `useNodeName: true`, now you must set:
+       ```yaml
+       meshGateway:
+         wanAddress:
+           source: "NodeName"
+       ```
+    
+    If previously setting `host: "example.com"`, now you must set:
+       ```yaml
+       meshGateway:
+         wanAddress:
+           source: "Static"
+           static: "example.com"
+       ```
+    where `meshGateway.wanAddress.static` is set to the previous `host` value.
+  
+  * `meshGateway.service.enabled` now defaults to `true`. If
+    previously you were enabling mesh gateways but not enabling the service,
+    you must now explicitly set this to `false`:
+    
+    Previously:
+    ```yaml
+    meshGateway:
+      enabled: true
+    ```
+    
+    Now:
+    ```yaml
+    meshGateway:
+      enabled: true
+      service:
+        enabled: false
+    ```
+    
+  * `meshGateway.service.type` now defaults to `LoadBalancer` instead of `ClusterIP`.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,135 @@
...
+    you will still be able to route to the mesh gateway as before.
+    If you wish to keep the port as `443` you must set:
+    ```yaml
+    meshGateway:
+      containerPort: 443
+    ```
+
+FEATURES:
+
+* ACLs: Support ACL replication. ACL replication allows two or more Kubernetes clusters
```
ishustava, line 77:
```suggestion
* ACLs: Support ACL replication. ACL replication allows two or more Consul clusters
```

CHANGELOG.md
```diff
@@ -1,5 +1,135 @@
...
+  to be federated when ACLs are enabled. One cluster is designated the primary
+  and the rest are secondaries. The primary cluster replicates its ACLs to
+  the secondaries. [[GH-368](https://github.com/hashicorp/consul-helm/pull/368)]
+  
+  NOTE: This feature requires that the clusters are federated. If the Kubernetes
+  clusters do not share a routable network, then you must use mesh federation.
+  Mesh federation is in tech-preview right now (see above).
```
ishustava, line 84: Should we say that if we're not exposing it in values.yaml?

CHANGELOG.md
```diff
@@ -32,6 +162,24 @@ BUG FIXES:
 * Fix incompatibility with Helm 3.1.2. [[GH-390](https://github.com/hashicorp/consul-helm/issues/390)]
 * Ensure the Consul Enterprise license gets applied, even if servers take a long time to come up. [[GH-348](https://github.com/hashicorp/consul-helm/pull/348))
 
+DEPRECATIONS:
```
ishustava, line 165: Is this for 0.18.0? or just weird git diff?

templates/mesh-gateway-deployment.yaml
```diff
@@ -160,11 +160,13 @@ spec:
                 service {
                   kind = "mesh-gateway"
                   name = "{{ default "mesh-gateway" .Values.meshGateway.consulServiceName }}"
+                  {{- if .Values.global.federation }}
```
ishustava, line 163: The joys of YAML templating 😃

ishustava (approved):
Nice work! I love all the examples 💯 I left a couple of comments and suggestions inline, then it should be good to go.

CHANGELOG.md
lkysow: Oops, I had this in before I removed the federation values. I'm just going to leave it as
>   NOTE: This feature requires that the clusters are federated.

CHANGELOG.md
lkysow: ack, thanks. Meant for this to be under upcoming.

templates/mesh-gateway-deployment.yaml
lkysow: 😆 yep. I was like "why isn't this short-circuiting?!!"

lkysow: 😆 yep, I was like "why isn't this short circuiting!?!"

PR #407, opened 2020-04-02
Run enterprise-license job on upgrades

ishustava:
I think running it on upgrades is a valid use case since someone could be `helm upgrade`ing from OSS to enterprise. The job is idempotent, so there is no risk in running it again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ IMPROVEMENTS:
       enableAutoEncrypt: true
   ```
 
+* Run the enterprise license job on Helm upgrades, as well as installs [[GH-407](https://github.com/hashicorp/consul-helm/pull/407)].
+
 FEATURES:
 
 * Add `externalServers` configuration to support configuring the Helm chart with Consul servers
--- a/templates/enterprise-license-job.yaml
+++ b/templates/enterprise-license-job.yaml
@@ -13,7 +13,7 @@ metadata:
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
   annotations:
-    "helm.sh/hook": post-install
+    "helm.sh/hook": post-install,post-upgrade
     "helm.sh/hook-weight": "100"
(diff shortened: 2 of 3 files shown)


adilyse:
What happens if someone removes their enterprise license secret after applying it for the first time? I could see a scenario where someone audits the k8s secrets and removes things that are no longer needed.

ishustava:
That's an interesting use-case. In that case, someone could delete the secret and disable the job by not setting the secret values. Are you asking in terms of whether to consider this a breaking change?

adilyse:
I guess I was wondering what the user experience would be in that case and whether we're okay with it. If someone set up enterprise awhile ago and has been updating with the chart without any issues after removing the secret (but possibly not knowing they need to remove it from the values file), does this break upgrades for them?

Given that OSS -> Ent upgrade is probably reasonably rare, but so is cleaning up secrets, It's hard to know which is more likely to cause users issues.

ishustava:
Yeah for the use-case you're describing, it'll break the upgrade. Although I'm not sure how common that is and whether it's worth calling this change out as a breaking change. 

Thank you for thinking through these edge cases. While thinking about what you asked, more reasons to support `post-upgrade` license update came to mind. Right now the job assumes that you only need to update the license on install, which is not always true. What happens if the license expires or if you update the license to include more features? It just doesn't seem like the right assumption to me.

I do understand that folks may not want to use Kube as their secret storage, but I feel like, given that right now this is our chosen secrets storage via the Helm chart, it's reasonable to expect that secrets are there.

adilyse:
I'm okay not listing it as a breaking change, but it might be good to add a note to the values file for the enterprise option that says that this is run both on initial install and on upgrade. That should at least point people in the right direction if they had deleted the secret.

ishustava:
Added!

adilyse (approved):
Looks good! Thanks for adding the extra note :)

lkysow:
I used this change just now upgrading from oss to ent and it worked 👍

PR #411, opened 2020-04-06
Image pull secrets

### LINKED ISSUE #270: Add `imagePullSecrets` to DaemonSets

lkysow:
Allow setting image pull secrets for service accounts to enable pulling the consul-k8s and consul images from private registries.

Kube docs: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account

Notes
* Refactors work from https://github.com/hashicorp/consul-helm/pull/402 (thanks!!)
* Closes https://github.com/hashicorp/consul-helm/pull/402
* Closes #270

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -31,6 +31,7 @@ global:
   #   # Consul Enterprise 1.5.0
   #   image: "hashicorp/consul-enterprise:1.5.0-ent"
   image: "consul:1.7.1"
+  imagePullSecrets: 
 
   # imageK8S is the name (and tag) of the consul-k8s Docker image that
   # is used for functionality such as catalog sync. This can be overridden
--- a/values.yaml
+++ b/values.yaml
@@ -31,7 +31,18 @@ global:
   #   # Consul Enterprise 1.5.0
   #   image: "hashicorp/consul-enterprise:1.5.0-ent"
   image: "consul:1.7.1"
-  imagePullSecrets: 
+
+  # array of objects containing image pull secret names that will be applied to
+  # each service account.
+  # This can be used to reference image pull secrets if using
+  # a custom consul or consul-k8s Docker image.
+  # See https://kubernetes.io/docs/concepts/containers/images/#using-a-private-registry.
+  #
+  # Example:
+  #   imagePullSecrets:
+  #   - name: pull-secret-name
+  #   - name: pull-secret-name-2
+  imagePullSecrets: []
 
   # imageK8S is the name (and tag) of the consul-k8s Docker image that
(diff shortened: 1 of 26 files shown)


values.yaml
```diff
@@ -32,6 +32,16 @@ global:
...
   image: "consul:1.7.1"
 
+  # array of image pull secret names that will be applied to each service
+  # account. This can be used to reference image pull secrets if using
+  # a custom consul or consul-k8s Docker image.
+  # See https://kubernetes.io/docs/concepts/containers/images/#using-a-private-registry.
+  #
+  # Example:
+  #   imagePullSecrets:
+  #   - name: my-secret-name
```
adilyse, line 42: These example feels confusing. When empty, the value needs `[]`, but it doesn't need them when I give a value? In the [`extraConfig`](https://github.com/hashicorp/consul-helm/blob/master/values.yaml#L237) text we mention that it's an array of objects, which might help a bit.

adilyse (commented):
Looking good. The one thing I noticed is that the value is defined as an array, but all of the tests only test a single value. It'd be good to make sure that the iteration through multiple values works as expected too.

values.yaml
lkysow: Yeah good call, changed to say "array of objects containing image pull secret names"

lkysow:
> The one thing I noticed is that the value is defined as an array, but all of the tests only test a single value. It'd be good to make sure that the iteration through multiple values works as expected too.

For sure. I've changed the tests to set two image pull secrets and assert that both are set.

adilyse (commented):
Generally when we're testing for several values, each one is checked separately. Example: [env variable](https://github.com/hashicorp/consul-helm/blob/master/test/unit/server-statefulset.bats#L451-L475)

adilyse (commented):
One more question-- should these be quoted? From the [docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account), it looks like they're not being quoted:

```
imagePullSecrets:
- name: myregistrykey
```

lkysow:
It doesn't matter, it's just a string.

adilyse (approved):
Looks good.

Though if the quoting doesn't matter, it's a little odd to go to the effort to add it.

lkysow:
Yeah I'll remove it.

PR #416, opened 2020-04-06
Use copied consul binary

lkysow:
Previously we were testing with a consul-k8s image that has consul in
its path but our regular consul-k8s image does not have the consul
binary included. Instead, we should use the consul binary that gets
copied into the shared volume by the consul-bin init container. init
containers run in order and wait until completion until running the next
init container so this is safe to do.


ishustava (approved):
LGTM

PR #417, opened 2020-04-07
Use /home/consul-k8s as working dir

lkysow:
consul-k8s runs as the consul-k8s user who doesn't have permissions to
write to / since that's owned by root. Instead we use our home directory
where we have permissions to write address.txt and service.hcl files.

### DIFF
--- a/templates/mesh-gateway-deployment.yaml
+++ b/templates/mesh-gateway-deployment.yaml
@@ -136,8 +136,8 @@ spec:
                 consul-k8s service-address \
                   -k8s-namespace={{ .Release.Namespace }} \
                   -name={{ template "consul.fullname" . }}-mesh-gateway \
-                  -output-file=address.txt
-                WAN_ADDR="$(cat address.txt)"
+                  -output-file=/tmp/address.txt
+                WAN_ADDR="$(cat /tmp/address.txt)"
                 {{- else if eq $source "Static" }}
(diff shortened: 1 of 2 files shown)


ishustava (approved):
That makes sense to me. Alternative solution would be to write `address.txt` it to a temp file. 

To confirm, we don't need this to write `service.hcl` since that file is already written to a volume that should have write permissions, correct?

adilyse (approved):
Are there other components that should also have this switched?

Also, do the files that are written to the `/home/consul-k8s` directory in the init container need to be referenced from the main container? (i.e. does the volume mount there need to change?)

lkysow:
> That makes sense to me. Alternative solution would be to write address.txt it to a temp file.

Yeah that's probably better. I'll change it to do that.

> To confirm, we don't need this to write service.hcl since that file is already written to a volume that should have write permissions, correct?

Correct, I'll update the commit and PR comments.

> Are there other components that should also have this switched?

No, there's nowhere else that we're writing to the working dir of the consul-k8s image.

> Also, do the files that are written to the /home/consul-k8s directory in the init container need to be referenced from the main container? (i.e. does the volume mount there need to change?)

The only file being written there is `address.txt` which then gets read into an environment variable which is used to template `service.hcl` which is output to the shared volume. So the location of `address.txt` doesn't actually matter.

PR #421, opened 2020-04-09
 Consul will support envoy 1.14.1 with Consul 1.8

i0rek:
See hashicorp/consul#7624.


david-yu:
We are now on 1.14.2 in our Helm chart! Thanks!

PR #401, opened 2020-03-30
Provide server stateful set DNS names to the ACL init job

ishustava:
This PR goes with hashicorp/consul-k8s#238

PR #427, opened 2020-04-16
ACLs: support external servers meta PR

ishustava:
This PR contains all changes for external servers combined from #401, #420, and #430.


scripts/generate-helm-config-for-hcs.sh
```diff
@@ -0,0 +1,110 @@
...
+
+############################################################################
+# WARNING: This script is experimental and is not meant for production use #
+############################################################################
+#
+# Script to retrieve configuration from Hashicorp Consul Service on Azure,
+# bootstrap ACLs, and create Kubernetes secrets and Helm config file
+# to install this Helm chart.
+
+set -e
```
lkysow, line 11: let's use `set -euo pipefile` (https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/)

scripts/generate-helm-config-for-hcs.sh
```diff
@@ -0,0 +1,110 @@
...
+      secretKey: tls.crt
+externalServers:
+  enabled: true
+  hosts: ${retry_join}
+  httpsPort: 443
+  useSystemRoots: true
+  k8sAuthMethodHost: ${kube_api_server}
+client:
+  enabled: true
+  # If you are using Kubenet in your AKS cluster,
```
lkysow, line 100:
```suggestion
  # If you are using Kubenet in your AKS cluster (the default network),
```

lkysow (approved):
🎉

scripts/generate-helm-config-for-hcs.sh
ishustava: 💯

scripts/generate-helm-config-for-hcs.sh
```diff
@@ -0,0 +1,110 @@
...
+  server_url=$(echo "${cluster_resource}" | jq -r .properties.consulExternalEndpointUrl)
+else
+  server_url=$(echo "${cluster_resource}" | jq -r .properties.consulPrivateEndpointUrl)
+fi
+
+# Call Consul bootstrap API and save the bootstrap secret
+# to a Kubernetes secret if successful.
+output=$(curl -sX PUT "${server_url}"/v1/acl/bootstrap)
+if grep -i "permission denied" <<< "$output"; then
+  echo "ACL system already bootstrapped."
```
lkysow, line 46: I totally missed this in my output. Can you use terminal colour chars to make his red.

lkysow: Even better would be if this prints at the end of the script.

ishustava: Added colors.

templates/client-daemonset.yaml
```diff
@@ -205,11 +205,18 @@ spec:
...
                 {{- end }}
+                {{- if or .Values.client.join (and .Values.externalServers.enabled .Values.externalServers.hosts) }}
                 {{- if .Values.client.join }}
                 {{- range $value := .Values.client.join }}
                 -retry-join={{ quote $value }} \
                 {{- end }}
                 {{- else }}
+                {{- $serfPort := (.Values.externalServers.serfLANPort | int) -}}
+                {{- range $host := .Values.externalServers.hosts }}
+                -retry-join={{ printf "%s:%d" $host $serfPort | quote}} \
```
adilyse, line 216: Echoing the comment I left (quite late) on the other PR: Does this still work if someone set a provider? Or a dns name? I think the docs also mention being able to set a go-sockaddr template.

ishustava: That's a good catch. I don't think this will work. I can change it to default the `serfLANPort` to null and only set it if it's provided in this template. I need to look into go-sockaddr.

lkysow: Nice catch. go-sockaddr templates look like `"{{ GetPrivateIP }}"` so as long as it's being echoed as a string in the case of serfPort being null you should be fine.

adilyse: There's also the (likely quite rare) case of someone having different ports for different servers.

lkysow: In that case they will only be able to set a single server but that's okay because once you reach one server it tells you the locations of the other servers.

adilyse: It'll be a confusing experience though- if they're trying to set multiple servers with different ports but can only set one port, which to choose? We'll also have drifted from Consul's definition of `retry-join`.

This awkwardness is why I recommended that we don't address the `client.join` switch and deprecation now. It means we'll need the port stripping in `consul-k8s`, but it gives us time to get this right.

ishustava: Could that confusion be addressed in a comment? The drift here is intentional because we're no longer shadowing the retry-join configuration. These hosts are used for two purposes/protocols, and so it made more sense to split the port out.

adilyse: There's also the option to not reuse `client.join` here and instead ask for a single server address and https port for this ACLs work. It potentially duplicates a config value, but doesn't run into any of the complications of trying to support all of the `retry-join` forms for clients while restricting what can be set for the acl connection.

That solves this problem without requiring special cases or port stripping in consul-k8s. Very tidy.

adilyse: > These hosts are used for two purposes/protocols

To me that means we should be asking for different configuration settings. We're overly complicating things trying to reuse this value.

values.yaml
```diff
@@ -333,9 +333,6 @@ externalServers:
   # Ports must be provided separately below.
   hosts: []
```
lkysow, line 334:
```
  # An array of external Consul server hosts that are used to make
  # HTTPS connections from the components in this Helm chart.
  # Valid values include IPs, DNS names, or Cloud auto-join string.
  # The port must be provided separately below.
  # NOTE: client.join must also be set to the hosts that should be
  # used to join the cluster. In most cases the client.join values
  # should be the same, however they may be different if you
  # wish to use separate hosts for the HTTPS connections.
```

ishustava: 👍

lkysow (approved):
✅

templates/client-daemonset.yaml
ishustava: For posterity, we've decided to treat retry-join hosts and https hosts as separate properties.

Issue #327, opened 2020-01-09
Sync-Catalog i/o timeout


tpetrychyn:
Using latest master on a fresh Digital Ocean cluster and only changing syncCatalog to enabled: true results in the following error:

```
2020-01-09T21:37:55.787Z [INFO ] to-consul/sink: starting service watcher: service-name=kubernetes-default
[GET /health/ready] Error getting leader status: Get http://10.137.120.213:8500/v1/status/leader: dial tcp 10.137.120.213:8500: i/o timeout
```

Cluster version is `TOR1 - 1.16.2-do.1` all default settings, I created a fresh cluster to test this.

I do know that sync catalog works in v0.8.1 of this chart but is now broken in master. Any help is appreciated!

lkysow:
Hi, what is the output of `kubectl get pods`? Are all the consul clients running ?

tpetrychyn:
```
$ kubectl get pods                                                                                                                                                                                                                                 ✔  7s  do-tor1-another-consul-test ⎈
NAME                                          READY   STATUS             RESTARTS   AGE
consul-consul-7g7bd                           1/1     Running            0          91s
consul-consul-njfwh                           1/1     Running            0          72s
consul-consul-server-0                        1/1     Running            0          11m
consul-consul-server-1                        1/1     Running            0          11m
consul-consul-server-2                        1/1     Running            0          11m
consul-consul-sync-catalog-758c65bbf7-ht4jr   0/1     CrashLoopBackOff   7          11m
consul-consul-zsd4t                           1/1     Running            0          54s
```

Thanks!

tpetrychyn:
Any idea? I pulled master this morning and did another fresh run, again only sync set true. The server and clients no longer log any network issues, they are quiet and running healthy for 100 minutes.

Sync catalog full output looks like 
```Listening on ":8080"...
2020-01-13T16:55:53.670Z [INFO]  to-consul/source: starting runner for endpoints
2020-01-13T16:55:53.670Z [INFO]  to-k8s/sink: starting runner for syncing
2020-01-13T16:55:53.766Z [INFO]  to-consul/controller: initial cache sync complete
2020-01-13T16:55:53.768Z [INFO]  to-k8s/controller: initial cache sync complete
2020-01-13T16:55:53.769Z [INFO]  to-k8s/sink: upsert: key=default/consul-consul-server
2020-01-13T16:55:53.770Z [INFO]  to-k8s/sink: upsert: key=default/consul-consul-ui
[...]

PR #428, opened 2020-04-17
Allow sync-catalog to run against external consul servers

kieranajp:
Hi,

This teeny PR allows running consul-sync from within K8s against Consul servers that are not in the K8s cluster.

Please let me know if this is a feature you'd be happy to merge in, including if adding this to `externalServers` is the right place (this seems to be the intention for that section, even if it's only used for TLS configuration right now).

Cheers,
--Kieran

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -54,6 +54,10 @@ spec:
       {{- if .Values.client.dnsPolicy }}
       dnsPolicy: {{ .Values.client.dnsPolicy }}
       {{- end }}
+      
+      {{- if .Values.client.hostNetwork }}
+      hostNetwork: {{ .Values.client.hostNetwork }}
+      {{- end }}
 
       volumes:
         - name: data
--- a/templates/sync-catalog-deployment.yaml
+++ b/templates/sync-catalog-deployment.yaml
@@ -54,9 +54,13 @@ spec:
           image: "{{ default .Values.global.imageK8S .Values.syncCatalog.image }}"
           env:
             - name: HOST_IP
+            {{- if .Values.global.externalHostIP }}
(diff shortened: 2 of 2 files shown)


adilyse:
Hi @kieranajp,

Thank you for the contribution! Unfortunately several things have changed with the external servers configuration since you submitted this. 

`externalServers.address` has become `externalServers.hosts`, which can be an array of [IPs, DNS names, or a Cloud auto-join string](https://github.com/hashicorp/consul-helm/blob/master/values.yaml#L332). This last option, the auto-join string, would cause a serious issue with the `HOST_IP` setting since it would be considered invalid (see the [Consul docs](https://www.consul.io/docs/agent/options.html#retry-join) for more context on this value).

As to the larger question of having sync catalog work when Consul servers are external to the Kubernetes cluster, the sync catalog process expects to talk to a Consul client located on the same k8s node. It should be possible to run Consul servers outside of Kubernetes, then run Consul clients within Kubernetes and have them connect to the external servers. Then there should be no issues running the catalog sync process, as it will have a local Consul client to talk to.

I'm going to go ahead and close this for now given the above information, but if you have a use case that is not addressed by the current features, please feel free to open an issue or another PR in the future. Thanks again!

kieranajp:
Thanks for the write-up on that, @adilyse! I like the way external servers is shaping up.

I actually did go down the route of running clients in K8s in the end - I hadn't done this at first since they were unfortunately crashing on my setup. This turned out to be related to #327, so I was able to sort it out with the workaround in that PR. 

I've submitted #431 related to the above - unfortunately without that I'm having to use a fork of these charts to run the consul-client successfully. Hopefully that one makes more sense to get merged in. 

Chat to y'all on that thread. 😊

PR #430, opened 2020-04-23
Reorganize externalServers section

ishustava:

* Consolidate `client.join` and `externalServers.https.address` under one value: `externalServers.hosts`. Unlike `client.join`, this value no longer accepts ports, and so if the serf port is non-default, it has to be provided separately. This is because we're now using the same host for multiple protocols, and so I think it's better to provide the ports for each protocol separately. Since it no longer accepts ports, I thought `hosts` will be a more appropriate name for it. This was largely based on [RFC2396](https://tools.ietf.org/html/rfc2396#section-3.2.2),
     where URL is defined as `<userinfo>@<host>:<port>`.
* Deprecate `client.join`.
* Add `externalServers.serfLANPort`.
* Rename `externalServers.https.port` -> `externalServers.httpsPort`
* Rename `externalServers.https.tlsServerName` -> `externalServers.tlsServerName`
* Rename `externalServers.https.useSystemRoots` -> `externalServers.useSystemRoots`
* Rename `connectInject.overrideAuthMethodHost` ->  `externalServers.k8sAuthMethodHost`


values.yaml
```diff
@@ -325,43 +325,46 @@ server:
...
-    # in the client.join property.
-    address: null
-
-    # The HTTPS port of the server.
-    port: 443
+  # An array of external Consul hosts.
+  # Valid values include IPs, DNS names, or Cloud auto-join string.
+  # Ports must be provided separately below.
+  # These hosts are used for clients to join the server cluster,
+  # as well as for HTTP/HTTPS connections from the components in this Helm chart.
```
lkysow, line 334:
```suggestion
  # An array of external Consul server hosts that clients will use to join with servers
  # as well as for HTTP/HTTPS connections from the components in this Helm chart.
  # Valid values include IPs, DNS names, or Cloud auto-join string.
  # Ports must be provided separately below.
```

values.yaml
```diff
@@ -325,43 +325,46 @@ server:
...
+  hosts: []
+
+  # The Serf LAN port that the clients will use to join the server cluster.
+  serfLANPort: 8301
+
+  # The HTTPS port of the Consul servers.
+  httpsPort: 443
```
lkysow, line 341: Really this should default to 8501.

values.yaml
```diff
@@ -325,43 +325,46 @@ server:
...
+  # system CAs for TLS verification when talking to Consul servers.
+  # Otherwise, the chart will use global.tls.caCert.
+  useSystemRoots: false
+
+  # If you are setting global.acls.manageSystemACLs and connectInject.enabled to true,
+  # set k8sAuthMethodHost to the address of the Kubernetes API server.
+  # This address must to be reachable from the Consul servers.
+  # Please see https://www.consul.io/docs/acl/auth-methods/kubernetes.html.
+  # Requires consul-k8s >= 0.14.0.
+  k8sAuthMethodHost: null
```
lkysow, line 358: Can you add an example, e.g. `https://7D96FAF246DABD092E8B0E04B73D967C.sk1.us-east-1.eks.amazonaws.com`

test/unit/client-daemonset.bats
```diff
@@ -86,6 +86,32 @@ load _helpers
...
+      -x templates/client-daemonset.yaml  \
+      --set 'server.enabled=false' \
+      --set 'externalServers.enabled=true' \
+      --set 'externalServers.hosts[0]=1.1.1.1' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.spec.containers[0].command | any(contains("-retry-join=\"1.1.1.1:8301\""))' | tee /dev/stderr)
+
+  [ "${actual}" = "true" ]
+}
+
```
lkysow, line 114: Can you also test:
* setting `externalServers.serfLANPort`
* setting multiple join/host urls

templates/client-daemonset.yaml
```diff
@@ -205,11 +205,18 @@ spec:
...
                 -encrypt="${GOSSIP_KEY}" \
                 {{- end }}
+                {{- if or .Values.client.join (and .Values.externalServers.enabled .Values.externalServers.hosts) }}
                 {{- if .Values.client.join }}
                 {{- range $value := .Values.client.join }}
                 -retry-join={{ quote $value }} \
                 {{- end }}
                 {{- else }}
+                {{- $serfPort := (.Values.externalServers.serfLANPort | int) -}}
+                {{- range $value := .Values.externalServers.hosts }}
```
lkysow, line 215:
```suggestion
                {{- range $host := .Values.externalServers.hosts }}
```

lkysow (approved):

values.yaml
ishustava: Will change this to 8501 for consistency.

values.yaml
ishustava: I'll provide a command to retrieve from your kube config.

templates/client-daemonset.yaml
```diff
@@ -205,11 +205,18 @@ spec:
...
                 {{- end }}
+                {{- if or .Values.client.join (and .Values.externalServers.enabled .Values.externalServers.hosts) }}
                 {{- if .Values.client.join }}
                 {{- range $value := .Values.client.join }}
                 -retry-join={{ quote $value }} \
                 {{- end }}
                 {{- else }}
+                {{- $serfPort := (.Values.externalServers.serfLANPort | int) -}}
+                {{- range $host := .Values.externalServers.hosts }}
+                -retry-join={{ printf "%s:%d" $host $serfPort | quote}} \
```
adilyse, line 216: Does this still work if someone set a provider? Or a dns name? I think the [docs](https://www.consul.io/docs/agent/options.html#retry-join) also mention being able to set a go-sockaddr template.

PR #438, opened 2020-04-27
Use OSS Consul image with HCS

ishustava:


ishustava:
Yep, I tested this with clients being OSS (v1.7.2; the current default).

PR #447, opened 2020-05-05
New create-federation-secret job

lkysow:
This job runs on post-install/upgrade and creates a Kubernetes secret
that can be exported into secondary datacenters. The secondary
datacenters will use the secret to federate with the primary.

Also:
1. Makes `mesh-gateway` the default value for `meshGateway.consulServiceName` rather than using `{{ default "mesh-gateway" .Values.meshGateway.consulServiceName }}` as per our "defaults in Helm" discussion. I did this in this PR because we're also using that value in the federation secret job.
1. Exposes `global.federation.enabled`

## Testing ##
- [x] acls enabled
- [x] acls disabled
- [x] auto-encrypt enabled
- [x] gossip encryption enabled

Base config:
```yaml
global:
  imageK8S: lkysow/consul-k8s-dev:may1-1
  image: ishustava/consul-dev:apr-1-2020
  federation:
    enabled: true
    createFederationSecret: true
  tls:
    enabled: true
    enableAutoEncrypt: false
  acls:
    manageSystemACLs: false
    createReplicationToken: false
  #gossipEncryption:
  #  secretName: consul-gossip-encryption-key
  #  secretKey: key
meshGateway:
  enabled: true
connectInject:
  enabled: true
```

Companion to https://github.com/hashicorp/consul-k8s/pull/253


test/unit/create-federation-secret-job.bats
```diff
@@ -0,0 +1,244 @@
...
+@test "createFederationSecet/Job: fails when global.federation.enabled=false" {
+  cd `chart_dir`
+  run helm template \
+      -x templates/create-federation-secret-job.yaml  \
+      --set 'global.federation.createFederationSecret=true' .
+  [ "$status" -eq 1 ]
+  [[ "$output" =~ "global.federation.enabled must be true when global.federation.createFederationSecret is true" ]]
+}
+
+# NOTE: This error actually comes from server-statefulset but we test it here
```
ishustava, line 23: Not a huge deal, but I think it's more intuitive if we test it in server tests because it should fail even when you set `createFederationSecret` to `false` and the failure code is in the server stateful set too.

values.yaml
```diff
@@ -198,6 +198,25 @@ global:
       secretName: null
       secretKey: null
 
+  # Settings related to federating with other Consul datacenter.
+  federation:
+    # **Experimental: Currently this is only available in Consul 1.8.0-beta.**
+    # If enabled, this datacenter will be federation-capable. Only federation
+    # through mesh gateways is supported.
+    # Mesh gateways and servers will be configured to allow federation.
+    # Requires tls.enabled, meshGateway.enabled and connectInject.enabled
```
ishustava, line 207: When you read it together with `meshGateway.enabled` and `connectInject.enabled`, it could be interpreted as a property that's outside of `global`
```suggestion
    # Requires global.tls.enabled, meshGateway.enabled and connectInject.enabled
```

values.yaml
```diff
@@ -198,6 +198,25 @@ global:
       secretName: null
       secretKey: null
 
+  # Settings related to federating with other Consul datacenter.
```
ishustava, line 201: Grammarly 😀 
```suggestion
  # Settings related to federating with another Consul datacenter.
```

values.yaml
```diff
@@ -198,6 +198,25 @@ global:
...
+  federation:
+    # **Experimental: Currently this is only available in Consul 1.8.0-beta.**
+    # If enabled, this datacenter will be federation-capable. Only federation
+    # through mesh gateways is supported.
+    # Mesh gateways and servers will be configured to allow federation.
+    # Requires tls.enabled, meshGateway.enabled and connectInject.enabled
+    # to be true.
+    enabled: false
+
+    # If true, the chart will create a federation Kubernetes secret that can be imported
```
ishustava, line 211:
```suggestion
    # If true, the chart will create a Kubernetes secret that can be imported
```

ishustava: because you already say that it'll be used for federation later in the sentence

ishustava (commented):
This looks really good. I left some questions/comments and suggestions. One major one in my mind is around not using cluster-wide perms anymore. Since it's a new job, maybe it's time to start?

test/unit/create-federation-secret-job.bats
lkysow: I was thinking it made sense to be in both? If it's not in the server as well I'll add it. I was thinking that because this spec requires that fail to be in place, if it's only tested in the server then if we disable it there we won't know that this job also requires it.

ishustava: Oh! yep it is tested in both places; I misenterpreted the comment. disregard, please 😄

templates/create-federation-secret-job.yaml
```diff
@@ -0,0 +1,128 @@
...
+        chart: {{ template "consul.chart" . }}
+        release: {{ .Release.Name }}
+        component: create-federation-secret
+      annotations:
+        "consul.hashicorp.com/connect-inject": "false"
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ template "consul.fullname" . }}-create-federation-secret
+      volumes:
+        - name: consul-ca-cert
```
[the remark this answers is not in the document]
lkysow: Added right above

test/unit/create-federation-secret-job.bats
lkysow: I've added the missing test to server but kept this one here too. lmk.

templates/create-federation-secret-clusterrole.yaml
```diff
@@ -0,0 +1,49 @@
+{{- if .Values.global.federation.createFederationSecret }}
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
```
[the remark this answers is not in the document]
lkysow: Yep, moved to `Role` and tested.

values.yaml
```diff
@@ -198,6 +198,25 @@ global:
       secretName: null
       secretKey: null
 
+  # Settings related to federating with another Consul datacenter.
+  federation:
+    # **Experimental: Currently this is only available in Consul 1.8.0-beta.**
```
ishustava, line 203: One thing that would be nice is to say what to set the `global.image` once we know what that image would be.

ishustava (approved):
Looks good!

templates/create-federation-secret-clusterrole.yaml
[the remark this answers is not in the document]
ishustava: thank you!

values.yaml
lkysow: 👍

PR #452, opened 2020-05-11
Support terminating gateways

adilyse:
This creates all of the necessary templates to support terminating
gateways in the helm chart. Using `terminatingGateways.defaults` and
`terminatingGateways.gateways`, it's possible to define multiple
gateways. Names must be provided for each and they must be unique.

Any gateway-specific values will override the default value with the
exception of annotations. Annotations will apply both the defaults
and gateway-specific annotations.

TODO: Add tests.

### DIFF
--- a/templates/terminating-gateways-deployment.yaml
+++ b/templates/terminating-gateways-deployment.yaml
@@ -1,0 +1,326 @@
+{{- if .Values.terminatingGateways.enabled }}
+
+{{- $root := . }}
+{{- $defaults := .Values.terminatingGateways.defaults }}
+{{- $names := dict }}
+
+{{- range .Values.terminatingGateways.gateways }}
+
+{{- if empty .name }}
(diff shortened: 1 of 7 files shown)


templates/terminating-gateways-deployment.yaml
```diff
@@ -0,0 +1,324 @@
+{{- if .Values.terminatingGateways.enabled }}
```
lkysow, line 1: Can we use `len(.Values.terminatingGateways.gateways)` instead?

adilyse: I was going to have one gateway defined automatically in the `gateways` field as an example. Then someone could enable gateways and get a gateway without doing any extra work. It should also simplify the explanation of required gateway fields-- though I suppose we'll still have to explain the `extraVolumes` field so maybe that's a wash.

PR #454, opened 2020-05-12
Restrict permissions for the server-acl-init job

ishustava:
Now that the server-acl-init command takes server DNS names as input, we don't need permissions to list pods and get statefulsets.

Additionally, only set one of `connectInject.enabled` to true in the generated config for HCS since we don't recommend enabling both `connectInject` and `syncCatalog`.


lkysow (approved):
Code looks good. I didn't test it.

PR #456, opened 2020-05-13
Support ingress gateways

adilyse:
This creates all of the necessary templates to enable ingress gateways
in the Helm chart. By making use of `ingressGateways.defaults` and
`ingressGateways.gateways` setting, it's possible to enable multiple
ingress gateways. Names must be provided for each and they must be
unique.

All fields defined in the `ingressGateways.defaults` may be overridden
by defining that field on the relevant `ingressGateways.gateways`
entry, with the exception of annotations. Annotations will apply
both the default annotations and any gateway specific annotations.

In an effort to reduce the intrusiveness of the installation, this
feature uses a role and a rolebinding rather than clusterrole/binding.

Added support for creating ingress gateway tokens to the server acl
init job. Added additional (unrendered) spaces in this job's
command to make it easier to reason about.

Note: the current default affinity for ingress gateways won't allow
more than one ingress gateway to be scheduled on the same node, even if
they are logically different gateways. Due to limitations in templating,
there's no good way to change this automatically. To work around it,
an `ingress-gateway-name` label has been added to the deployment.
With it, users will be able to define an affinity per gateway if
they would like to allow multiple, but different, gateways per node.

### DIFF
--- a/templates/ingress-gateways-deployment.yaml
+++ b/templates/ingress-gateways-deployment.yaml
@@ -1,0 +1,376 @@
+{{- if .Values.ingressGateways.enabled }}
+{{- if not .Values.connectInject.enabled }}{{ fail "connectInject.enabled must be true" }}{{ end -}}
+{{- if not .Values.client.grpc }}{{ fail "client.grpc must be true" }}{{ end -}}
+{{- if not (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}{{ fail "clients must be enabled" }}{{ end -}}
+
+{{- $root := . }}
+{{- $defaults := .Values.ingressGateways.defaults }}
+{{- $names := dict }}
+
+{{- range .Values.ingressGateways.gateways }}
+
+{{- $wanAddress := .wanAddress }}
+{{- $service := .service }}
+
+{{- if empty .name }}
+# Check that the gateway name is provided
+{{ fail "Ingress gateway names cannot be empty"}}
+{{ end -}}
+{{- if hasKey $names .name }}
+#  Check that the gateway name is unique
+{{ fail "Ingress gateway names must be unique"}}
+{{ end -}}
+{{- $_ := set $names .name .name }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ .name }}
+  namespace: {{ $root.Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" $root }}
+    chart: {{ template "consul.chart" $root }}
+    heritage: {{ $root.Release.Service }}
+    release: {{ $root.Release.Name }}
+    component: ingress-gateway
+    ingress-gateway-name: {{ .name }}
+spec:
+  replicas: {{ default $defaults.replicas .replicas }}
+  selector:
+    matchLabels:
+      app: {{ template "consul.name" $root }}
+      chart: {{ template "consul.chart" $root }}
+      release: {{ $root.Release.Name }}
+      component: ingress-gateway
+  template:
+    metadata:
+      labels:
+        app: {{ template "consul.name" $root }}
+        chart: {{ template "consul.chart" $root }}
+        release: {{ $root.Release.Name }}
+        component: ingress-gateway
+      annotations:
+        "consul.hashicorp.com/connect-inject": "false"
+        {{- if $defaults.annotations }}
+        # We allow both default annotations and gateway-specific annotations
+        {{- tpl $defaults.annotations $root | nindent 8 }}
+        {{- end }}
+        {{- if .annotations }}
+        {{- tpl .annotations $root | nindent 8 }}
+        {{- end }}
+    spec:
+      {{- if (or $defaults.affinity .affinity) }}
+      affinity:
+        {{ tpl (default $defaults.affinity .affinity) $root | nindent 8 | trim }}
+      {{- end }}
+      {{- if (or $defaults.tolerations .tolerations) }}
+      tolerations:
+        {{ toYaml (default $defaults.tolerations .tolerations) | nindent 8 | trim }}
+      {{- end }}
+      terminationGracePeriodSeconds: 10
+      serviceAccountName: {{ .name }}
+      volumes:
+        - name: consul-bin
+          emptyDir: {}
+        - name: consul-service
+          emptyDir:
+            medium: "Memory"
+        {{- if $root.Values.global.tls.enabled }}
+        {{- if not (and $root.Values.externalServers.enabled $root.Values.externalServers.useSystemRoots) }}
+        - name: consul-ca-cert
+          secret:
+            {{- if $root.Values.global.tls.caCert.secretName }}
+            secretName: {{ $root.Values.global.tls.caCert.secretName }}
+            {{- else }}
+            secretName: {{ template "consul.fullname" $root }}-ca-cert
+            {{- end }}
+            items:
+            - key: {{ default "tls.crt" $root.Values.global.tls.caCert.secretKey }}
+              path: tls.crt
+        {{- end }}
+        {{- if $root.Values.global.tls.enableAutoEncrypt }}
+        - name: consul-auto-encrypt-ca-cert
+          emptyDir:
+            medium: "Memory"
+        {{- end }}
+        {{- end }}
+      initContainers:
+        # We use the Envoy image as our base image so we use an init container to
+        # copy the Consul binary to a shared directory that can be used when
+        # starting Envoy.
+        - name: copy-consul-bin
+          image: {{ $root.Values.global.image | quote }}
+          command:
+          - cp
+          - /bin/consul
+          - /consul-bin/consul
+          volumeMounts:
+          - name: consul-bin
+            mountPath: /consul-bin
+        {{- if (and $root.Values.global.tls.enabled $root.Values.global.tls.enableAutoEncrypt) }}
+        {{- include "consul.getAutoEncryptClientCA" $root | nindent 8 }}
+        {{- end }}
+        # service-init registers the ingress gateway service.
+        - name: service-init
+          image: {{ $root.Values.global.imageK8S }}
+          env:
+            - name: HOST_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.hostIP
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
+            {{- if $root.Values.global.tls.enabled }}
+            - name: CONSUL_HTTP_ADDR
+              value: https://$(HOST_IP):8501
+            - name: CONSUL_CACERT
+              value: /consul/tls/ca/tls.crt
+            {{- else }}
+            - name: CONSUL_HTTP_ADDR
+              value: http://$(HOST_IP):8500
+            {{- end }}
+          command:
+            - "/bin/sh"
+            - "-ec"
+            - |
+                {{- if (or $root.Values.global.acls.manageSystemACLs $root.Values.global.bootstrapACLs) }}
+                consul-k8s acl-init \
+                  -secret-name="{{ template "consul.fullname" $root }}-{{ .name }}-ingress-gateway-acl-token" \
+                  -k8s-namespace={{ $root.Release.Namespace }} \
+                  -token-sink-file=/consul/service/acl-token
+                {{ end }}
+
+                {{- $source := (default $defaults.wanAddress.source $wanAddress.source) }}
+                {{- $serviceType := (default $defaults.service.type $service.type) }}
+                {{- $serviceEnabled := (ternary $defaults.service.enabled $service.enabled (kindIs "invalid" $service.enabled)) }}
+                {{- if and (eq $source "Service") (not $serviceEnabled) }}{{ fail "if ingressGateways .wanAddress.source=Service then ingressGateways .service.enabled must be set to true in either the defaults or in the gateway definition" }}{{ end }}
+                {{- if or (eq $source "NodeIP") (and (eq $source "Service") (eq $serviceType "NodePort")) }}
+                WAN_ADDR="${HOST_IP}"
+                {{- else if eq $source "NodeName" }}
+                WAN_ADDR="${NODE_NAME}"
+                {{- else if and (eq $source "Service") (or (eq $serviceType "ClusterIP") (eq $serviceType "LoadBalancer")) }}
+                consul-k8s service-address \
+                  -k8s-namespace={{ $root.Release.Namespace }} \
+                  -name={{ .name }} \
+                  -output-file=/tmp/address.txt
+                WAN_ADDR="$(cat /tmp/address.txt)"
+                {{- else if eq $source "Static" }}
+                {{- if eq (default $defaults.wanAddress.static $wanAddress.static) "" }}{{ fail "if ingressGateways .wanAddress.source=Static then ingressGateways .wanAddress.static cannot be empty and must be set in either the defaults or in the gateway definition" }}{{ end }}
+                WAN_ADDR="{{ (default $defaults.wanAddress.static $wanAddress.static) }}"
+                {{- else }}
+                {{- fail "currently set ingressGateway values for wanAddress.source and service.type are not supported" }}
+                {{- end }}
+
+                {{- if eq $source "Service" }}
+                {{- if eq $serviceType "NodePort" }}
+                {{- if not (default $defaults.service.nodePort $service.nodePort) }}{{ fail "if ingressGateways .wanAddress.source=Service and ingressGateways .service.type=NodePort, ingressGateways .service.nodePort must be set in either the defaults or in the gateway definition" }}{{ end }}
+                WAN_PORT="{{ (default $defaults.service.nodePort $service.nodePort) }}"
+                {{- else }}
+                WAN_PORT="{{ (default $defaults.service.port $service.port) }}"
+                {{- end }}
+                {{- else }}
+                WAN_PORT="{{ (default $defaults.wanAddress.port $wanAddress.port) }}"
+                {{- end }}
+
+                cat > /consul/service/service.hcl << EOF
+                service {
+                  kind = "ingress-gateway"
+                  name = "{{ .name }}"
+                  {{- if $root.Values.global.enableConsulNamespaces }}
+                  namespace = "{{ (default $defaults.consulNamespace .consulNamespace) }}"
+                  {{- end }}
+                  port = ${WAN_PORT}
+                  address = "${WAN_ADDR}"
+                  tagged_addresses {
+                    lan {
+                      address = "${POD_IP}"
+                      port = 8443
+                    }
+                    wan {
+                      address = "${WAN_ADDR}"
+                      port = ${WAN_PORT}
+                    }
+                  }
+                  proxy {
+                    config {
+                      envoy_gateway_no_default_bind = true
+                      envoy_gateway_bind_addresses {
+                        all-interfaces {
+                          address = "0.0.0.0"
+                        }
+                      }
+                    }
+                  }
+                  checks = [
+                    {
+                      name = "Ingress Gateway Listening"
+                      interval = "10s"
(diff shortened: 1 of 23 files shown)


ishustava (commented):
This approach looks good to me! 

I left a question about supporting host port. I'm not sure we _really_ need to support it.

The behavior of annotations makes sense, although it did seem a bit confusing that it's behaving differently from other defaults. I tried to think of an alternative, but none of them seem like good solutions to me, so I think your solution is the best.

values.yaml
```diff
@@ -1004,6 +1004,59 @@ meshGateway:
   #     "annotation-key": "annotation-value"
   annotations: null
 
+ingressGateways:
+  enabled: true
```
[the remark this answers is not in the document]
adilyse: Yes, it will default to false! I had it set to true for testing :)

values.yaml
```diff
@@ -1004,6 +1004,59 @@ meshGateway:
...
+
+  # Defaults sets default values for all gateway fields. With the exception
+  # of annotations, defining any of these values in the `gateways` list
+  # will override the default values provided here.
+  defaults:
+    replicas: 2
+    imageEnvoy: "envoyproxy/envoy:v1.13.0"
```
lkysow, line 1015: Can we pull this up to a global imageEnvoy so it's not duplicated across all 3 gateways?

values.yaml
```diff
@@ -1004,6 +1004,59 @@ meshGateway:
...
+
+    wanAddress:
+      source: "Service"
+      port: 443
+      static: ""
+
+    hostPort: null
```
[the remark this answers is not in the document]
lkysow: If folks want to expose their ingress gateways on a node to avoid the 2 extra hops (load balancer -> random node in cluster -> ingress pod vs  -> ingress pod) then they'd need this.

We can remove it and add it later but if it's already coded up in mesh gateways and so doesn't cost us any more to implement it *and* it makes the gateway configs consistent then I think we should keep it.

values.yaml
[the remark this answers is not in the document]
lkysow: ahh, I see that by having this you can set the `gateways` array with a default element so the most common use-case (1 gateway) is enabled by just turning this to `true`. That's pretty slick.

values.yaml
adilyse: Connect also uses an `imageEnvoy` field. Do we want to consolidate all of them? Or leave the option to override that one?

values.yaml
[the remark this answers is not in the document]
ishustava: I think we should just recommend service type NodePort for those cases. I believe it's the same number of hops for service type NodePort vs hostport.

I'm just thinking that if we don't have a compelling use case for host port that wouldn't be also solved by service type NodePort, then I think it makes sense to avoid including a configuration that's explicitly not recommended in kube docs.

values.yaml
lkysow: yeah not 100% sure without looking through all the implications so this might not work but my thought was add `global.imageEnvoy` and then for where we were previously using imageEnvoy keep that but use global as the default and for new locations don't offer a component-specific imageEnvoy.

values.yaml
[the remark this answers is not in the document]
lkysow: > I believe it's the same number of hops for service type NodePort vs hostport.

Can you double check that coming in through the nodeport guarantees you route to the pod on that host?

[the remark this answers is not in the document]
lkysow: I'm pretty sure it just hits the kubeproxy iptables and gets randomly forwarded to one of the pod ips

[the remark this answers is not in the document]
ishustava: Oh, I see what you mean. Yeah, you're right, by default kubeproxy will pick a backend pod at random, which could add an extra hop. Although you can [configure](https://kubernetes.io/docs/concepts/services-networking/service/#proxy-mode-ipvs) the kubeproxy to run in a different mode. This configurability likely is not supported by most cloud providers.

I think then if we wanted to support the use case you're describing we'd need to support host port.

templates/ingress-gateways-role.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - {{ .name }}
+    verbs:
+      - use
+{{- end }}
+{{- if (or $root.Values.global.acls.manageSystemACLs $root.Values.global.bootstrapACLs) }}
+  - apiGroups: [""]
+    resources:
+      - secrets
+    resourceNames:
+      - {{ .name }}-ingress-gateway-acl-token
```
lkysow, line 35:
```suggestion
      - {{ template "consul.fullname" $root }}-{{ .name }}-ingress-gateway-acl-token
```


otherwise I get
```
Error getting Kubernetes secret: secrets "consul-consul-ingress-gateway-ingress-gateway-acl-token" is forbidden: User "system:serviceaccount:default:ingress-gateway" cannot get resource "secrets" in API group "" in the namespace "default"

File: templates/ingress-gateways-deployment.yaml [hunk 5]
```diff
@@ -0,0 +1,376 @@
...
+                  port = ${WAN_PORT}
+                  address = "${WAN_ADDR}"
+                  tagged_addresses {
+                    lan {
+                      address = "${POD_IP}"
+                      port = 8443
+                    }
+                    wan {
+                      address = "${WAN_ADDR}"
+                      port = ${WAN_PORT}
```
lkysow, line 192: todo: do we need to define a port? This is only used for DNS so then we'd only need the IP, however if they're using SRV records then maybe we'd use the port? But as we figured out this port can't actually be used for any ingress traffic, only for the healthcheck.

need to check if this port is returned in an SRV lookup. Otherwise I think we can remove it.

lkysow (commented):
This is awesome work!

Here's what I've tested so far:
- simple routing to static-server [works]
- tls enabled [works]
  - `curl -k --resolve static-server.ingress.consul:443:35.202.96.53 -vvv https://static-server.ingress.consul`
- delete gateway consul agent, it gets re-registered [works]
- set gateway replicas to 1, the other one gets deregistered [works]
- `.ingress` dns [works]
- multiple gateways (with tls autoencrypt) [works]
- tls enabled on cluster [works]
- tls with auto encrypt [works]
- acls enabled (no namespaces) **[does not work]**


I deployed the `static-server` pod and then used these configs:
```hcl
kind = "ingress-gateway"
name = "ingress-gateway"
tls {
  enabled = true
}
listeners = [
  {
    port = 8080
    protocol = "http"
    services = [
      {
        name = "static-server"
      },
    ]
  },
]
```

```
kind = "service-defaults"
name = "static-server"
protocol = "http"
```

I also had to manually edit the kube service to use targetPort 8080. I then created an intention from `ingress-gateway` => `static-server` however the ingress gateway isn't creating a listener on 8080 like it should. I tried to edit the ACLs and give it full permissions but that didn't work. Need to dig into that more.

I haven't gotten into the main helm changes yet, just testing functionality.

values.yaml
```diff
@@ -1004,6 +1006,141 @@ meshGateway:
...
+    # annotations defined for a specific gateway in `ingressGateways.gateways`.
+    # Example:
+    #   annotations: |
+    #     "annotation-key": "annotation-value"
+    annotations: null
+
+    # [Enterprise Only] `consulNamespace` defines the Consul namespace to register
+    # the gateway into.  Requires `global.enableConsulNamespaces` to be true and
+    # Consul Enterprise v1.7+ with a valid Consul Enterprise license.
+    consulNamespace: null
```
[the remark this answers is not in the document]
lkysow: Should add a note here that this namespace must be created by the user

templates/ingress-gateways-deployment.yaml
```diff
@@ -0,0 +1,376 @@
...
+          ports:
+            - name: gateway
+              containerPort: 8443
+              {{- if default $defaults.hostPort .hostPort }}
+              hostPort:  {{ default $defaults.hostPort .hostPort }}
+              {{- end }}
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh", "-ec", "/consul-bin/consul services deregister -id=\"{{ .name }}\""]
```
lkysow, line 320: does this need a namespace flag?

templates/ingress-gateways-deployment.yaml
```diff
@@ -0,0 +1,376 @@
...
+            - name: CONSUL_HTTP_ADDR
+              value: http://$(HOST_IP):8500
+            - name: CONSUL_GRPC_ADDR
+              value: $(HOST_IP):8502
+            {{- end }}
+          command:
+            - /consul-bin/consul
+            - connect
+            - envoy
+            - -gateway=ingress
```
lkysow, line 293: this must have -namespace for consul namespaces

lkysow:
I've finished testing:
- acls enabled
  - namespaces
    - route to same ns [works after suggested changes]
    - gateway in other ns routing to same ns [works after suggested changes]
    - gateway in other ns routing to service in other ns [works after suggested changes]

templates/ingress-gateways-deployment.yaml
```diff
@@ -0,0 +1,376 @@
...
+
+                cat > /consul/service/service.hcl << EOF
+                service {
+                  kind = "ingress-gateway"
+                  name = "{{ .name }}"
+                  {{- if $root.Values.global.enableConsulNamespaces }}
+                  namespace = "{{ (default $defaults.consulNamespace .consulNamespace) }}"
+                  {{- end }}
+                  port = ${WAN_PORT}
+                  address = "${WAN_ADDR}"
```
lkysow, line 184:
```suggestion
                  port = 8443
                  address = "${POD_IP}"
```

These values should match the lan address. This is for the individual service instance so it makes sense to be the pod values instead of the top-level kube service.

lkysow (commented):
Wow great work on all those tests. I've got a couple of small comments. The larger items I think we need to deal with are:
1. Prefixing with the release prefix
1. How will we expose ports
1. Not exposing a service.enabled value

templates/ingress-gateways-deployment.yaml
lkysow: I'm wrong! Ignore this.

templates/ingress-gateways-deployment.yaml
lkysow: Ignore. This is good!

values.yaml
```diff
@@ -1004,6 +1006,141 @@ meshGateway:
...
+    service:
+      # Whether to create a Service or not.
+      enabled: true
+
+      # Type of service, ex. LoadBalancer, ClusterIP.
+      type: LoadBalancer
+
+      # Port that the service will be exposed on.
+      # The targetPort will be set to 8443.
+      port: 443
```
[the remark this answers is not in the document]
adilyse: I created another config entry in the service section for any additional ports to be exposed. There are several places where we need a single port, so it made sense to keep those as-is but still allowing folks to expose additional ports.

adilyse (commented):
Updated based on review comments. I haven't had a chance to test out that exposing port `80` by default works, but there are enough changes that it's worth it to take a look again anyway.

lkysow:
@adilyse what scenarios have been tested and which scenarios should I re-test?

values.yaml
```diff
@@ -1039,6 +1043,146 @@ meshGateway:
...
+
+      # Optionally hardcode the nodePort of the service if using type: NodePort.
+      # If not set and using type: NodePort, Kubernetes will automatically assign
+      # a port.
+      nodePort: null
+
+      # Additional ports to expose on both the service and the ingress gateway
+      # container for any listeners that will be defined in the gateway's
+      # configuration entry.
+      additionalPorts: [80]
```
lkysow, line 1121: What about hostPort and nodePort? The use-cases I'm thinking about:

1. I'm running my ingress gateway on one node and using that node's IP address to route to it through the hostPort instead of using the kubernetes service
1. I want to hard-code all the nodePorts for my ingress gateway service–this one I'm not sure if it will ever be used

Currently there's no way to configure nodePorts and hostPorts.

In the end, we need to configure on the deployment for each listener:
* containerPort
* hostPort

and on the service for each listener:
* port
* targetPort
* nodePort

deploy/containerPort == service/targetPort so we can use a single value for that, but what about hostPort and nodePort?

What about supporting:
```yaml
additionalPorts:
- port: 8443 # sets service/port, service/targetPort, deploy/containerPort
  nodePort: 8443 # sets service/nodePort
  hostPort: 8443 # set deploy/hostPort
```

Or maybe have
```yaml
useHostPorts: true
additionalPorts:
- 8443 # sets service/port, service/targetPort, deploy/containerPort, deploy/hostPort
```

That one won't work for setting nodePorts but not sure if that's a legit use-case?

values.yaml
```diff
@@ -1004,6 +1006,141 @@ meshGateway:
...
+      # Example:
+      #   annotations: |
+      #     "annotation-key": "annotation-value"
+      annotations: null
+
+      # Optional YAML string that will be appended to the Service spec.
+      additionalSpec: null
+
+    resources:
+      requests:
```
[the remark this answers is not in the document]
lkysow: That got merged so these can match the mesh gateway with
```
    resources:
      requests:
        memory: "100Mi"
        cpu: "100m"
      limits:
        memory: "100Mi"
        cpu: "100m"
```

test/unit/ingress-gateways-podsecuritypolicy.bats
```diff
@@ -0,0 +1,54 @@
...
+  [ "${actual}" = "false" ]
+}
+
+@test "ingressGateways/PodSecurityPolicy: enabled with ingressGateways, connectInject and client.grpc enabled and global.enablePodSecurityPolicies=true" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/ingress-gateways-podsecuritypolicy.yaml  \
+      --set 'ingressGateways.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      --set 'client.grpc=true' \
```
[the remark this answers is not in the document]
lkysow: Thanks!

test/unit/server-acl-init-job.bats
```diff
@@ -271,6 +286,103 @@ load _helpers
...
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/server-acl-init-job.yaml  \
+      --set 'global.acls.manageSystemACLs=true' \
+      --set 'ingressGateways.enabled=true' \
+      --set 'connectInject.enabled=true' \
+      --set 'client.grpc=true' \
+      --set 'global.enableConsulNamespaces=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command | any(contains("-ingress-gateway-name=\"ingress-gateway\""))' | tee /dev/stderr)
```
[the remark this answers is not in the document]
lkysow: Nice, that's a good UX.

templates/ingress-gateways-deployment.yaml
```diff
@@ -0,0 +1,376 @@
...
+
+{{- if empty .name }}
+# Check that the gateway name is provided
+{{ fail "Ingress gateway names cannot be empty"}}
+{{ end -}}
+{{- if hasKey $names .name }}
+#  Check that the gateway name is unique
+{{ fail "Ingress gateway names must be unique"}}
+{{ end -}}
+{{- $_ := set $names .name .name }}
```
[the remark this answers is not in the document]
lkysow: 🙏

templates/ingress-gateways-deployment.yaml
```diff
@@ -0,0 +1,376 @@
...
+metadata:
+  name: {{ .name }}
+  namespace: {{ $root.Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" $root }}
+    chart: {{ template "consul.chart" $root }}
+    heritage: {{ $root.Release.Service }}
+    release: {{ $root.Release.Name }}
+    component: ingress-gateway
+    ingress-gateway-name: {{ .name }}
```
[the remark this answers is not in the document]
lkysow: > Related to affinity, it's also necessary for the service selectors to be able to reference the correct gateways without overlapping between distinct gateways.

Ooh nice catch!!

templates/ingress-gateways-deployment.yaml
```diff
@@ -0,0 +1,376 @@
...
+        {{- tpl .annotations $root | nindent 8 }}
+        {{- end }}
+    spec:
+      {{- if (or $defaults.affinity .affinity) }}
+      affinity:
+        {{ tpl (default $defaults.affinity .affinity) $root | nindent 8 | trim }}
+      {{- end }}
+      {{- if (or $defaults.tolerations .tolerations) }}
+      tolerations:
+        {{ toYaml (default $defaults.tolerations .tolerations) | nindent 8 | trim }}
```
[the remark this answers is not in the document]
lkysow: Yes I think it makes sense to move to that for tolerations but we should probably do it all at once so it's consistent.

templates/ingress-gateways-deployment.yaml
lkysow: Have you tested this?

lkysow: In my testing the service is not deregistered. This command doesn't refer to the service registration file, it just calls consul directly saying "deregister service with id <blah>". This call defaults to the default namespace so if the gateway is in another namespace then it won't get deregistered. It also fails silently because you can put any id in this call and it will always succeed since if the service doesn't exist it assumes it's already been deregistered.

values.yaml
```diff
@@ -1061,64 +1061,23 @@ ingressGateways:
...
-    # to expose the gateways directly from the node.
-    # NOTE: Cannot set to 8500 or 8502 because those are reserved for the Consul
-    # agent.
-    hostPort: null
-
     # The service options configure the Service that fronts the gateway Deployment.
     service:
-      # Type of service, ex. LoadBalancer, ClusterIP.
+      # Type of service: LoadBalancer, ClusterIP or NodePort. If using NodePort service
+      # type, you must set the desired nodePorts in the `port` setting below.
```
lkysow, line 1067:
```suggestion
      # type, you must set the desired nodePorts in the `ports` setting below.
```

values.yaml
```diff
@@ -1061,64 +1061,23 @@ ingressGateways:
...
-      additionalPorts: [80]
+      # Ports that will be exposed on the service and gateway container. Any
+      # ports defined as ingress listeners on the gateway's Consul configuration
+      # entry should be included here. The first port will be used as part of
+      # the Consul service registration for the gateway and be listed in its
+      # SRV record. If using a NodePort service type, you must specify the
+      # desired nodePort for each exposed port.
+      ports:
+        - port: 80
+          nodePort: null
```
lkysow, line 1078: If I just set
```
port: 80
```

does it work? without `nodePort:null`.

I think it would look better without the nodePort:null and have an example of nodeports in the doc block but not a big deal.

lkysow (commented):
The changes for ports and removing wanAddress look good!

templates/ingress-gateways-deployment.yaml
adilyse: Updated to include the namespace flag.

values.yaml

values.yaml
adilyse: Yes, it would work without explicitly showing the `nodePort: null`, though I prefer to have the entire set of fields visible in the `defaults` section to make it completely clear what settings are available. I think that's especially important here since we're creating an arbitrary port object.

values.yaml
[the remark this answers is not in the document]
adilyse: Updated.

templates/ingress-gateways-deployment.yaml
```diff
@@ -0,0 +1,376 @@
...
+                  }
+                  checks = [
+                    {
+                      name = "Ingress Gateway Listening"
+                      interval = "10s"
+                      tcp = "${POD_IP}:8443"
+                      deregister_critical_service_after = "6h"
+                    }
+                  ]
+                }
```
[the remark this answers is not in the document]
lkysow: this is now in the name of the service itself so likely not necessary.

lkysow (approved):
🎉 

Great work on the refactored deployment tests, they look much easier to change for future contributors.

I've tested :allthethings: I could think of so this is good to go!

PR #462, opened 2020-05-19
Update README.md

lkysow:
Remove the part about installing Helm into Kubernetes because with helm 3 they don't need to do this and presumably if they don't already have helm they're going to use helm3 and if they do already have helm 2 then it's already installed on their cluster so they don't need instructions to do so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,14 +9,7 @@ use Consul with Kubernetes, please see the
 [Consul and Kubernetes documentation](https://www.consul.io/docs/platform/k8s/index.html).
 
 ## Prerequisites
-
-To use the charts here, [Helm](https://helm.sh/) must be installed in your
-Kubernetes cluster. Setting up Kubernetes and Helm and is outside the scope
-of this README. Please refer to the Kubernetes and Helm documentation.
-
-The versions required are:
-
-  * **Helm 2.10+** - This is the earliest version of Helm tested. It is possible
+  * **Helm 2.10+ or Helm 3.0+** - This is the earliest version of Helm tested. It is possible
     it works with earlier versions but this chart is untested for those versions.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,14 +9,7 @@ use Consul with Kubernetes, please see the
...
-
-To use the charts here, [Helm](https://helm.sh/) must be installed in your
-Kubernetes cluster. Setting up Kubernetes and Helm and is outside the scope
-of this README. Please refer to the Kubernetes and Helm documentation.
-
-The versions required are:
-
-  * **Helm 2.10+** - This is the earliest version of Helm tested. It is possible
+  * **Helm 2.10+ or Helm 3.0+** - This is the earliest version of Helm tested. It is possible
     it works with earlier versions but this chart is untested for those versions.
```
ishustava, line 13: Do we need to add tests for both versions of helm if we're saying that?

lkysow: I've been using Helm 3 extensively so I don't think so. We should switch to Helm 3 for our template tests but I don't think that needs to block this.

ishustava: Yeah me too, but I think this refers to automatic testing. I can't guarantee that I will test every PR or every change myself. I think if we're not testing both, we shouldn't say that we test it. Maybe we could just rephrase this to not say which versions we're testing with and say that it *works* with these versions of Helm.

lkysow: Ahh gotcha, yeah will reword.

lkysow: I just changed it to "Helm 2.10+ or Helm 3.0+". What do you think?

ishustava: Nice!

ishustava (approved):

PR #463, opened 2020-05-20
Helm 3 Tests

### LINKED ISSUE #290: Tests fail with helm 3
Our bats tests use `helm template -x` to render only the template under test, however that flag is removed in helm 3 and it doesn't look like there are immediate plans to bring it back (https://github.com/helm/helm/issues/6290)

lkysow:
Fixes #290 

Requires changing how we test empty templates due to https://github.com/helm/helm/issues/7295 (`-x` is replaced with `-s` which now errors out if the rendered template is empty). Run tests in both helm 2 and 3.


templates/client-config-configmap.yaml
```diff
@@ -1,6 +1,6 @@
+{{- if (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}
```
ishustava, line 1: Is this change intentional?

test/unit/_helpers.bash
```diff
@@ -2,3 +2,39 @@
...
+helm() {
+  if [[ $(v2) ]]; then
+    command helm template -x "${@:3}"
+  else
+    command helm "$@"
+  fi
+}
+
+# Usage: assert_empty helm template -s <template> [flags] .
+# assert_empty makes is possible to test that a template is not rendered in
```
ishustava, line 19:
```suggestion
# assert_empty makes it possible to test that a template is not rendered in
```

ishustava (dismissed):
Looks good to me! Left a couple of minor comments.

templates/client-config-configmap.yaml
lkysow: Yes, we need to move all comments inside the `if` statement. Otherwise the template gets rendered (with just the comments) and we can't do the same helm 3 error assertion. Since some templates have comments and some don't, it's better to standardize on all templates having comments inside the enclosing `if` statements so they always result in the same helm 3 error.

ishustava: Cool, makes sense. Thanks for clarifying!

test/unit/_helpers.bash
```diff
@@ -2,3 +2,41 @@
 chart_dir() {
     echo ${BATS_TEST_DIRNAME}/../..
 }
+
+# helm is used to intercept the helm command in tests and change flags depending
```
ishustava, line 6:
```suggestion
# helm is used to intercept the Helm command in tests and change flags depending
```

test/unit/client-daemonset.bats
```diff
@@ -401,28 +397,26 @@ load _helpers
...
   cd `chart_dir`
   local actual=$(helm template \
-      -x templates/client-daemonset.yaml  \
+      -s templates/client-daemonset.yaml  \
       . | tee /dev/stderr |
       yq '.spec.template.spec.containers[] | select(.name=="consul") | .env[] | select(.name == "GOSSIP_KEY") | length > 0' | tee /dev/stderr)
   [ "${actual}" = "" ]
 }
 
 @test "client/DaemonSet: gossip encryption disabled in client DaemonSet when clients are disabled" {
```
ishustava, line 406: This test is confusing to me. It seems a bit redundant given that every feature is disabled when clients are disabled, and I don't think we test this for every feature. I'd delete it, but up to you!

test/unit/client-daemonset.bats
```diff
@@ -14,7 +14,7 @@ load _helpers
 @test "client/DaemonSet: enable with global.enabled false" {
```
ishustava, line 14:
```suggestion
@test "client/DaemonSet: enabled with global.enabled=false and client.enabled=true" {
```

test/unit/client-daemonset.bats
```diff
@@ -24,28 +24,24 @@ load _helpers
 
 @test "client/DaemonSet: disable with client.enabled" {
```
ishustava, line 25:
```suggestion
@test "client/DaemonSet: disabled with client.enabled=false" {
```

test/unit/client-daemonset.bats
```diff
@@ -24,28 +24,24 @@ load _helpers
...
+  assert_empty helm template \
+      -s templates/client-daemonset.yaml  \
       --set 'client.enabled=false' \
-      . | tee /dev/stderr |
-      yq 'length > 0' | tee /dev/stderr)
-  [ "${actual}" = "false" ]
+      .
 }
 
 @test "client/DaemonSet: disable with global.enabled" {
```
ishustava, line 33:
```suggestion
@test "client/DaemonSet: disabled with global.enabled=false and client.enabled='-'" {
```

test/unit/client-snapshot-agent-serviceaccount.bats
```diff
@@ -4,17 +4,13 @@ load _helpers
 
 @test "client/SnapshotAgentServiceAccount: disabled by default" {
   cd `chart_dir`
-  local actual=$(helm template \
-      -x templates/client-snapshot-agent-serviceaccount.yaml  \
-      . | tee /dev/stderr |
-      yq 'length > 0' | tee /dev/stderr)
-  [ "${actual}" = "false" ]
+  assert_empty helm template -s templates/client-snapshot-agent-serviceaccount.yaml .
```
ishustava, line 7: wow! kind of cool that it became a one-liner 😄

test/unit/server-acl-init-job.bats
```diff
@@ -185,20 +179,21 @@ load _helpers
...
 # aclBindingRuleSelector/global.acls.manageSystemACLs
 
-@test "serverACLInit/Job: no acl-binding-rule-selector flag by default" {
+@test "serverACLInit/Job: acl-binding-rule-selector flag set by default" {
   cd `chart_dir`
   local actual=$(helm template \
-      -x templates/server-acl-init-job.yaml \
+      -s templates/server-acl-init-job.yaml \
+      --set 'global.acls.manageSystemACLs=true' \
       --set 'connectInject.aclBindingRuleSlector=foo' \
```
ishustava, line 187: Hmm, this test implies that we shouldn't be setting aclBindingRuleSelector. It's strange that it's passing 🤔

ishustava: Oh there's a typo in `aclBindingRuleSlector`, that's why it works!

ishustava (approved):
Looks great 💃 💃 💃 I left a couple of minor comments, but won't block the approval.

test/unit/_helpers.bash
lkysow: I'm gonna use
```
`helm`
```

here to be clear that we're intercepting the literal helm string.

test/unit/client-daemonset.bats
lkysow: agreed!

test/unit/server-acl-init-job.bats
lkysow: nice catch, I deleted the `--set 'connectInject.aclBindingRuleSlector=foo' \`

test/unit/client-snapshot-agent-serviceaccount.bats
lkysow: haha yes!

PR #465, opened 2020-05-21
Deprecate setting resources as string.

lkysow:
Deprecate setting:
```yaml
client:
    resources: |
      requests:
        memory: "128Mi"
        cpu: "250m"
      limits:
        memory: "256Mi"
        cpu: "500m"
```

for
```yaml
client:
    resources:
      requests:
        memory: "128Mi"
        cpu: "250m"
      limits:
        memory: "256Mi"
        cpu: "500m"
```

Applies to server, client and mesh gateway resources. Lays groundwork for upcoming resource settings.

Related to the work done on the vault chart in https://github.com/hashicorp/vault-helm/pull/272

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -252,9 +252,10 @@ server:
   connect: true
 
   # Resource requests, limits, etc. for the server cluster placement. This
-  # should map directly to the value of the resources field for a PodSpec,
-  # formatted as a multi-line string. By default no direct resource request
-  # is made.
+  # should map directly to the value of the resources field for a PodSpec.
+  # By default no direct resource request is made.
+  # NOTE: The use of a YAML string is deprecated. Instead, set directly as a
+  #       YAML map.
   resources: null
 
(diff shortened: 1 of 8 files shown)


values.yaml
```diff
@@ -252,9 +252,8 @@ server:
   connect: true
 
   # Resource requests, limits, etc. for the server cluster placement. This
-  # should map directly to the value of the resources field for a PodSpec,
-  # formatted as a multi-line string. By default no direct resource request
-  # is made.
+  # should map directly to the value of the resources field for a PodSpec.
+  # By default no direct resource request is made.
```
ishustava, line 256: Could we add a deprecation notice here and other values as well?

ishustava (approved):
Looks good 👍

values.yaml
lkysow: Will do.

PR #466, opened 2020-05-21
Set default resources

### LINKED ISSUE #393: Consul clients and servers should have sensible resources defaults set
Currently, we don't have defaults for resource limits for the Consul client daemonset or the server statefulset. Not setting resource limits could cause agents to not leave the cluster gracefully if Kube evicts them due to resource constraints (see #387).

We should have sensible defaults for resource limits instead to avoid this.

lkysow:
This enables our chart to be deployed into clusters that require
resources to be set for each container. This also prevents our pods
from being evicted first when Kubernetes runs out of resources because
Kubernetes will prefer to evict pods that have no resource settings.
To that end we have set defaults for clients and servers because even in
first-run tests we observed those pods being evicted, causing confusing
issues.

We are setting our requests and limits to the same value so we are
considered to have a 'Guaranteed' quality of service class.

These default settings were chosen based on observation in a running
cluster using `kubectl top` and also looking at GKE metrics. We ran a
Consul connect workload at 1 req/s. For the helper components, the load
we saw is representative of a production cluster because the workload of
these components doesn't change, e.g. the lifecycle sidecar. For the
core components like clients, servers and mesh gateways, their resource
requirements will increase depending on the workload. We have set their
resources to more than double what was observed in a base-case installation so
that first-run doesn't require an unnecessary amount of resources but
also won't run up against any limits. It's expected that for production,
users will tune their settings based on their specific use-case and
observed load.

In our observations, most of the "helper" containers
used a maximum of 10 millicores and 12 Mi of memory. We have set these
smaller containers and init containers to 50 millicores and 25 Mi of memory to be safe.
For the lifecycle sidecar we observed 2m/12Mi. We set this to 10m/25Mi
to keep this footprint small. This setting will also be used for connect-injected
apps so we can't increase it too much or it will cause apps to have
large resource footprints unnecessarily.
For components like sync catalog, connect inject and snapshot agent we
saw max usage of 44m/14Mi. Since these are long-running components, we
set their resources to 50m/50Mi.
For Consul clients we saw maximum usage of 10m/15Mi, we have set these
to 100m/100Mi. For Consul servers we same maximum usage of 40m/13Mi. We
have set these to 100m/100Mi. For mesh gateways with multi-dc federation
we observed maximum usage of 15m/13Mi. We have set these to 100m/100Mi.

Fixes #393

## Testing
Create a resource quota:
```yaml
apiVersion: v1
kind: ResourceQuota
metadata:
  name: compute-resources
spec:
  hard:
    requests.cpu: "1"
    requests.memory: 1Gi
    limits.cpu: "2"
    limits.memory: 2Gi
```

Try to install the old Helm chart (it will fail). Now install this version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,60 @@
 ## Unreleased
 
+FEATURES:
+
+* Resources are now set on all containers. This enables the chart to be deployed
+  in clusters that have resource quotas set. This also ensures that Consul
+  server and client pods won't be evicted by Kubernetes when nodes reach their
+  resource limits.
+  
+  Resource settings have been made configurable for sync catalog, connect inject
+  and client snapshot deployments.
+  
+  The default settings were chosen based on a cluster with a small workload.
+  For production, we recommend monitoring resource usage and modifying the
+  defaults according to your usage.
+
 BREAKING CHANGES:
 
 * Mesh Gateway: `meshGateway.enableHealthChecks` is no longer supported. This config
   option was to work around an issue where mesh gateways would not listen on their
   bind ports until a Connect service was registered. This issue was fixed in Consul 1.6.2. ([GH-464](https://github.com/hashicorp/consul-helm/pull/464))
+
+* Mesh Gateway: The default resource settings have been changed. To keep
+  the previous settings, you must set `meshGateway.resources` in your own Helm config. ([GH-466](https://github.com/hashicorp/consul-helm/pull/466))
+
+  Before:
+  ```yaml
+  meshGateway:
+    resources:
+      requests:
+        memory: "128Mi"
+        cpu: "250m"
+      limits:
+        memory: "256Mi"
+        cpu: "500m"
+  ```
+
+  After:
+  ```yaml
+  meshGateway:
+    resources:
+      requests:
+        memory: "100Mi"
+        cpu: "100m"
+      limits:
+        memory: "100Mi"
+        cpu: "100m"
+  ```
+
+ * Clients and Servers: There are now default resource settings for Consul clients
+   and servers. Previously, there were no default settings which meant the default
+   was unlimited. This change was made because Kubernetes will prefer to evict
+   pods that don't have resource settings and that resulted in the Consul client
+   and servers being evicted. The default resource settings were chosen based
+   on a low-usage cluster. If you are running a production cluster, use the
+   `kubectl top` command to see how much CPU and memory your clients and servers
+   are using and set the resources accordingly. ([GH-466](https://github.com/hashicorp/consul-helm/pull/466)]
 
 DEPRECATIONS
 
--- a/values.yaml
+++ b/values.yaml
@@ -251,12 +251,16 @@ server:
   # via the extraConfig setting.
   connect: true
 
-  # Resource requests, limits, etc. for the server cluster placement. This
-  # should map directly to the value of the resources field for a PodSpec.
-  # By default no direct resource request is made.
+  # Resource settings for Server agents.
   # NOTE: The use of a YAML string is deprecated. Instead, set directly as a
-  #       YAML map.
-  resources: null
+  # YAML map.
+  resources:
+    requests:
+      memory: "100Mi"
+      cpu: "100m"
+    limits:
+      memory: "100Mi"
+      cpu: "100m"
 
   # updatePartition is used to control a careful rolling update of Consul
   # servers. This should be done particularly when changing the version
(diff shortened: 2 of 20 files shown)


lkysow:
Still need to test this further.

ishustava (commented):
This generally looks good to me, but I have a couple of questions:
1. You've mentioned that you ran a workload at 1req/s. Was it one connect app talking to another at 1req/s throughput?
1. I thought we will make sync catalog, connect injector, and snapshot agent deployment resource settings configurable. Has that changed?

lkysow:
> You've mentioned that you ran a workload at 1req/s. Was it one connect app talking to another at 1req/s throughput?

Yes.

> I thought we will make sync catalog, connect injector, and snapshot agent deployment resource settings configurable. Has that changed?

🤦 Yes you're totally right I just missed this. Will fix.

values.yaml
```diff
@@ -411,12 +413,14 @@ client:
   # also changes the clients' advertised IP to the hostIP rather than podIP.
   exposeGossipPorts: false
 
-  # Resource requests, limits, etc. for the client cluster placement. This
-  # should map directly to the value of the resources field for a PodSpec.
-  # By default no direct resource request is made.
-  # NOTE: The use of a YAML string is deprecated. Instead, set directly as a
-  #       YAML map.
```
ishustava, line 418: Should we keep the comment about YAML string deprecation?

values.yaml
```diff
@@ -966,15 +997,13 @@ meshGateway:
   hostPort: null
 
   # Resource settings for mesh gateway pods.
-  # NOTE: The use of a YAML string is deprecated. Instead, set directly as a
-  #       YAML map.
```
ishustava, line 970: Hmm, similar to above, why are we removing the comment? We still support both.

values.yaml
```diff
@@ -966,15 +997,13 @@ meshGateway:
...
   resources:
     requests:
-      memory: "128Mi"
-      cpu: "250m"
+      memory: "100Mi"
+      cpu: "100m"
```
ishustava, line 1003: Since we're decreasing default resources, should this be called out as a breaking change too?

ishustava (commented):
All the code changes look great 👍 🎉 I left some questions around values docs and changelog entries. Once those are cleared, I think it's good to go.

values.yaml
lkysow: yes, rebase mistake

values.yaml
lkysow: yes, rebase mistake

values.yaml
lkysow: good call! will add.

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 ## Unreleased
 
+FEATURES:
```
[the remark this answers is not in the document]
lkysow: Added for mesh gateways.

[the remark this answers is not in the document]
lkysow: Ack yes totally.

CHANGELOG.md
```diff
@@ -1,11 +1,61 @@
...
+      requests:
+        memory: "100Mi"
+        cpu: "100m"
+      limits:
+        memory: "100Mi"
+        cpu: "100m"
+  ```
+
+ * Clients and Servers: There are now default resource settings for Consul clients
+   and servers. Previously there were no default settings which meant the default
```
ishustava, line 51: 🤓 
```suggestion
   and servers. Previously, there were no default settings which meant the default
```

CHANGELOG.md
```diff
@@ -1,11 +1,61 @@
...
+   was unlimited. This change was made because Kubernetes will prefer to evict
+   pods that don't have resource settings and that resulted in the Consul client
+   and servers being evicted. The default resource settings were chosen based
+   on a low-usage cluster. If you are running a production cluster, use the
+   kubectl top command to see how much cpu and memory your clients and servers
+   are using and set the resources accordingly. ([GH-466](https://github.com/hashicorp/consul-helm/pull/466))
```
ishustava, line 57:
```suggestion
   are using and set the resources accordingly. ([GH-466](https://github.com/hashicorp/consul-helm/pull/466)]
```

CHANGELOG.md
```diff
@@ -1,11 +1,61 @@
...
+        cpu: "100m"
+  ```
+
+ * Clients and Servers: There are now default resource settings for Consul clients
+   and servers. Previously there were no default settings which meant the default
+   was unlimited. This change was made because Kubernetes will prefer to evict
+   pods that don't have resource settings and that resulted in the Consul client
+   and servers being evicted. The default resource settings were chosen based
+   on a low-usage cluster. If you are running a production cluster, use the
+   kubectl top command to see how much cpu and memory your clients and servers
```
ishustava, line 56:
```suggestion
   `kubectl top` command to see how much CPU and memory your clients and servers
```

ishustava (approved):
🎉 Left a couple of non-blocking editing changes, but otherwise looks great to me!

CHANGELOG.md
lkysow: 🍺

PR #469, opened 2020-05-26
Fully remove support for global.bootstrapACLs

lkysow:
`bootstrapACLs` was deprecated in v0.19.0. We should merge this for release v0.22.0.
Anyone still using `bootstrapACLs` will get an error so they won't accidentally turn off ACLs.

To Do:
- [ ] Add changelog entry once we know next release will be 0.22.0. Adding it now might cause a merge conflict later.

### DIFF
--- a/templates/NOTES.txt
+++ b/templates/NOTES.txt
@@ -20,7 +20,7 @@ To learn more about the release if you are using Helm 3, run:
   $ helm get all {{ .Release.Name }}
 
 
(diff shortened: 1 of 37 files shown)


ishustava (commented):
Looks good! I've added some suggestions to replace the "deprecated" word with "removed" because deprecation means we still support it. Once that's done, it's good to go!

ishustava (approved):
🎉

PR #470, opened 2020-05-26
Support default sidecar proxy resources

lkysow:
Companion to https://github.com/hashicorp/consul-k8s/pull/267

Allow setting default sidecar proxy resources. The defaults are set to `null` so that by default the envoy proxy doesn't run up against any limits.

To test, use
```yaml
global:
  imageK8S: lkysow/consul-k8s-dev:may26-resources3
connectInject:
  enabled: true
  sidecarProxy:
    resources:
      limits:
        memory: 200Mi
        cpu: 200m
      requests:
        memory: 100Mi
        cpu: 100m

```

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -893,6 +893,26 @@ connectInject:
     proxyDefaults: |
       {}
 
+  sidecarProxy:
+    # Set default resources for sidecar proxy. If null, that resource won't
+    # be set.
+    # These settings can be overridden on a per-pod basis via these annotations:
+    # - consul.hashicorp.com/sidecar-proxy-cpu-limit
+    # - consul.hashicorp.com/sidecar-proxy-cpu-request
+    # - consul.hashicorp.com/sidecar-proxy-memory-limit
+    # - consul.hashicorp.com/sidecar-proxy-memory-request
+    resources:
+      requests:
+        # Recommended default: 100Mi
+        memory: null
+        # Recommended default: 100m
+        cpu: null
+      limits:
+        # Recommended default: 100Mi
+        memory: null
+        # Recommended default: 100m
+        cpu: null
+
 # Mesh Gateways enable Consul Connect to work across Consul datacenters.
 meshGateway:
   # If mesh gateways are enabled, a Deployment will be created that runs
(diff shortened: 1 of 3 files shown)


values.yaml
```diff
@@ -862,6 +862,21 @@ connectInject:
     proxyDefaults: |
       {}
 
+  sidecarProxy:
+    # Set default resources for sidecar proxy. If null, that resource won't
+    # be set.
```
ishustava, line 867: Should this be on the line above? Also, should we mention here that they could be overwritten via an annotation?

ishustava (commented):
This looks good! I've tested it out, and everything looked good to me. Going back to my comment in the hashicorp/consul-k8s#267 about checking if limit >= resource, I don't think we need to do it for annotations. I was able to see the error message in the status of my deployment. But I still think it'd be useful to check that for defaults in the connect inject command.

I'm curious though about this choice of defaults. Are you concerned that making our recommended defaults actual defaults will be a breaking change? Would it make sense to only set requests but not limits to eliminate the concern around running up against limits?

values.yaml
lkysow: I put the docs here because sidecarProxy is a top-level setting that we might put other settings into. Yes, will mention the ability to override.

lkysow:
> But I still think it'd be useful to check that for defaults in the connect inject command.
Will do

> I'm curious though about this choice of defaults. Are you concerned that making our recommended defaults actual defaults will be a breaking change? Would it make sense to only set requests but not limits to eliminate the concern around running up against limits?

I made this decision (not setting defaults) to make the first-run experience better. Needing 100Mi/100m per pod just for the sidecar could add up quickly on a small cluster where someone is testing things out. Since all of our defaults are catered towards the first-run experience and we expect production users to customize settings themselves this decision felt inline with the rest of the chart defaults.

ishustava (approved):
This looks good! Keeping defaults as null makes sense for now.

lkysow:
NOTE: Waiting for consul-k8s build.

lkysow:
Require https://github.com/hashicorp/consul-helm/pull/499

Issue #358, opened 2020-02-21
Support external-to-k8s Consul clients


larssb:
**THE ISSUE**

Connecting `Consul` clients external to a K8S cluster, when `Pod` IP's are not route-able. 

**SETUP**

* K8S v1.14.6
* Docker v18.3.1
* Using the Official Consul Helm chart found [here](https://hub.helm.sh/charts/stable/consul). v0.16.2
* Consul agent running on a Windows node outside the cluster
  * Agent commandline: `.\consul.exe agent -retry-join 'provider=k8s kubeconfig="./consul-conf/kubeconfig" label_selector=""app=consul,component=server""' -enable-local-script-checks -data-dir="./consul-data" -config-dir="./consul-conf" -log-level=DEBUG`

**ERROR/ISSUE**

```
2020-02-21T10:01:30.622+0100 [ERROR] agent.anti_entropy: failed to sync remote state: error="No known Consul servers"
   2020-02-21T10:01:34.896+0100 [DEBUG] agent: Check status updated: check=service:scanandpay:1 status=passing
   2020-02-21T10:01:36.541+0100 [DEBUG] agent.client.manager: No healthy servers during rebalance, aborting
   2020-02-21T10:01:36.940+0100 [DEBUG] agent: discover: Using provider "k8s": cluster=LAN
   2020-02-21T10:01:37.090+0100 [INFO]  agent: Discovered servers: cluster=LAN cluster=LAN servers="10.42.22.77 10.42.21.65 10.42.20.82"
   2020-02-21T10:01:37.091+0100 [INFO]  agent: (LAN) joining: lan_addresses=[10.42.22.77, 10.42.21.65, 10.42.20.82]
   2020-02-21T10:01:43.328+0100 [WARN]  agent: Check is now critical: check=service:scanandpay:2
   2020-02-21T10:01:45.197+0100 [DEBUG] agent: Check status updated: check=service:scanandpay:1 status=passing
   2020-02-21T10:01:47.093+0100 [DEBUG] agent.client.memberlist.lan: memberlist: Failed to join 10.42.22.77: dial tcp 10.42.22.77:8301: i/o timeout
   2020-02-21T10:01:47.266+0100 [WARN]  agent.client.manager: No servers available
   2020-02-21T10:01:47.267+0100 [ERROR] agent.anti_entropy: failed to sync remote state: error="No known Consul servers"
[...]

PR #475, opened 2020-06-01
Support consul external to k8s cluster clients

zakkg3:
If pod netwok is not routeable form external nodes/ agents (which is the most common case) we need to run the servers in the node network

Fixes:
https://github.com/hashicorp/consul-helm/issues/358

hostPort definitions are not needed as we run in hostNetwork.

```
      dnsPolicy: ClusterFirstWithHostNet
      hostNetwork: true
```

* inside the k8s cluster consul is reachable via k8s dns:
  euler-server.consul.svc (from inside a pod)
  euler-server.consul.svc.cluster.local (from a k8s node)

* on the external nodes, the consul servers are reachable via dns at consul.service.consul for 
  nodes using the consul dns server or, and recommended for big cluster, any other caching dns that stub-zone or forward-zone (i.e. unbound with consul stub-zone).

Signed-off-by: Nico <zakkg3@gmail.com>
in collaboration with @asteven

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -52,6 +52,10 @@ spec:
       tolerations:
         {{ tpl .Values.server.tolerations . | nindent 8 | trim }}
     {{- end }}
+      {{- if .Values.global.hostNetwork }}
+      dnsPolicy: ClusterFirstWithHostNet
+      hostNetwork: true
+      {{- end }}
       terminationGracePeriodSeconds: 30
       serviceAccountName: {{ template "consul.fullname" . }}-server
       securityContext:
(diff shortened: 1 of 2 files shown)


lseelenbinder:
I'd like to use this, if possible to deploy in a cluster setup where the Pod IPs are not visible to the LAN private network in the same provider's local DC.

Is there any timeline for merging this at this point?

mleklund:
I am in the same boat.  I am considering a cluster of non-k8s consul in the same DC and treating them as separate clusters, but I would rather not.

lseelenbinder:
> I am in the same boat. I am considering a cluster of non-k8s consul in the same DC and treating them as separate clusters, but I would rather not.

I briefly considered the same, and then decided a cluster outside of k8s is better (and clients inside of k8s, since I think that's supported). However, I'd like to avoid deploying Consul without Helm, myself.

david-yu:
We now have hostNetwork support as described here as part of Consul Helm: https://www.consul.io/docs/k8s/helm#v-client-hostnetwork

moreinhardt:
@david-yu 

> We now have hostNetwork support as described here as part of Consul Helm for Consul Servers: https://www.consul.io/docs/k8s/helm#v-client-hostnetwork

I just checked that link and the PR that introduced that functionality - If I understand correctly this is only for _clients_, right?

So it does something different than this PR and doesn't make the server available to clients outside of the k8s cluster...

t-eckert:

PR #481, opened 2020-06-05
Support custom CA cert for snapshot agent

### LINKED ISSUE #467: Support custom CA certs for consul snapshot agent
> When using a S3-compatible storage exposing a self-signed certificate the agent will not be able to perform the snapshot operations unless the CA used to sign the storage certificate is trusted by the node running the agent. You can add the CA root certificate to the OS trust store to have Consul trust the storage endpoint.

https://www.consul.io/docs/commands/snapshot/agent.html

We need to support a way to set a custom CA cert if users are using an S3-compatible storage for their snapshots that's using a custom cert.

lkysow:
For users running an S3 compatible storage for the snapshot agent with
their own CA cert they need a way to include that in the system certs so
that the snapshot agent can connect with their URL.

Fixes https://github.com/hashicorp/consul-helm/issues/467

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -529,7 +529,7 @@ client:
         memory: "50Mi"
         cpu: "50m"
 
-    # Optional CA certificate that will be added to the trusted system CAs.
+    # Optional PEM-encoded CA certificate that will be added to the trusted system CAs.
     # Useful if using an S3-compatible storage exposing a self-signed certificate.
     # Example
     #   caCert: |
--- a/values.yaml
+++ b/values.yaml
@@ -529,6 +529,15 @@ client:
         memory: "50Mi"
         cpu: "50m"
 
+    # Optional CA certificate that will be added to the trusted system CAs.
+    # Useful if using an S3-compatible storage exposing a self-signed certificate.
+    # Example
+    #   caCert: |
+    #     -----BEGIN CERTIFICATE-----
+    #     MIIC7jCCApSgAwIBAgIRAIq2zQEVexqxvtxP6J0bXAwwCgYIKoZIzj0EAwIwgbkx
+    #     ...
+    caCert: null
+
 # Configuration for DNS configuration within the Kubernetes cluster.
 # This creates a service that routes to all agents (client or server)
 # for serving DNS requests. This DOES NOT automatically configure kube-dns
--- a/templates/client-snapshot-agent-deployment.yaml
+++ b/templates/client-snapshot-agent-deployment.yaml
@@ -99,6 +99,11 @@ spec:
             - "/bin/sh"
             - "-ec"
             - |
+              {{- if .Values.client.snapshotAgent.caCert }}
+              cat <<EOF > /etc/ssl/certs/custom-ca.pem
+              {{- .Values.client.snapshotAgent.caCert | nindent 14 }}
+              EOF
+              {{- end }}
               exec /bin/consul snapshot agent \
                 {{- if (and .Values.client.snapshotAgent.configSecret.secretName .Values.client.snapshotAgent.configSecret.secretKey) }}
                 -config-dir=/consul/config \
--- a/test/unit/client-snapshot-agent-deployment.bats
+++ b/test/unit/client-snapshot-agent-deployment.bats
@@ -362,3 +362,25 @@ load _helpers
       yq -rc '.spec.template.spec.containers[0].resources' | tee /dev/stderr)
   [ "${actual}" = '{"limits":{"cpu":"200m","memory":"200Mi"},"requests":{"cpu":"100m","memory":"100Mi"}}' ]
 }
+
+#--------------------------------------------------------------------
+# client.snapshotAgent.caCert
+
+@test "client/SnapshotAgentDeployment: if caCert is set it is used in command" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/client-snapshot-agent-deployment.yaml  \
+      --set 'client.snapshotAgent.enabled=true' \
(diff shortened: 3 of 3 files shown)


lkysow:
I've tested this as follows. Note that I did not test with an s3 compatible storage and a self-signed cert because that was going to be a lot more work to set up. Instead I tested with a golang binary that made https calls to a server with a self-signed cert and it correctly used the ca cert in `/etc/ssl/certs` and didn't complain about the signing.

SSL server with self-signed cert
```go
package main

import (
	"log"
	"net/http"
)

func HelloServer(w http.ResponseWriter, req *http.Request) {
	w.Header().Set("Content-Type", "text/plain")
	w.Write([]byte("This is an example server.\n"))
}

func main() {
	http.HandleFunc("/hello", HelloServer)
	err := http.ListenAndServeTLS(":443", "server.crt", "server.key", nil)
	if err != nil {
		log.Fatal("ListenAndServe: ", err)
	}
}
```

Golang binary that runs in a docker container:

```go
package main

import (
	"fmt"
	"io/ioutil"
	"net/http"
	"time"
)

func main() {
	for {
		time.Sleep(1 * time.Second)
		makeCall()
	}
}

func makeCall() {
	resp, err := http.Get("https://0.tcp.ngrok.io:13871/hello")
	if err != nil {
		fmt.Printf("err: %s\n", err)
		return
	}

	defer resp.Body.Close()
	body, err := ioutil.ReadAll(resp.Body)
	fmt.Printf("resp: %s\n", string(body))
}
```

When running locally:
```
err: Get "https://0.tcp.ngrok.io:13871/hello": x509: certificate signed by unknown authority
```

When running on kube with the following YAML it works:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test
  labels:
    app: test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test
  template:
    metadata:
      labels:
        app: test
    spec:
      containers:
        - name: test
          image: "lkysow/ssl-client"
          command:
            - "/bin/sh"
            - "-ec"
            - |
              cat <<EOF > /etc/ssl/certs/custom-ca.pem
              -----BEGIN CERTIFICATE-----
              MIICFjCCAZsCCQCdwLtdjbzlYzAKBggqhkjOPQQDAjB0MQswCQYDVQQGEwJDQTEL
              MAkGA1UECAwCQ0ExCzAJBgNVBAcMAkhpMQswCQYDVQQKDAJIaTELMAkGA1UECwwC
              SGkxFzAVBgNVBAMMDjAudGNwLm5ncm9rLmlvMRgwFgYJKoZIhvcNAQkBFgloaUBo
              aS5jb20wHhcNMjAwNjA1MTg1MTE0WhcNMzAwNjAzMTg1MTE0WjB0MQswCQYDVQQG
              EwJDQTELMAkGA1UECAwCQ0ExCzAJBgNVBAcMAkhpMQswCQYDVQQKDAJIaTELMAkG
              A1UECwwCSGkxFzAVBgNVBAMMDjAudGNwLm5ncm9rLmlvMRgwFgYJKoZIhvcNAQkB
              FgloaUBoaS5jb20wdjAQBgcqhkjOPQIBBgUrgQQAIgNiAAQUB68g5PDQZRdi6dtg
              zSMvnN4BLSW+mX1H7dVt4ZLsJlPTh2HnlVxdRJhkpmgzt4xY45orqkxEWIxIgAjG
              19SDA2sKoAPxfJUiwQ3lQGwPcjsYJbF0dz1hv6JmYDs1NQowCgYIKoZIzj0EAwID
              aQAwZgIxALwKYooFENHYoLgBO03kFJ8fqtTyidXscmVigG1jEAJB4SgU9P4FEQQO
              zyw1iDaFMQIxAMMqUETpxr6ZPnaWvaCjETHYwx/h8dqdt7s3KDMEb93sG63jdmJ+
              n8tKTS1x8ZYPRA==
              -----END CERTIFICATE-----
              EOF
              /go/bin/app
```

values.yaml
```diff
@@ -529,6 +529,15 @@ client:
         memory: "50Mi"
         cpu: "50m"
 
+    # Optional CA certificate that will be added to the trusted system CAs.
+    # Useful if using an S3-compatible storage exposing a self-signed certificate.
+    # Example
+    #   caCert: |
+    #     -----BEGIN CERTIFICATE-----
+    #     MIIC7jCCApSgAwIBAgIRAIq2zQEVexqxvtxP6J0bXAwwCgYIKoZIzj0EAwIwgbkx
+    #     ...
```
ishustava, line 538: For [`connectInject.certs.caBundle`](https://github.com/hashicorp/consul-helm/blob/df74135dbae999add1e9d71b3d0e72e72db691f3/values.yaml#L827-L830) we are asking that ca bundle should be provided as a base64 encoded string. We're probably doing it so that we can pass it through to the webhook registration, which requires it to be base64 encoded. 

I'm wondering if there is any reason for this value to also to be base64 encoded other than consistency. I don't think so since certificate data is already base64 encoded and so there isn't any danger for weird byte data, but curious what you think.

lkysow: I don't think so either. It's more work to base64 encode and more room for error.

values.yaml
```diff
@@ -529,6 +529,15 @@ client:
         memory: "50Mi"
         cpu: "50m"
 
+    # Optional CA certificate that will be added to the trusted system CAs.
```
ishustava, line 532:
```suggestion
    # Optional PEM-encoded CA certificate that will be added to the trusted system CAs.
```

ishustava (approved):
Looks good, but I haven't tested it myself.

PR #483, opened 2020-06-05
Add acceptance tests framework and tests for basic install, connect, and sync

ishustava:
Implement the test framework and a few tests to get started. 

### Running tests

Note: tests require go 1.14 as they use `t.Cleanup` function.

To run the tests:
```
cd test/acceptance/tests
go test -p 1 ./...
```
`-p 1` is very important because the tests currently cannot run in parrallel

### Changes in this PR

Framework includes the following:
- Suite
- Consul cluster with the underlying helm commands
- Test environment and context
- Adds some basic flags (`-kubeconfig`, `-kubecontext`, `-namespace`, `-consul-image`, and `-consul-k8s-image`). Other flags will be added in subsequent PRs based on need.

The tests I've added:
- Basic install test
- Basic install test with secure configuration
- Sync catalog test
- Sync catalog test with secure configuration
- Connect inject test
- Connect inject test with secure configuration

Other tests will be added in subsequent PRs.

These tests take about 6-7min to run currently.

### Notes:
I've added tests to some functions in the test framework mostly for functions that don't need to shell out or rely on any actual infrastructure to be there, but for the most part, the framework is tested in the acceptance tests themselves. I'd love to hear some feedback about that if you think it needs to be tested more.


test/acceptance/fixtures/static-server.yaml
```diff
@@ -0,0 +1,36 @@
...
+      name: static-server
+      labels:
+        app: static-server
+      annotations:
+        "consul.hashicorp.com/connect-inject": "true"
+    spec:
+      containers:
+        # This name will be the service name in Consul.
+        - name: static-server
+          image: ishustava/http-echo:latest
```
lkysow, line 26: did you build your own that actually shuts down on SIGTERM?

ishustava: I don't remember why I created my own image 😃 ! I should I update it to our basic one.

lkysow: I notice that one (not sure static-client or server) takes like 5m to shutdown on pod delete so it might take a long time for your test cleanup.

ishustava: Oh, I see what you mean. It still takes a while to terminate, but because it's a deployment, kubectl doesn't wait for it to finish terminating.

test/docker/Test.dockerfile
```diff
@@ -6,7 +6,7 @@
 # a script to configure kubectl, potentially install Helm, and run the tests
 # manually. This image only has the dependencies pre-installed.
 
-FROM alpine:3.11
+FROM circleci/golang:1.14
```
ishustava, line 9: I haven't built an image yet, but this change is needed so that we can run these tests in CI.

test/acceptance/framework/config_test.go
```diff
@@ -0,0 +1,59 @@
...
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t1 *testing.T) {
+			cfg := &TestConfig{
+				ConsulImage:    tt.fields.ConsulImage,
+				ConsulK8SImage: tt.fields.ConsulK8SImage,
+			}
+			require.Equal(t, cfg.HelmValuesFromConfig(), tt.want)
```
lkysow, line 56:
```suggestion
			require.Equal(t1, cfg.HelmValuesFromConfig(), tt.want)
```

Not sure about `t1` as the var but you should use whatever you call it here rather than the outer `t`

test/acceptance/framework/consul_cluster.go
```diff
@@ -0,0 +1,151 @@
...
+	releaseName string
+	kubernetesClient *kubernetes.Clientset
+}
+
+func NewHelmCluster(
+	t *testing.T,
+	helmValues map[string]string,
+	ctx TestContext,
+	cfg *TestConfig,
+	releaseName string,) Cluster {
```
lkysow, line 42:
```suggestion
	releaseName string) Cluster {
```

test/acceptance/framework/consul_cluster.go
```diff
@@ -0,0 +1,151 @@
...
+		Logger:         logger.TestingT,
+	}
+	return &HelmCluster{
+		helmOptions: opts,
+		releaseName: releaseName,
+		kubernetesClient: ctx.KubernetesClient(t),
+	}
+}
+
+func (h *HelmCluster) Create(t *testing.T) {
```
lkysow, line 67: Can this cleanup old envs or error out if there's already something running? I keep running into an issue where I Ctrl-C the tests, it doesn't clean up properly and then when I re-run the test it fails again because there's already a server running and it's taking up the only port.

test/acceptance/framework/consul_cluster.go
```diff
@@ -0,0 +1,151 @@
...
+	// todo: don't hard-code helm-chart path like this
```
lkysow, line 68: What did you want to do instead?

test/acceptance/framework/consul_cluster.go
```diff
@@ -0,0 +1,151 @@
...
+	helpers.WaitForAllPodsToBeReady(t, h.kubernetesClient, h.helmOptions.KubectlOptions.Namespace, fmt.Sprintf("release=%s", h.releaseName))
+}
+
+func (h *HelmCluster) Destroy(t *testing.T) {
+	helm.Delete(t, h.helmOptions, h.releaseName, false)
+
+	// delete PVCs
+	h.kubernetesClient.CoreV1().PersistentVolumeClaims(h.helmOptions.KubectlOptions.Namespace).DeleteCollection(&metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: "release="+h.releaseName})
+
+	// delete any secrets that have h.releaseName in their name
```
lkysow, line 83: Not something to do here but I think we need to label the secrets we create via consul-k8s for easier identification

test/acceptance/framework/consul_cluster.go
```diff
@@ -0,0 +1,151 @@
...
+	for _, secret := range secrets.Items {
+		if strings.Contains(secret.Name, h.releaseName) {
+			err := h.kubernetesClient.CoreV1().Secrets(h.helmOptions.KubectlOptions.Namespace).Delete(secret.Name, nil)
+			require.NoError(t, err)
+		}
+	}
+}
+
+func (h *HelmCluster) Upgrade(t *testing.T) {
+	t.Errorf("not implemented yet")
```
lkysow, line 95:
```
func (h *HelmCluster) Upgrade(t *testing.T, helmValues map[string]string) {
	// todo: don't hard-code helm-chart path like this
	h.helmOptions.SetValues = helmValues
	helm.Upgrade(t, h.helmOptions, "../../../..", h.releaseName)
	// todo: replace this with helm install --wait
	helpers.WaitForAllPodsToBeReady(t, h.kubernetesClient, h.helmOptions.KubectlOptions.Namespace, fmt.Sprintf("release=%s", h.releaseName))
}
```

Does this require more than ☝️ ?

test/acceptance/framework/consul_cluster.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func (h *HelmCluster) SetupConsulClient(t *testing.T, secure bool) *api.Client {
```
lkysow, line 98: Love this function, nice work.

lkysow (commented):
This is super cool, great work.

In addition to my comments:
1. Do we want to merge this before we get it working in CI?
1. I think we need to add some go fmt/vet to our .circleci/config.yml for this directory now
1. Could we add instructions on how to run these tests to contributing.md?

test/acceptance/framework/config_test.go
ishustava: Thank you. This is how goland auto-generated it, and I've missed this one. I Will update here and above!

test/acceptance/framework/consul_cluster.go
ishustava: I like this idea! I think erroring out is probably the best so that we don't accidentally delete anything you might actually need.

test/acceptance/framework/consul_cluster.go
ishustava: I couldn't come up with a good replacement actually and decided to keep it there just as a reminder that I need to think about it (before we merge the whole feature branch). If it doesn't seem 😬 to you, then maybe we can just keep it (and maybe make it a constant).

test/acceptance/framework/consul_cluster.go
ishustava: Yes! totally agree!

test/acceptance/framework/consul_cluster.go
ishustava: That looks right to me, except that it doesn't need to take in helm values. I think it'd be better if you update them on the `HelmCluster` object itself.

test/acceptance/framework/consul_cluster.go
ishustava: thank you!

test/acceptance/helpers/helpers.go
```diff
@@ -0,0 +1,54 @@
...
+	"time"
+
+	"github.com/gruntwork-io/terratest/modules/random"
+	"github.com/hashicorp/consul/sdk/testutil/retry"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+func RandomName() string {
```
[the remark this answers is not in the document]
ishustava: To some of these, I didn't add docs because they seemed "self-explanatory". I'm curious what you think. Do you think we should generally add comments/docs to all functions? For example, `KubectlDelete` function just does what its name is, so the doc for it would seem like it will just rephrase the function name as an English sentence. Perhaps, I'm missing something though.

test/acceptance/tests/basic/main_test.go
```diff
@@ -0,0 +1,15 @@
+package basic
+
+import (
+	"os"
+	"testing"
+
+	"github.com/hashicorp/consul-helm/test/acceptance/framework"
+)
+
+var suite framework.Suite
```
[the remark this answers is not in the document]
ishustava: yes, I totally understand. Maybe this seems weird because the `MainTest` and the "actual" tests are in different files. Would you feel differently if they all were in one file? We're using global `var`s all the time in consul-k8s tests ([example](https://github.com/hashicorp/consul-k8s/blob/9bbf6df4e23b9b1a6bf251fe88215328b967520d/subcommand/server-acl-init/command_test.go)). 

Also, as an example, you could look at istio's integration tests, where they have global vars in the `main_test.go` file; those vars are shared and used by the tests within a package. [Here](https://github.com/istio/istio/blob/8f24e6ca078ac12c02f31e5bc00b038a1359d472/tests/integration/pilot/main_test.go) is an example.

The parallelism concerns would be similar to having global vars in a single test file: you'd have to make sure no two tests are writing to it at the same time. In our case, we'd have to be careful that tests aren't modifying the environment and are just reading info they need. I could be missing something here, but I think we can make it work.

test/acceptance/tests/connect/connect_inject_test.go
```diff
@@ -0,0 +1,105 @@
...
+	require.NoError(t, err)
+	require.Len(t, pods.Items, 1)
+
+	retry.Run(t, func(r *retry.R) {
+		output, err := helpers.RunKubectlAndGetOutputE(t, options, "exec", pods.Items[0].Name, "-c", "static-client", "--", "curl", "-vvvs", "http://127.0.0.1:1234/")
+		if expectSuccess {
+			require.NoError(r, err)
+			require.Contains(r, output, "hello world")
+		} else {
+			require.Error(t, err)
```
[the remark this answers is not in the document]
ishustava: Yes, that's a great catch!

test/acceptance/tests/basic/main_test.go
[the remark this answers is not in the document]
lkysow: Gotcha, so as long as we don't have any mutation of `suite` then we're good.

test/acceptance/helpers/helpers.go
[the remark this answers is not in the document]
lkysow: Well randomname seems self explanatory but I see it prepends `test-`. I think it'll be easier to have a blanket "document all exported helpers" convention than have each eng decide for themselves.

test/acceptance/framework/consul_cluster.go
lkysow: Yeah if there's no other way then this isn't terrible.

test/acceptance/framework/consul_cluster.go
lkysow: Right, makes sense.

test/acceptance/framework/consul_cluster.go
ishustava: I've added erroring and also set up things so that they are cleaned up when you ctrl-c. I've added `helpers.Cleanup` function that can be called instead of `t.Cleanup`. This function will set up an interrupt handler that will call the same cleanup function in case an interrupt signal is caught. Let me know what you think!

test/acceptance/framework/consul_cluster.go
ishustava: Made it a constant.

test/acceptance/framework/consul_cluster.go
ishustava: Added this implementation, although I haven't tried it yet.

test/acceptance/framework/consul_cluster_test.go
```diff
@@ -0,0 +1,17 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// Test that if TestConfig has values that need to provided
+// to a helm install, it will respect the helmValues over
+// the values from TestConfig.
+func TestNewHelmCluster(t *testing.T) {
+	helmValues := map[string]string{"global.image": "test-image", "feature.enabled": "true"}
```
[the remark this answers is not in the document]
ishustava: I've added those values but set them to `3` instead so that it's different from default.

test/acceptance/tests/basic/main_test.go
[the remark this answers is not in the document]
ishustava: yes, I believe so.

test/acceptance/tests/sync/sync_catalog_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+			releaseName := helpers.RandomName()
+			consulCluster := framework.NewHelmCluster(t, helmValues, env.DefaultContext(t), suite.Config(), releaseName)
+
+			consulCluster.Create(t)
+
+			t.Logf("creating a test service and pod called %s", releaseName)
+			createTestService(t, env.DefaultContext(t).KubernetesClient(t), releaseName)
+
+			consulClient := consulCluster.SetupConsulClient(t, false)
```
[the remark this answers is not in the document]
ishustava: Good catch! 🤦‍♀️ that is because I forgot to provide helmValues from the test case. Instead, I'm using the same ones (that are set up a few lines above) in both test scenarios. Fixed!

test/acceptance/framework/consul_cluster.go
ishustava: created #546 to track this because tls-init also needs to do that.

ishustava:
Thanks so much for the review, Luke! I've made some changes based on your reivew.

>    Do we want to merge this before we get it working in CI?

Yes. I'd like to merge that in separately into the feature branch and create another PR for CI changes. My motivation is that I'd like to review it separately, so that it's easier to review without all the other changes. I

>    I think we need to add some go fmt/vet to our .circleci/config.yml for this directory now

Yes! I'll make sure this is part of the changes for CI in future PR.

> Could we add instructions on how to run these tests to contributing.md?

I'd like to add instructions in a different PR if that's OK.

lkysow (approved):
🎉 !

PR #490, opened 2020-06-09
Update comments to match consul-k8s envoy version

i0rek:
See https://github.com/hashicorp/consul-k8s/pull/273 and https://github.com/hashicorp/consul/pull/8057.


ishustava:
Closing this too because we kept the old version for now. We've updated the default version for all gateways though in #502. 

We will figure out a better approach to how we're updating envoy versions going forward. Thanks again for the PR, Hans!

PR #492, opened 2020-06-11
Added securityContext as a param for consul client pods

domgoodwin:
Not having the ability to set securityContext means in environments where you're running PSPs that enforce runAsNonRoot you can't. Even though the consul docker image is non root because in the DockerFile `USER 100` isn't set it won't guess the UID. 

This allows all the securityContext options as per: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/


lkysow:
Thanks for the PR! How come you only need to run this on the consul clients and not the consul servers?

domgoodwin:
It just so happened in the deployment I was using it was just client pods. I've added the same for servers now.

PR #431, opened 2020-04-23
Feature/client host networking

kieranajp:
Hi,

After running into the issue described in #327, I added the workaround from that comment thread, which works perfectly.

Please let me know if this needs anything more from me.

Cheers,
--Kieran

PR #496, opened 2020-06-15
add support for client host networking and tests

kschoche:
Add support for client host networking along with unit tests and podsecurity support.

It should be duly noted that this feature introduces security considerations which are outlined in values.yaml 

This is an extension of @kieranajp 's #431  and should also resolve #193 and #431

### DIFF
--- a/templates/client-podsecuritypolicy.yaml
+++ b/templates/client-podsecuritypolicy.yaml
@@ -26,7 +26,7 @@ spec:
     {{- if .Values.client.dataDirectoryHostPath }}
     - 'hostPath'
     {{- end }}
-  hostNetwork: false
+  hostNetwork: {{ .Values.client.hostNetwork }}
   hostPorts:
   {{- if (not (and .Values.global.tls.enabled .Values.global.tls.httpsOnly)) }}
   # HTTP Port
--- a/templates/client-podsecuritypolicy.yaml
+++ b/templates/client-podsecuritypolicy.yaml
@@ -26,7 +26,11 @@ spec:
     {{- if .Values.client.dataDirectoryHostPath }}
     - 'hostPath'
     {{- end }}
+  {{- if .Values.client.hostNetwork }}
   hostNetwork: {{ .Values.client.hostNetwork }}
+  {{- else }}
+  hostNetwork: false
+  {{- end }}
   hostPorts:
   {{- if (not (and .Values.global.tls.enabled .Values.global.tls.httpsOnly)) }}
   # HTTP Port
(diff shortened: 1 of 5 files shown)


templates/client-podsecuritypolicy.yaml
```diff
@@ -26,7 +26,7 @@ spec:
     {{- if .Values.client.dataDirectoryHostPath }}
     - 'hostPath'
     {{- end }}
-  hostNetwork: false
+  hostNetwork: {{ .Values.client.hostNetwork }}
```
ashwin-venkatesh, line 29: qq: does this need to be wrapped by an `if` check to ensure this value is set only is `.Values.client.hostNetwork` is provided in the values file? I suspect it would be set to `null`/`nil` otherwise which I suspect is an invalid value?

test/unit/client-podsecuritypolicy.bats
```diff
@@ -129,3 +129,16 @@ load _helpers
...
+@test "client/PodSecurityPolicy: enabled with global.enablePodSecurityPolicies=true and hostNetwork=true" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -x templates/client-podsecuritypolicy.yaml  \
+      --set 'global.enablePodSecurityPolicies=true' \
+      --set 'client.hostNetwork=true' \
+      . | tee /dev/stderr |
+      yq '.spec.hostNetwork == true' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
```
ishustava, line 144: could you add another test checking that it's set to false by default. We didn't test it in the past because there was no templating logic, but now that there is I think it's worth adding a test.

values.yaml
```diff
@@ -489,6 +489,13 @@ client:
   # dnsPolicy to use.
   dnsPolicy: null
 
+  # hostNetwork defines whether or not we use host networking instead of hostPort in the event
+  # that a CNI plugin doesnt support hostPort. This has security implications and is not recommended
+  # as doing so gives the consul client unnecessary access to all network traffic on the host.
+  # In most cases, pod network and host network are on different subnets so this should be
```
ishustava, line 495: I feel that subnet is a bit of an IaaS-specific term. I could be wrong though, so this suggestion could be ignored!
```suggestion
  # In most cases, pod network and host network are on different networks so this should be
```

test/unit/client-podsecuritypolicy.bats
kschoche: good catch, I'll add one for this, thanks!

Issue #403, opened 2020-03-31
Not enough permissions to create clusterroles


rakshitcgupta:
I am getting RBAC errors when trying to install consul via helm charts on Openshift because of the restrictions. I have to edit individual files for the consul to deploy.
The error I am getting is as follows- 

$ helm install --name hashicorp ./consul-helm -f config.yaml
Error: release hashicorp failed: clusterroles.rbac.authorization.k8s.io is forbidden: User "system:serviceaccount:dk1280-a:sa-edit" cannot create clusterroles.rbac.authorization.k8s.io at the cluster scope: User "system:serviceaccount:dk1280-a:sa-edit" cannot create clusterroles.rbac.authorization.k8s.io at the cluster scope

adilyse:
Hi @rakshitcgupta,

It looks like the account you're using doesn't have the permissions needed to created the associated RBAC clusterrroles. We've generally assumed that folks installing Consul are doing so with operator permissions. If that isn't true for you, could you tell us a bit more about your use case? Additionally, it would be helpful to know what sort of solution you would like to see that would solve this for you.

Thanks!

rakshitcgupta:
Hi @adilyse 

For now, we are doing poc in a single namespace and not the cluster. Therefore cannot have admin privileges to create RBAC clusterroles. I am looking for solution to disable the RBAC clusterroles or maybe strip some code which will deploy consul on a namespace and not cluster wide.

Thanks & Regards

adilyse:
Hi @rakshitcgupta,

Disabling RBAC for certain features used to be an option in the project, but it wasn't implemented cleanly so we decided to remove it. The fact that the managed Kubernetes available from the major cloud providers all had RBAC enabled by default also played a role in the decision.

To your point, however, it would be helpful to have the option to install Consul in a less privileged manner for folks testing it out. We appreciate the feedback and will keep this in mind as we fine tune our day zero experience.

lkysow:
Also related: https://github.com/hashicorp/consul-helm/issues/354 (we don't need cluster level perms for a lot of our stuff) and https://github.com/hashicorp/consul-helm/issues/146 (make disabling rbac possible)

rakshitcgupta:
Hi @adilyse @lkysow 

Temporarily, how can I convert cluster roles to roles. What files can I change to achieve that and make it work in a restricted environment.

ishustava:
Hey @rakshitcgupta,

[...]

PR #498, opened 2020-06-16
Do not use ClusterRole/Bindings unnecessarily for tls-init templates

### LINKED ISSUE #354: more strict RBAC rules wherever possible
Hi,

I'm looking at the resources generated by the chart when TLS is enabled.
I *think* it's possible to be a bit more strict, with regards to RBAC rules.
For example, for the _tls-init_ and _tls-init-cleanup_ stuff, couldn't _Role_ and _RoleBinding_ be used instead of _ClusterRole_ and _ClusterRoleBinding_? Everything happens in one namespace, right, so the _tls-init_ service account does not need to be able to create secrets in ALL namespaces.

No?

kschoche:
It is not necessary to use ClusterRole and ClusterRoleBinding's for tls-init and tls-init-cleanup, these can be standard Role and RoleBinding's as they do not require access to multiple namespaces.
This resolves #354 and partially addresses #403, follow-up PR's will individually address other components.
The PR : 
1. changes the tls-init-* templates to use standard Role / RoleBindings
2. renames the template files from *-clusterrole.yaml to *-role.yaml
3. modifies the unit tests to reflect changes from 1 + 2
4. renames the unit test bats files to reflect these changes

This PR does not reduce the permissions allotted by the Roles, just the number of namespaces accessible.

### DIFF
--- a/templates/tls-init-cleanup-role.yaml
+++ b/templates/tls-init-cleanup-role.yaml
@@ -1,7 +1,7 @@
 {{- if (or (and (ne (.Values.server.enabled | toString) "-") .Values.server.enabled) (and (eq (.Values.server.enabled | toString) "-") .Values.global.enabled)) }}
 {{- if .Values.global.tls.enabled }}
 apiVersion: rbac.authorization.k8s.io/v1beta1
-kind: ClusterRoleBinding
+kind: Role
 metadata:
   name: {{ template "consul.fullname" . }}-tls-init-cleanup
   namespace: {{ .Release.Namespace }}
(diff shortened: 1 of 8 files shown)


kschoche:
For whatever reason GitHub is displaying this PR as if I had renamed the bats files incorrectly, probably due to the fact the 4 bats files have identical contents and they were all renamed + modified.

lkysow:
> It does leave some clusterroles and clusterrolebindings behind, so I think we'd need to mention it in the changelog.

Interesting! I was testing the other PR and they were deleted. Maybe it's because these are hooks?

ishustava:
> Interesting! I was testing the other PR and they were deleted. Maybe it's because these are hooks?

Yep, I think you're right. This is probably because it's a hook. I saw that for other components helm will delete the old cluster roles/rolebindings.

PR #500, opened 2020-06-17
Support disabling creation of helm cleanup hooks

chancez:
When using this chart with argo-cd, kustomize, and many other tools, the
helm annotations aren't used, or are used differently than how helm uses
them, and results in these cleanup hook resources being created at the
same time as everything else.

To better support these tools, add a new global.tls.disableCleanupHooks
option to disable creating the helm hook resources responsible for doing
cleanup actions.


chancez:
One thought I also had was if this should be `global.hooks.disableCleanup` or `hooks.disableCleanup` , as it may scale better for disabling other hooks.

chancez:
@adilyse Got it. Sounds good. I'll take a look at updating this later this week.

ishustava:
Hey @chancez, just checking in. Are you still working on this PR?

chancez:
I've become a bit undecided if this is the best choice to expose these options. I've been working around it by using Kustomize patches to remove unneeded resources, and it's not the worst.

lkysow:
Since kustomize is working for you then we'll close this. If other users need this feel free to comment and we can look at reopening.

PR #503, opened 2020-06-18
Support terminating gateways

adilyse:
Following up from #491 after a tragic git accident.

This creates all of the necessary templates to support terminating
gateways in the helm chart. Using terminatingGateways.defaults and
terminatingGateways.gateways, it's possible to define multiple
gateways. Names must be provided for each and they must be unique.

Any gateway-specific values will override the default value with the
exception of annotations. Annotations will apply both the defaults
and gateway-specific annotations.

Note: This is ready for a static code review, but has not been tested
yet because of a bug in the gateway implementation. It's also targeting
the ingress gateway branch for now to make it easier for settings and
tests that overlap between gateway implementations.

Replaces and builds on #452

### DIFF
--- a/test/unit/terminating-gateways-deployment.bats
+++ b/test/unit/terminating-gateways-deployment.bats
@@ -724,6 +724,7 @@ cat > /consul/service/service.hcl << EOF
 service {
   kind = "terminating-gateway"
   name = "terminating-gateway"
+  id = "${POD_NAME}"
   address = "${POD_IP}"
   port = 8443
   checks = [
@@ -762,6 +763,7 @@ cat > /consul/service/service.hcl << EOF
 service {
   kind = "terminating-gateway"
   name = "terminating-gateway"
+  id = "${POD_NAME}"
   address = "${POD_IP}"
   port = 8443
   checks = [
@@ -798,6 +800,7 @@ cat > /consul/service/service.hcl << EOF
 service {
   kind = "terminating-gateway"
   name = "terminating-gateway"
+  id = "${POD_NAME}"
   namespace = "namespace"
   address = "${POD_IP}"
   port = 8443
@@ -836,6 +839,7 @@ cat > /consul/service/service.hcl << EOF
 service {
   kind = "terminating-gateway"
   name = "terminating-gateway"
+  id = "${POD_NAME}"
   namespace = "new-namespace"
   address = "${POD_IP}"
   port = 8443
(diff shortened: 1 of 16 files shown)


test/unit/terminating-gateways-deployment.bats
```diff
@@ -0,0 +1,949 @@
...
+
+  local actual=$(echo $object | yq '.[0] | length > 0' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+
+  local actual=$(echo $object | yq '.[1] | length > 0' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+
+  local actual=$(echo $object | yq '.[2] | length > 0' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
+}
```
ishustava, line 949: I think we're missing a test case checking that `CONSUL_HTTP_TOKEN` is set if `global.manageSystemACLs` is set for the `terminating-gateway` and `lifecycle-sidecar` containers.

values.yaml
```diff
@@ -1178,6 +1177,88 @@ ingressGateways:
...
     - name: ingress-gateway
 
+# Configuration options for terminating gateways. Default values for all
+# terminating gateways are defined in `terminatingGateways.defaults`. Any of
+# these values may be overridden in `terminatingGateways.gateways` for a
+# specific gateway with the exception of annotations. Annotations will
+# include both the default annotations and any additional ones defined
+# for a specific gateway.
+# Requirements: consul >= 1.8.0 and consul-k8s >= 0.16.0 if using
+# global.acls.manageSystemACLs.
```
ishustava, line 1187: For the lifecycle-sidecar command
```suggestion
# global.acls.manageSystemACLs and >= 0.10.0 if not using global.acls.manageSystemACLs.
```

values.yaml
```diff
@@ -1178,6 +1177,88 @@ ingressGateways:
...
+terminatingGateways:
+  # Enable terminating gateway deployment. Requires `connectInject.enabled=true`.
```
ishustava, line 1189: Should we inlcude `client.enabled` and `client.grpc` as a requirement as well?

values.yaml
```diff
@@ -1178,6 +1177,88 @@ ingressGateways:
...
+  enabled: false
+
+  # Defaults sets default values for all gateway fields. With the exception
+  # of annotations, defining any of these values in the `gateways` list
+  # will override the default values provided here.
```
ishustava, line 1194: I thought it's worth repeating the same thing as you said in the comment for `terminatingGateways` here for clarity.
```suggestion
  # will override the default values provided here. Annotations will
  # include both the default annotations and any additional ones defined 
  # for a specific gateway.
```

values.yaml
```diff
@@ -1178,6 +1177,88 @@ ingressGateways:
...
+      requests:
+        memory: "100Mi"
+        cpu: "100m"
+      limits:
+        memory: "100Mi"
+        cpu: "100m"
+
+    # By default, we set an anti affinity so that two of the same gateway pods
+    # won't be on the same node. NOTE: Gateways require that Consul client agents are
+    # also running on the nodes alongside each gateway Pod.
```
ishustava, line 1221:
```suggestion
    # also running on the nodes alongside each gateway pod.
```

values.yaml
```diff
@@ -1178,6 +1177,88 @@ ingressGateways:
...
+    # Resource limits for all terminating gateway pods
+    resources:
+      requests:
+        memory: "100Mi"
+        cpu: "100m"
+      limits:
+        memory: "100Mi"
+        cpu: "100m"
+
+    # By default, we set an anti affinity so that two of the same gateway pods
```
ishustava, line 1219:
```suggestion
    # By default, we set an anti-affinity so that two of the same gateway pods
```

test/unit/terminating-gateways-deployment.bats
adilyse: Added a test for the `terminating-gateway` container's `CONSUL_HTTP_TOKEN` and the `-token-file` flag on the lifecycle sidecar command.

values.yaml

values.yaml
adilyse: Added `client.enabled`, but `client.rpc` is enabled by default so leaving it off.

values.yaml

values.yaml
adilyse: Fixed on all 3 gateways.

values.yaml
adilyse: Fixed on all 3 gateways.

test/unit/terminating-gateways-deployment.bats
```diff
@@ -0,0 +1,949 @@
...
+      yq -s -r '.[0].spec.template.spec.containers[0].env[]' | tee /dev/stderr)
+
+  local actual=$(echo $env | jq -r '. | select(.name == "CONSUL_HTTP_ADDR") | .value' | tee /dev/stderr)
+  [ "${actual}" = 'https://$(HOST_IP):8501' ]
+
+  local actual=$(echo $env | jq -r '. | select(.name == "CONSUL_GRPC_ADDR") | .value' | tee /dev/stderr)
+  [ "${actual}" = 'https://$(HOST_IP):8502' ]
+
+  local actual=$(echo $env | jq -r '. | select(.name == "CONSUL_CACERT") | .value' | tee /dev/stderr)
+  [ "${actual}" = "/consul/tls/ca/tls.crt" ]
```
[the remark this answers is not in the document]
ishustava: That makes sense, thanks for adding it in all tests!

test/unit/terminating-gateways-deployment.bats
```diff
@@ -0,0 +1,949 @@
...
+}
```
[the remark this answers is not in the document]
ishustava: Ah, yes those tests are great. I was referring to the TLS volumes for the Consul client, e.g. `consul-ca-cert`. Sorry, that wasn't clear, but I meant it in the context of `global.tls.enabled`.

PR #506, opened 2020-06-18
remove unnecessary server ClusterRoles

kschoche:
It is not necessary to use ClusterRoles for all server components as they operate within a singular namespace, we can instead use Role/RoleBinding.

This PR : 
1. Changes ClusterRole/Binding to Role/Binding for server-acl-* and server-role/binding templates
2. Renames the template files accordingly
3. Modifies bats unit tests to point to the new template names and renames them accordingly

This partially addresses #403, and is a follow-on to PR #498

### DIFF
--- a/templates/server-acl-init-cleanup-rolebinding.yaml
+++ b/templates/server-acl-init-cleanup-rolebinding.yaml
@@ -18,6 +18,5 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: {{ template "consul.fullname" . }}-server-acl-init-cleanup
-    namespace: {{ .Release.Namespace }}
 {{- end }}
 {{- end }}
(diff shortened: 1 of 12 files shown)


kschoche:
There is an added complexity here in the upgrade path in that we are leaving the old ClusterRole/Bindings in place and unused.  Will need to document a safe cleanup procedure.

templates/server-acl-init-cleanup-rolebinding.yaml
```diff
@@ -2,17 +2,18 @@
...
     chart: {{ template "consul.chart" . }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
+  kind: Role
   name: {{ template "consul.fullname" . }}-server-acl-init-cleanup
 subjects:
   - kind: ServiceAccount
```
lkysow, line 19: I think it makes sense to remove the
```yaml
    namespace: {{ .Release.Namespace }}
```

line below for all rolebindings since the namepace is implied now.

kschoche: Ah good catch on both points!
I'll update the rolebindings in this and the other PRs related to clusterroles to remove the duplicate namespace entry from the `subject:`!
Thanks!

PR #526, opened 2020-07-05
Add readiness probe params to client (#1)

dianazv:
* add readinessProbe parameters for client daemonSet template

* add readinessProbe parameters for client daemonSet template

* remove conditionaks from readinessProbe

* add examples for readinessProbe

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -297,6 +297,9 @@ spec:
                   curl http://127.0.0.1:8500/v1/status/leader \
                   {{- end }}
                   2>/dev/null | grep -E '".+"'
+            {{- if .Values.client.readinessProbe }}      
+            {{- .Values.client.readinessProbe | toYaml | nindent 12 }}
+            {{- end}}
           {{- if .Values.client.resources }}
           resources:
             {{- if eq (typeOf .Values.client.resources) "string" }}
--- a/values.yaml
+++ b/values.yaml
@@ -544,6 +544,14 @@ client:
     #     MIIC7jCCApSgAwIBAgIRAIq2zQEVexqxvtxP6J0bXAwwCgYIKoZIzj0EAwIwgbkx
     #     ...
     caCert: null
+    
+    # Optional fields to control the thresholds of readiness checks.
+    # Example
+    # failureThreshold: 3
+    # periodSeconds: 10
+    # successThreshold: 1
+    # timeoutSeconds: 1
+    readinessProbe: null
 
 # Configuration for DNS configuration within the Kubernetes cluster.
 # This creates a service that routes to all agents (client or server)


dianazv:
Hi @lkysow - wanted to ping you kindly and ask what is the status of the pr and see if there is anything else I can add. Thanks!

kschoche:
Hi! thank you for submitting this PR!
I think maybe it would be clearer to provide these as default settings to the client-daemonset similar to how it is done in the server-statefulset template?  This would allow you to override it in your own values file as well.
Since it is the same readinessProbe in each template they could use the same defaults.

```
diff --git a/templates/client-daemonset.yaml b/templates/client-daemonset.yaml
index 9c5fa9f..f9e62ac 100644
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -301,6 +301,11 @@ spec:
                   curl http://127.0.0.1:8500/v1/status/leader \
                   {{- end }}
                   2>/dev/null | grep -E '".+"'
+            failureThreshold: 2
+            initialDelaySeconds: 5
+            periodSeconds: 3
+            successThreshold: 1
+            timeoutSeconds: 5
           {{- if .Values.client.resources }}
           resources:
             {{- if eq (typeOf .Values.client.resources) "string" }}
```

I've tested this out and it has the added benefit of increasing the period from kubernetes default of 1s to 3s so that we're not probing 3x more often than for the servers.

Thoughts?

dianazv:
Hi, thanks for your comment!
So if I understand you correctly, then I should add defaults to the client-daemonset template as in the server-statefulset template. But if I don't add a check in the template to see if these parameters exist in my values file as well how can I override them?
Thanks!

dianazv:
hi, wanted to make sure you read my last comment, thanks

elcomtik:
This would be great enhancement for me, I would appreciate it if this PR was merged.

lkysow:
Hi @dianazv and @elcomtik,
Sorry, looks like this one fell through the cracks. @dianazv I think this is what we'd suggested:

In `values.yaml`:
```yaml
server:
  # YAML template string specifying the readiness probe.
  # This string can contain Helm template statements.
  readinessProbe: |
    exec:
      command:
        - "/bin/sh"
        - "-ec"
        - |
          curl \
            --cacert /consul/tls/ca/tls.crt \
            https://127.0.0.1:8501/v1/status/leader \
          {{- else }}
          curl http://127.0.0.1:8500/v1/status/leader \
          {{- end }}
          2>/dev/null | grep -E '".+"'
    failureThreshold: 2
    initialDelaySeconds: 5
    periodSeconds: 3
    successThreshold: 1
    timeoutSeconds: 5
```

And then in `server-statefulset.yaml`:
```yaml
          readinessProbe:
            {{ tpl .Values.server.readinessProbe . | nindent 12 | trim }}
```

We'd do the same for the client helm values.

t-eckert:

PR #528, opened 2020-07-06
allow to disable fsgroup, when installing to openshift

bakito:
When trying to install this chart to OpenShift, the StatefulSet fails, since the fsGroup is set by default to 1000, which is not valid in OpenShift by default.

Adding this option allows the fsGroup not to be set, when running in OpenShift.


bakito:
Hi

Thank you for the review.

I've applied the changes as requested.

Initially I've named the property openshift, to have it work the same as it is implemented in the vault-helm.

Once the chart officially, the fsGroup property could probably be changed back to an openshift flag.

Regards,
Marc

PR #531, opened 2020-07-06
Pass extraVolumes last for -config-dir

lkysow:
This will allow user-defined config to override any config set in the
helm-created config map because Consul will override previous
configuration with the last -config-dir.

For example, this user (https://github.com/hashicorp/consul-k8s/pull/220) wants to specify the `"acl": {"down_policy": "your-policy"}` config which is set in `/consul/aclconfig` by the helm chart (technically it's written by `consul-k8s acl-init`). By swapping the ordering here, they can now create their own configmap which sets
```
"acl": {"down_policy": "their-policy"}
```

and pass it in via

```yaml
client:
  extraVolumes:
  - type: configMap
     name: my-config-map
     load: true
```

And that single key will be overridden.

Fixes https://github.com/hashicorp/consul-k8s/pull/220

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Unreleased
 
+IMPROVEMENTS:
+
+* The volumes set by `client.extraVolumes` are now passed as the last `-config-dir` argument.
+  This means any settings there will override previous settings. This allows users to override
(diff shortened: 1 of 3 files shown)


kschoche (approved):
Looks great!

thisisnotashwin (approved):
NICE!

PR #532, opened 2020-07-07
refactor resource requests and limits for init containers and lifecycle sidecar

### LINKED ISSUE #515: lifecycle-sidecar at 100% of CPU limit
I'm seeing right at 100% of the requested limit CPU usage for the lifecycle sidecar. Would it be possible to increase, or allow configuration of the resource requests for this.

I _think_ https://github.com/hashicorp/consul-helm/blob/e2fb73ca3fa00074ceec26f6b4b15dd860a094ad/templates/mesh-gateway-deployment.yaml#L359-L365 is the spot that would need to allow for configuration.

### LINKED ISSUE #529: consul-connect-inject-init terminates
Hi, 

we were working with Consul 1.6 without any problem. 
But just after upgrading to Consul 1.8 on k8s, we are facing these problems:

consul-connect-inject-init pod randomly throws error. I tried to see what is wrong and I found this log once:
```
kubectl logs ... consul-connect-inject-init
Registered service: x-sidecar-proxy
Registered service: x
cp: write error: Out of memory
```
I think it happens here:
```
      /bin/consul services register \
        /consul/connect-inject/service.hcl
      
      # Generate the envoy bootstrap code
      /bin/consul connect envoy \
        -proxy-id="${PROXY_SERVICE_ID}" \
        -bootstrap > /consul/connect-inject/envoy-bootstrap.yaml
      
      # Copy the Consul binary
      cp /bin/consul /consul/connect-inject/consul

```
On that time there was not any problem with my k8s cluster but there is a resource section that seems to be new in `consul-connect-inject-init`: 
```
    name: consul-connect-inject-init
    resources:
      limits:
        cpu: 50m
        memory: 25Mi
      requests:
        cpu: 50m
        memory: 25Mi
``` 
and I suspect that might cause errors. I tried to override in `values.yml` it but it was not possible.

kschoche:
We now define resources for the connect inject init copy container in values.yaml and standardize usage of CPU Limit/Request and Memory Limit/Request for it across the ingress/terminating/mesh gateways and connect-inject deployments via new entries in the values.yaml file.

These are passed in as command line arguments to consul-k8s and are thereby are now configurable.
This PR resolves #515 and resolves #529 by increasing the default cpu request+limit and memory limit, respectively, making them configurable to the user.

### DIFF
--- a/templates/connect-inject-deployment.yaml
+++ b/templates/connect-inject-deployment.yaml
@@ -138,6 +138,34 @@ spec:
                 {{- if not (kindIs "invalid" $resources.requests.cpu) }}
                 -default-sidecar-proxy-cpu-request={{ $resources.requests.cpu }} \
                 {{- end }}
+                {{- $resources := .Values.global.initCopyContainerResources.resources }}
+                {{- /* kindIs is used here to differentiate between null and 0 */}}
+                {{- if not (kindIs "invalid" $resources.limits.memory) }}
+                -init-copy-container-memory-limit={{ $resources.limits.memory }} \
+                {{- end }}
+                {{- if not (kindIs "invalid" $resources.requests.memory) }}
+                -init-copy-container-memory-request={{ $resources.requests.memory }} \
+                {{- end }}
+                {{- if not (kindIs "invalid" $resources.limits.cpu) }}
+                -init-copy-container-cpu-limit={{ $resources.limits.cpu }} \
+                {{- end }}
+                {{- if not (kindIs "invalid" $resources.requests.cpu) }}
+                -init-copy-container-cpu-request={{ $resources.requests.cpu }} \
+                {{- end }}
+                {{- $resources := .Values.global.lifecycleSidecarContainerResources.resources }}
+                {{- /* kindIs is used here to differentiate between null and 0 */}}
+                {{- if not (kindIs "invalid" $resources.limits.memory) }}
+                -lifecycle-sidecar-memory-limit={{ $resources.limits.memory }} \
+                {{- end }}
+                {{- if not (kindIs "invalid" $resources.requests.memory) }}
+                -lifecycle-sidecar-memory-request={{ $resources.requests.memory }} \
+                {{- end }}
+                {{- if not (kindIs "invalid" $resources.limits.cpu) }}
+                -lifecycle-sidecar-cpu-limit={{ $resources.limits.cpu }} \
+                {{- end }}
+                {{- if not (kindIs "invalid" $resources.requests.cpu) }}
+                -lifecycle-sidecar-cpu-request={{ $resources.requests.cpu }} \
+                {{- end }}
           livenessProbe:
             httpGet:
               path: /health/ready
--- a/templates/connect-inject-deployment.yaml
+++ b/templates/connect-inject-deployment.yaml
@@ -138,22 +138,20 @@ spec:
                 {{- if not (kindIs "invalid" $resources.requests.cpu) }}
                 -default-sidecar-proxy-cpu-request={{ $resources.requests.cpu }} \
                 {{- end }}
-                {{- $resources := .Values.global.initCopyContainerResources.resources }}
-                {{- /* kindIs is used here to differentiate between null and 0 */}}
+                {{- $resources := .Values.global.initContainer.resources }}
                 {{- if not (kindIs "invalid" $resources.limits.memory) }}
-                -init-copy-container-memory-limit={{ $resources.limits.memory }} \
+                -init-container-memory-limit={{ $resources.limits.memory }} \
                 {{- end }}
                 {{- if not (kindIs "invalid" $resources.requests.memory) }}
-                -init-copy-container-memory-request={{ $resources.requests.memory }} \
+                -init-container-memory-request={{ $resources.requests.memory }} \
                 {{- end }}
                 {{- if not (kindIs "invalid" $resources.limits.cpu) }}
-                -init-copy-container-cpu-limit={{ $resources.limits.cpu }} \
+                -init-container-cpu-limit={{ $resources.limits.cpu }} \
                 {{- end }}
                 {{- if not (kindIs "invalid" $resources.requests.cpu) }}
-                -init-copy-container-cpu-request={{ $resources.requests.cpu }} \
+                -init-container-cpu-request={{ $resources.requests.cpu }} \
                 {{- end }}
-                {{- $resources := .Values.global.lifecycleSidecarContainerResources.resources }}
-                {{- /* kindIs is used here to differentiate between null and 0 */}}
+                {{- $resources := .Values.global.lifecycleSidecarContainer.resources }}
                 {{- if not (kindIs "invalid" $resources.limits.memory) }}
                 -lifecycle-sidecar-memory-limit={{ $resources.limits.memory }} \
                 {{- end }}
--- a/templates/ingress-gateways-deployment.yaml
+++ b/templates/ingress-gateways-deployment.yaml
@@ -110,6 +110,9 @@ spec:
           volumeMounts:
           - name: consul-bin
             mountPath: /consul-bin
+          {{- if ( default $root.Values.global.initCopyContainerResources .resources) }}
+          resources: {{ toYaml (default $root.Values.global.initCopyContainerResources.resources .resources) | nindent 12 }}
+          {{- end }}
         {{- if (and $root.Values.global.tls.enabled $root.Values.global.tls.enableAutoEncrypt) }}
         {{- include "consul.getAutoEncryptClientCA" $root | nindent 8 }}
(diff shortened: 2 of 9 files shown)


templates/ingress-gateways-deployment.yaml
```diff
@@ -110,6 +110,9 @@ spec:
           volumeMounts:
           - name: consul-bin
             mountPath: /consul-bin
+          {{- if ( default $root.Values.global.initCopyContainerResources .resources) }}
+          resources: {{ toYaml (default $root.Values.global.initCopyContainerResources.resources .resources) | nindent 12 }}
```
ashwin-venkatesh, line 114:
```suggestion
          resources: 
            {{ toYaml (default $root.Values.global.initCopyContainerResources.resources .resources) | nindent 12 }}
```

templates/ingress-gateways-deployment.yaml
```diff
@@ -260,8 +263,7 @@ spec:
         - name: ingress-gateway
           image: {{ $root.Values.global.imageEnvoy | quote }}
           {{- if (default $defaults.resources .resources) }}
-          resources:
-            {{ toYaml (default $defaults.resources .resources) | nindent 12 }}
+          resources: {{ toYaml (default $defaults.resources .resources) | nindent 12 }}
```
ashwin-venkatesh, line 266:
```suggestion
          resources: 
            {{ toYaml (default $defaults.resources .resources) | nindent 12 }}
```

templates/ingress-gateways-deployment.yaml
kschoche: Thanks for the suggestion, oddly enough that is how I had it at first, and the formatting actually results in a yaml formatted like this : 
```
  resources:
  
    requests:
      memory: "100Mi"
       cpu: "100m"
....
```
Thoughts?

ashwin-venkatesh: That is surprising. I was looking at the rest of the diff, and it looked like it was set in a newline everywhere else. 🤯

ashwin-venkatesh: Actually, i think i see it maybe? 
`{{ toYaml (default .Values.global.initCopyContainerResources .resources) | nindent 12 }}` vs `{{ toYaml (default $root.Values.global.initCopyContainerResources.resources .resources) | nindent 12 }}`

One of them has a `.resources` at the end of the `initCopyContainerResources` and the other doesn't.

templates/connect-inject-deployment.yaml
```diff
@@ -138,6 +138,34 @@ spec:
                 {{- if not (kindIs "invalid" $resources.requests.cpu) }}
                 -default-sidecar-proxy-cpu-request={{ $resources.requests.cpu }} \
                 {{- end }}
+                {{- $resources := .Values.global.initCopyContainerResources.resources }}
+                {{- /* kindIs is used here to differentiate between null and 0 */}}
```
lkysow, line 142: Since we've got that same comment above I don't think it's needed here.

values.yaml
```diff
@@ -219,6 +219,28 @@ global:
...
+      requests:
+        memory: "25Mi"
+        cpu: "20m"
+      limits:
+        memory: "25Mi"
+        cpu: "20m"
+
+  # Resource settings for copy-consul-bin init containers.
+  # these settings are bounded by the size of the consul binary
+  # as we issue a cp of it during the init container.
```
lkysow, line 234:
```suggestion
  # as we use `cp` to copy it into a shared volume in the init container.
```

Probably best to be specific.

values.yaml
```diff
@@ -219,6 +219,28 @@ global:
...
+    resources:
+      requests:
+        memory: "25Mi"
+        cpu: "20m"
+      limits:
+        memory: "25Mi"
+        cpu: "20m"
+
+  # Resource settings for copy-consul-bin init containers.
+  # these settings are bounded by the size of the consul binary
```
lkysow, line 233:
```suggestion
  # These settings are bounded by the size of the consul binary
```

values.yaml
```diff
@@ -219,6 +219,28 @@ global:
     # Requires consul-k8s 0.15.0+.
     createFederationSecret: false
 
+  # Resource settings for lifecycle-sidecar containers
```
lkysow, line 222:
```suggestion
  # Resource settings for lifecycle-sidecar containers.
```

values.yaml
```diff
@@ -219,6 +219,28 @@ global:
...
+  lifecycleSidecarContainerResources:
```
lkysow, line 223: I'm wondering if we should add another level here:
```
lifecycleSidecar:
  resources:
```

I think that will future proof us if we want to add more settings to the lifecycle sidecar in the future (like disabling it).

values.yaml
lkysow, line 222: Can you explain a bit what the lifecycle sidecar is and where it is used, i.e. it ensures Consul services are always registered with their local consul clients and it's used in mesh, ingress and terminating gateways along with every connect-injected service.

values.yaml
```diff
@@ -219,6 +219,28 @@ global:
...
+        memory: "25Mi"
+        cpu: "20m"
+      limits:
+        memory: "25Mi"
+        cpu: "20m"
+
+  # Resource settings for copy-consul-bin init containers.
+  # these settings are bounded by the size of the consul binary
+  # as we issue a cp of it during the init container.
+  initCopyContainerResources:
```
lkysow, line 235: again, we should probably nest this
```
initCopyContainer:
  resources:
```

values.yaml
lkysow: Can you also add (like with the lifecycle sidecar) where these containers are used.

templates/ingress-gateways-deployment.yaml
lkysow: The use of `default` isn't needed here. default is used in the other location to allow setting a default for the gateway container and overriding it subsequently. As written, if they set resources for the gateway container those settings will actually get used for the copy containers which isn't what we want, e.g.
```yaml
ingressGateways:
  resources:
    limits:
      memory: 1000m # this is now used for the copy, lifecycle AND the main envoy container.
```

In this case, because we don't allow users to override the copy container/lifecycle container's settings at all. We just need to reference the single global setting.

So this section should read
```yaml
          resources: {{ toYaml $root.Values.global.initCopyContainerResources.resources | nindent 12 }}
```

There's no need for the `if` either.

templates/ingress-gateways-deployment.yaml
lkysow: without kyle's change I'm getting an extra newline here

templates/mesh-gateway-deployment.yaml
```diff
@@ -90,13 +90,10 @@ spec:
...
+          {{- if ( default .Values.global.initCopyContainerResources .resources) }}
           resources:
-            requests:
-              memory: "25Mi"
-              cpu: "50m"
-            limits:
-              memory: "25Mi"
-              cpu: "50m"
+            {{ toYaml (default .Values.global.initCopyContainerResources .resources) | nindent 12 }}
+          {{- end }}
```
lkysow, line 96: I think this should be
```yaml
resources:
  {{ toYaml  .Values.global.initCopyContainerResources | nindent 12 }}
```

There's no need to use `default` here.

templates/mesh-gateway-deployment.yaml
```diff
@@ -356,13 +353,10 @@ spec:
             {{- if .Values.global.acls.manageSystemACLs }}
             - -token-file=/consul/service/acl-token
             {{- end }}
+          {{- if ( default .Values.global.lifecycleSidecarContainerResources .resources) }}
```
lkysow, line 356: again, no need for default

templates/terminating-gateways-deployment.yaml
```diff
@@ -124,6 +124,10 @@ spec:
           volumeMounts:
           - name: consul-bin
             mountPath: /consul-bin
+          {{- if ( default $root.Values.global.initCopyContainerResources .resources) }}
```
lkysow, line 127: same comments as ingress gateways

templates/terminating-gateways-deployment.yaml
```diff
@@ -124,6 +124,10 @@ spec:
...
+          resources:
+            {{ toYaml (default $root.Values.global.initCopyContainerResources .resources) | nindent 12 }}
+          {{- end }}
         {{- if (and $root.Values.global.tls.enabled $root.Values.global.tls.enableAutoEncrypt) }}
```
lkysow, line 131: You're missing the change to the lifecycle sidecar

lkysow:
Oh, we also need to add tests that the resources are getting set as expected for the gateways and for connect-inject that those flags are being passed as expected. There's already some tests there that you can look at.

PR #534, opened 2020-07-09
Remove scripts/generate-helm-config-for-hcs.sh

ishustava:
This was a temporary helper script to make it easier to install Consul clients running on Kubernetes via this Helm chart with HCS as an external Consul server. We have now implemented this behavior in the [HCS CLI](https://releases.hashicorp.com/hcs/), which is an Azure CLI plugin.


thisisnotashwin (approved):
💥

lkysow (approved):

PR #547, opened 2020-07-22
Run acceptance tests on every PR

ishustava:
This PR proposes to make the following CI changes:

- Runs acceptance tests on every PR except forks. I have followed [this](https://circleci.com/blog/managing-secrets-when-you-have-pull-requests-from-outside-contributors/) blog post to do that.
- Adds `go-fmt-and-vet` job
- Adds a job to run acceptance tests framework unit tests
- Make some changes to our test docker file. The main reason for those is because the circleci/golang image is an ubuntu rather than an alpine based image. It also already installs many base packages we've been installing, so I've removed them. I've run Helm unit tests with that image and it works. 

Notes: 
- the code changes are the result of fixing `go-fmt-and-vet` job 😄  
- `go.sum` changes are because I've switched from `kubernetes.Clientset` to `kubernetes.Interface` everywhere so we can test it more easily.


.circleci/config.yml
```diff
@@ -24,11 +31,73 @@ jobs:
...
+      - checkout
+
+      # Restore go module cache if there is one
+      - restore_cache:
+          keys:
+            - consul-helm-modcache-v1-{{ checksum "test/acceptance/go.mod" }}
+
+      - run: mkdir -p $TEST_RESULTS
+
+      # check go fmt output because it does not report non-zero when there are fmt changes
```
alvin-huang, line 83: Wrong comment for this

.circleci/config.yml
```diff
@@ -83,15 +171,17 @@ workflows:
   version: 2
   test:
     jobs:
+      - go-fmt-and-vet
+      - unit-acceptance-framework:
+          requires:
+            - go-fmt-and-vet
       - unit-helm2
```
alvin-huang, line 178: Do you care to have the fmt/vet checks pass before running the helm2/3 unit tests? ie: also require `go-fmt-and-vet` for these 2 jobs

alvin-huang (approved):
Looks great! I only looked at the CircleCI stuff and will let someone else from the team look at the Go code.

.circleci/config.yml
ishustava: We don't at the moment since they are using `bats`, so we don't need to require it (at least for now 😃 ).

.circleci/config.yml
ishustava: 🤭 good catch, thanks!

kschoche (approved):
Great work, this looks awesome!

PR #552, opened 2020-07-24
Add multi-cluster support to the acceptance tests framework, plus a few mesh gateway tests

ishustava:

* Add support for multi-cluster tests to the framework
* Extend terraform templates to support creating multiple clusters and output kubernetes context names
* Move terraform/* -> terraform/gke/*
* Add mesh gateway and wan federation test with a default installation (TLS enabled, ACLs disabled)
* Add mesh gateway and wan federation test with TLS and ACLs enabled

There is now a CI job that runs these tests 🎉  but if you'd like to run them yourself, remember to set `-p 1` because tests cannot be run in parallel. Other than that, you'd also need two Kubernetes clusters. Assuming both of your clusters live in the same kubeconfig at `~/.kube/config`, you would need the following command if running only the mesh gateway tests:

```
cd test/acceptance/tests/mesh-gateway
go test -enable-multi-cluster -kubecontext <name of the primary kubecontext> -secondary-kubecontext <name of the secondary kubecontext>
```

Note: I have only tested this with GKE, and I'm pretty sure it won't work on Kind out-of-the-box because Kind doesn't support load balancer services.


ashwin-venkatesh:
Tests seem to run flake-ly on EKS clusters. @ishustava and I paired on trying to get to the bottom of it but it was not quiet obvious what exactly had failed. Additionally, `ctrl + c` halfway through the tests appeared to not clean up some of the `Opaque` secrets. 

Let me know if you want me to update this comment with more information Iryna!!

lkysow:
I think we'll need a flag to configure the cluster come-up timeout because it doesn't work on aks.

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
```diff
@@ -0,0 +1,250 @@
...
+	primaryConsulCluster.Create(t)
+
+	secondaryContext := env.Context(t, framework.SecondaryContextName)
+
+	// Get the federation secret from the primary cluster and apply it to secondary cluster
+	federationSecretName := fmt.Sprintf("%s-consul-federation", releaseName)
+	t.Logf("retrieving federation secret %s from the primary cluster and applying to the secondary", federationSecretName)
+	secretYAML, err := helpers.RunKubectlAndGetOutputWithLoggerE(t, env.DefaultContext(t).KubectlOptions(), logger.Discard, "get", "secret", federationSecretName, "-o", "yaml")
+	require.NoError(t, err)
+	helpers.KubectlApplyFromString(t, secondaryContext.KubectlOptions(), secretYAML)
```
ashwin-venkatesh, line 50: This is more of a personal preference but could we use the k8s client to perform this operation. Using a client to manage resources feels more robust than shell-ing out. Roughly something that looks like:
```
	federationSecret, err := env.DefaultContext(t).KubernetesClient(t).CoreV1().Secrets(env.DefaultContext(t).KubectlOptions().Namespace).Get(federationSecretName, metav1.GetOptions{})
	require.NoError(t, err)
	_, err = secondaryContext.KubernetesClient(t).CoreV1().Secrets(secondaryContext.KubectlOptions().Namespace).Create(federationSecret)
	require.NoError(t, err)
```

lkysow (commented):
I tried to get this working with EKS => AKS and vice versa. I ended up running into a lot of timeout issues. These changes are an attempt to get it working: https://github.com/hashicorp/consul-helm/compare/hashicorp:acceptance-tests/multi-cluster-framework...hashicorp:acceptance-tests/multi-cluster-framework-lkysow

I don't think those changes block merging this but perhaps we can agree on which ones we'd like to add into the framework.

I also ran into a persistent issue with EKS where the client => server continually gets 503 no healthy upstreams. I'd bet this is an envoy outlier detection issue. Not sure how we could fix it.

Note that I was running with `-fastfail -timeout 20m`

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
```diff
@@ -0,0 +1,250 @@
...
+		"meshGateway.replicas": "1",
+	}
+
+	releaseName := helpers.RandomName()
+
+	// Install the primary consul cluster in the default kubernetes context
+	primaryConsulCluster := framework.NewHelmCluster(t, primaryHelmValues, env.DefaultContext(t), suite.Config(), releaseName)
+	primaryConsulCluster.Create(t)
+
+	secondaryContext := env.Context(t, framework.SecondaryContextName)
```
ashwin-venkatesh, line 43: We can extract the primary context as a var here as well so that it is easier to pass around instead of env.DefaultContext(t) 😃

thisisnotashwin (commented):
This looks really great @ishustava !! I really like the structuring and the overall layout of the tests.

I have a few suggestions/questions that are a little more meta and I'll leave them here but wouldn't treat them as blockers:

• We could possibly save the k8s client inside the kubernetes context so that we dont have to instantiate it each time
```
type kubernetesContext struct {
	pathToKubeConfig string
	contextName      string
	namespace        string
	client           kubernetes.Interface
}

func (k *kubernetesContext) KubernetesClient(t *testing.T) kubernetes.Interface {
	if k.client != nil {
		return k.client
	}
	configPath, err := k.KubectlOptions().GetConfigPath(t)
	require.NoError(t, err)

	config, err := k8s.LoadApiClientConfigE(configPath, k.contextName)
	require.NoError(t, err)

	client, err := kubernetes.NewForConfig(config)
	require.NoError(t, err)

	k.client = client

	return client
}
```

• With the use of the k8s client, it should be possible to get away with not using `kubectl` and we can potentially move the fixture files into being k8s go objects. This is a definitely trickier as `--wait` logic is more complicated to implement and possibly something to keep an eye on. My rationale here is that it is easier to track client go versions than it is `kubectl` versions as that dependency is external. Also, I am definitely biased towards using libraries over shell-ing out.

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
ishustava: Done!

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
ishustava: Yeah! I think I just codified our docs and didn't think to convert the client here 🤦‍♀️ . Thanks!

ashwin-venkatesh: This is why we ~pair~ review code 😉

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
```diff
@@ -0,0 +1,250 @@
...
+
+	t.Log("creating static-client in dc1")
+	createClient(t, env.DefaultContext(t).KubectlOptions())
+
+	t.Log("checking that connection is successful")
+	checkConnection(t, env.DefaultContext(t).KubectlOptions(), env.DefaultContext(t).KubernetesClient(t), true)
+}
+
+// Test that Connect and wan federation over mesh gateways work in a secure installation,
+// with ACLs and TLS with auto-encrypt enabled
```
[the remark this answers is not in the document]
ishustava: Added!

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
```diff
@@ -0,0 +1,250 @@
...
+		"meshGateway.enabled":  "true",
+		"meshGateway.replicas": "1",
+	}
+
+	// Install the secondary consul cluster in the secondary kubernetes context
+	secondaryConsulCluster := framework.NewHelmCluster(t, secondaryHelmValues, secondaryContext, suite.Config(), releaseName)
+	secondaryConsulCluster.Create(t)
+
+	// Verify federation between servers
+	// todo: add logs
```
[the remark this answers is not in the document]
ishustava: yep, fixed!

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
ishustava: Fixed! I had to set resourceVersion to `""` on the secret before applying it to another cluster, otherwise, kubernetes would reject the request, but I think it's ok and still better than shelling out.

ishustava:
@ashwin-venkatesh 

> We could possibly save the k8s client inside the kubernetes context so that we dont have to instantiate it each time

Yes, definitely. I've switched to the code you suggested instead.

> With the use of the k8s client, it should be possible to get away with not using kubectl and we can potentially move the fixture files into being k8s go objects. This is a definitely trickier as --wait logic is more complicated to implement and possibly something to keep an eye on. My rationale here is that it is easier to track client go versions than it is kubectl versions as that dependency is external. Also, I am definitely biased towards using libraries over shell-ing out.

I have been somewhat inconsistent in using client vs fixture files throughout the tests (sync uses the client and go objects, while connect and mesh gateways use fixtures). I think for connect I've opted out for fixtures because it seemed easier to understand/more readable because I could just look at a yaml file the test is using. It also might help with debugging if you're running commands yourself, you could also use these files to reproduce an issue. I'd be in favor of making it all consistent, but I have a slight preference for using fixtures in acceptance tests. But I can be convinced otherwise!

ishustava:
@lkysow @ashwin-venkatesh 

I've added `-no-cleanup-on-failure` flag, however, other flags/work to make it work on AKS I'd prefer to do separately. I've mentioned that I've only tested it on GKE and this was intentional, but unfortunately not very clear. This was meant to be an indication that I've not verified it works on other clouds, and so I don't expect the reviewers to test it on those clouds. I'll be more clear in my future PRs. Sorry, you had to discover all those bugs!! I'll make these changes as part of running these tests on AKS and EKS in CI (we have a separate placeholder ticket for that). For this PR though, I'd like to focus on framework support and that the tests pass on GKE.

lkysow (approved):
🎉

lkysow:
I like using the yaml fixtures for things like the static-client because it makes it easy to debug and also introspect what's being installed.

thisisnotashwin (approved):
🏅

test/acceptance/tests/connect/connect_inject_test.go
```diff
@@ -108,8 +108,8 @@ func checkConnection(t *testing.T, options *k8s.KubectlOptions, client kubernete
 			require.NoError(r, err)
 			require.Contains(r, output, "hello world")
 		} else {
-			require.Error(t, err)
-			require.Contains(t, output, "503 Service Unavailable")
+			require.Error(r, err)
```
lkysow, line 111: you're gonna give me conflicts aren't you :D

ishustava: Sorry! but I think this was making tests fail in CI, so thank you!!!

PR #553, opened 2020-07-24
Add ability to add additional labels to server pods

peterklijn:
I would like to add the ability to add additional labels to consul-server pods, the same way annotations can be added.

At the company I work for, we use specific Kubernetes pod labels to tell Fluentd which Elasticsearch index should be used to send logs to. With this change I can the additional label :)

### DIFF
--- a/test/unit/server-statefulset.bats
+++ b/test/unit/server-statefulset.bats
@@ -346,6 +346,28 @@ load _helpers
 }
 
 #--------------------------------------------------------------------
+# podLabels
+
+@test "server/StatefulSet: no pod labels defined by default" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/server-statefulset.yaml  \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.metadata.labels | del(."app") | del(."chart") | del(."release") | del(."component") | del(."hasDNS")' | tee /dev/stderr)
+  [ "${actual}" = "{}" ]
+}
+
+@test "server/StatefulSet: pod labels can be set" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/server-statefulset.yaml  \
+      --set 'server.podLabels=foo: bar' \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.metadata.labels.foo' | tee /dev/stderr)
+  [ "${actual}" = "bar" ]
+}
+
+#--------------------------------------------------------------------
 # annotations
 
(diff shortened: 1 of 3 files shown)


test/unit/server-statefulset.bats
```diff
@@ -345,6 +345,28 @@ load _helpers
...
+@test "server/StatefulSet: no pod labels defined by default" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/server-statefulset.yaml  \
+      . | tee /dev/stderr |
+      yq -r '.spec.template.metadata.labels | del(."app") | del(."chart") | del(."release") | del(."component") | del(."hasDNS")' | tee /dev/stderr)
+  [ "${actual}" = "{}" ]
+}
+
+@test "server/StatefulSet: pod labels can be set" {
```
ashwin-venkatesh, line 360: Could you add another testcase here which would verify multiple labels getting set if `podLabels` has multiple values set?

values.yaml
```diff
@@ -321,6 +321,13 @@ server:
   # ref: https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/
   priorityClassName: ""
 
+  # Extra labels to attach to the server pods.
+  # This should be a multi-line YAML string.
+  # Example:
+  #   podLabels: |
+  #     "label-key": "label-value"
+  podLabels: null
```
ashwin-venkatesh, line 329: Can this be renamed to `extraLabels`? That seems to be aligned with how we have addressed similar constructs in the values file. see: extraConfig, extraVolumes.

Additionally, can you update the example to demonstrate setting multiple labels? It should help other users when they are trying to set this up

test/unit/server-statefulset.bats
peterklijn: Sure thing :)

values.yaml
peterklijn: Done :)

peterklijn:
No problem @ashwin-venkatesh :)

I've updated the `extraLabels` to use a yaml block.

Additionally I've updated the `annotations` to also support a yaml block (it currently supports both a yaml block and a string which is documented as deprecated). Let me know if that's ok :)

lkysow:
I think we should leave out the annotations as string deprecation from this PR and put it in a separate PR that does the deprecation across all places where annotations can be set so we don't have inconsistency across the chart between components.

peterklijn:
Thanks for the review @ashwin-venkatesh and @lkysow,

I've removed the annotation change and rebased with upstream

PR #555, opened 2020-07-27
Customize the DNS service (fixes #545)

### LINKED ISSUE #545: Expose Consul DNS as internal LoadBalancer
I'm running consul on GKE in GCP and would like to expose the DNS service via an internal LoadBalancer (`cloud.google.com/load-balancer-type: "Internal"`).

Currently the the service type is harcoded to ClusterIP.

https://github.com/hashicorp/consul-helm/blob/63037756ce11d8824f81af9149e7b0691b3e66c1/templates/dns-service.yaml#L20

Would you accept a PR for that?

slauger:
This PR adds some options to enable the customization the DNS service. See #545 for more information.

### DIFF
--- a/templates/dns-service.yaml
+++ b/templates/dns-service.yaml
@@ -17,7 +17,7 @@ metadata:
   {{- end }}
 spec:
 {{- if .Values.dns.type }}
-type: {{ .Values.dns.type }}
+  type: {{ .Values.dns.type }}
 {{- end }}
 {{- if .Values.dns.clusterIP }}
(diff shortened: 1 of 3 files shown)


slauger:
Hi,

sorry, for the late response. I'm already familiar with the usage of bats and CircleCI. Tests are added. Let me know what you think. 

I also introduced `dns.additionalSpec`, just as `ui.additionalSpec`. This also obsoletes `dns.clusterIP`, but i would keep the parameter for keep backwards compatibility.

PR #559, opened 2020-07-30
Acceptance tests for ingress gateways

lkysow:
Also implements helm upgrade.

### DIFF
--- a/test/acceptance/framework/consul_cluster.go
+++ b/test/acceptance/framework/consul_cluster.go
@@ -27,7 +27,10 @@ const helmChartPath = "../../../.."
 type Cluster interface {
 	Create(t *testing.T)
 	Destroy(t *testing.T)
-	Upgrade(t *testing.T)
+	// Upgrade runs helm upgrade. It will update the helm values from the
+	// initial install with helmValues. Any keys that were previously set
+	// will be overridden by the helmValues keys.
+	Upgrade(t *testing.T, helmValues map[string]string)
 	SetupConsulClient(t *testing.T, secure bool) *api.Client
 }
 
@@ -117,7 +120,8 @@ func (h *HelmCluster) Destroy(t *testing.T) {
 	}
 }
 
-func (h *HelmCluster) Upgrade(t *testing.T) {
+func (h *HelmCluster) Upgrade(t *testing.T, helmValues map[string]string) {
+	mergeMaps(h.helmOptions.SetValues, helmValues)
 	helm.Upgrade(t, h.helmOptions, helmChartPath, h.releaseName)
 	helpers.WaitForAllPodsToBeReady(t, h.kubernetesClient, h.helmOptions.KubectlOptions.Namespace, fmt.Sprintf("release=%s", h.releaseName))
 }
--- a/test/acceptance/framework/consul_cluster.go
+++ b/test/acceptance/framework/consul_cluster.go
@@ -27,7 +27,7 @@ const helmChartPath = "../../../.."
 type Cluster interface {
 	Create(t *testing.T)
 	Destroy(t *testing.T)
-	// Upgrade runs helm upgrade. It will update the helm values from the
+	// Upgrade runs helm upgrade. It will merge the helm values from the
 	// initial install with helmValues. Any keys that were previously set
 	// will be overridden by the helmValues keys.
 	Upgrade(t *testing.T, helmValues map[string]string)
--- a/test/acceptance/framework/consul_cluster.go
+++ b/test/acceptance/framework/consul_cluster.go
@@ -190,7 +190,7 @@ func (h *HelmCluster) checkForPriorInstallations(t *testing.T) {
 	require.NoError(t, err)
 
 	for _, r := range installedReleases {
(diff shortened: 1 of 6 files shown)


lkysow:
Adding @ishustava because of the helm upgrade change

test/acceptance/framework/consul_cluster.go
```diff
@@ -27,7 +27,10 @@ const helmChartPath = "../../../.."
 type Cluster interface {
 	Create(t *testing.T)
 	Destroy(t *testing.T)
-	Upgrade(t *testing.T)
+	// Upgrade runs helm upgrade. It will update the helm values from the
```
ishustava, line 30:
```suggestion
	// Upgrade runs helm upgrade. It will merge the helm values from the
```

ishustava: more specific?

test/acceptance/tests/connect/connect_inject_test.go
```diff
@@ -108,8 +108,8 @@ func checkConnection(t *testing.T, options *k8s.KubectlOptions, client kubernete
 			require.NoError(r, err)
 			require.Contains(r, output, "hello world")
 		} else {
-			require.Error(t, err)
-			require.Contains(t, output, "503 Service Unavailable")
+			require.Error(r, err)
+			require.Contains(r, output, "503 Service Unavailable")
```
ishustava, line 112: 🙏

test/acceptance/tests/ingress-gateway/ingress_gateway_test.go
```diff
@@ -0,0 +1,192 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// Test that ingress gateways work in a default installation.
+func TestIngressGatewayDefault(t *testing.T) {
+	env := suite.Environment()
+
+	// Initially we start without ingress gateways enabled because we need
+	// to create their config entry first.
```
ishustava, line 22: Is this a requirement? I feel like I was able to do that with a single install, but maybe I'm missing something. Also, I don't think we mention it in [our docs](https://www.consul.io/docs/k8s/connect/ingress-gateways).

test/acceptance/tests/ingress-gateway/ingress_gateway_test.go
```diff
@@ -0,0 +1,192 @@
...
+		"ingressGateways.enabled":              "1",
+		"ingressGateways.gateways[0].name":     "ingress-gateway",
+		"ingressGateways.gateways[0].replicas": "1",
+	})
+
+	// With the ingress gateway up, we test that we can make a call to it
+	// via the bounce pod. It should route to the static-server pod.
+	t.Log("trying calls to ingress gateway")
+	k8sClient := env.DefaultContext(t).KubernetesClient(t)
+	k8sOptions := env.DefaultContext(t).KubectlOptions()
```
ishustava, line 75: These aren't used anywhere else? are you extracting them into variables for brevity/readibility?

ishustava (commented):
Looks great! I've added a couple of questions. I don't remember needing to have a config entry before enabling the ingress gateway in the helm chart.

test/acceptance/tests/ingress-gateway/ingress_gateway_test.go
```diff
@@ -0,0 +1,192 @@
...
+			},
+		},
+	}, nil)
+	require.NoError(t, err)
+	require.Equal(t, true, created, "config entry failed")
+
+	// Now we can upgrade the cluster and enable ingress gateways.
+	t.Log("upgrading helm release with ingress gateways enabled")
+	consulCluster.Upgrade(t, map[string]string{
+		"ingressGateways.enabled":              "1",
```
ishustava, line 66:
```suggestion
		"ingressGateways.enabled":              "true",
```

test/acceptance/tests/ingress-gateway/ingress_gateway_test.go
lkysow: Fixed to remove the upgrade. I left in the code for upgrade because it might still be useful.

test/acceptance/tests/ingress-gateway/ingress_gateway_test.go
lkysow: I think it was copy pasta but now that I've merged the defaults and secure tests it's needed since we call checkConnection twice now.

ishustava (approved):
Looks great!!

test/acceptance/tests/ingress-gateway/ingress_gateway_test.go
ishustava: makes sense.

thisisnotashwin (approved):
Love the approach of using the bounce pod to curl into the mesh!! 🥇

lkysow:
> Love the approach of using the bounce pod to curl into the mesh!! 🥇

Haha thanks. This is mostly so that the test works without needing to expose a load balancer which means it will also work with kind.

PR #561, opened 2020-07-31
Make sure GKE clusters get cleaned up from the right directory in CI

ishustava:


kschoche (approved):
Looks great, good fix :)

adilyse (approved):
Thanks for updating this!

PR #568, opened 2020-08-07
Test Consul DNS

thisisnotashwin:
Consul DNS is tested by creating a pod that runs dig against the DNS service. ~As its behavior is independent of security policies, it has been tested only in the insecure installation~.

### DIFF
--- a/test/acceptance/tests/consul-dns/consul_dns_test.go
+++ b/test/acceptance/tests/consul-dns/consul_dns_test.go
@@ -3,77 +3,91 @@ package consuldns
 import (
 	"fmt"
 	"testing"
-	"time"
 
 	"github.com/hashicorp/consul-helm/test/acceptance/framework"
 	"github.com/hashicorp/consul-helm/test/acceptance/helpers"
-	"github.com/hashicorp/consul/sdk/testutil/retry"
 	"github.com/stretchr/testify/require"
-	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const podName = "dns-pod"
 
 func TestConsulDNS(t *testing.T) {
-	env := suite.Environment()
-	context := env.DefaultContext(t)
-	releaseName := helpers.RandomName()
-
-	cluster := framework.NewHelmCluster(t, nil, context, suite.Config(), releaseName)
-	cluster.Create(t)
-
-	k8sClient := context.KubernetesClient(t)
-	contextNamespace := context.KubectlOptions().Namespace
-
-	dnsService, err := k8sClient.CoreV1().Services(contextNamespace).Get(fmt.Sprintf("%s-%s", releaseName, "consul-dns"), metav1.GetOptions{})
-	require.NoError(t, err)
-
-	dnsIP := dnsService.Spec.ClusterIP
-
-	createdDNSPod(t, releaseName, context)
-
-	testDNS(t, context, dnsIP)
-}
-
-func createdDNSPod(t *testing.T, releaseName string, context framework.TestContext) {
-	dnsPod := corev1.Pod{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: podName,
+	cases := []struct {
+		name       string
+		helmValues map[string]string
+	}{
+		{
+			"Default installation",
+			nil,
 		},
-		Spec: corev1.PodSpec{
-			Containers: []corev1.Container{
-				{
-					Name:    "dns",
-					Image:   "anubhavmishra/tiny-tools",
-					Command: []string{"dig", fmt.Sprintf("@%s-consul-dns", releaseName), "consul.service.consul"},
-				},
+		{
+			"Secure installation (with TLS and ACLs enabled)",
+			map[string]string{
+				"global.tls.enabled":           "true",
+				"global.acls.manageSystemACLs": "true",
 			},
-			RestartPolicy: corev1.RestartPolicyNever,
 		},
 	}
-	_, err := context.KubernetesClient(t).CoreV1().Pods(context.KubectlOptions().Namespace).Create(&dnsPod)
-	require.NoError(t, err)
 
-	helpers.Cleanup(t, suite.Config().NoCleanupOnFailure, func() {
-		// Note: this delete command won't wait for pods to be fully terminated.
-		// This shouldn't cause any test pollution because the underlying
-		// objects are deployments, and so when other tests create these
-		// they should have different pod names.
-		helpers.RunKubectl(t, context.KubectlOptions(), "delete", "pod", "dns-pod")
-	})
-}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			env := suite.Environment()
+			context := env.DefaultContext(t)
+			releaseName := helpers.RandomName()
 
-func testDNS(t *testing.T, context framework.TestContext, dnsIP string) {
-	retrier := &retry.Timer{
-		Timeout: 20 * time.Second,
-		Wait:    500 * time.Millisecond,
+			cluster := framework.NewHelmCluster(t, c.helmValues, context, suite.Config(), releaseName)
+			cluster.Create(t)
+
+			k8sClient := context.KubernetesClient(t)
+			contextNamespace := context.KubectlOptions().Namespace
+
+			dnsService, err := k8sClient.CoreV1().Services(contextNamespace).Get(fmt.Sprintf("%s-%s", releaseName, "consul-dns"), metav1.GetOptions{})
+			require.NoError(t, err)
+
+			dnsIP := dnsService.Spec.ClusterIP
+
+			consulServerList, err := k8sClient.CoreV1().Pods(contextNamespace).List(metav1.ListOptions{
+				LabelSelector: "app=consul,component=server",
+			})
+			require.NoError(t, err)
+
+			serverIPs := make([]string, len(consulServerList.Items))
+			for _, serverPod := range consulServerList.Items {
+				serverIPs = append(serverIPs, serverPod.Status.PodIP)
+			}
+
+			dnsTestPodArgs := []string{
+				"run", "-it", podName, "--restart", "Never", "--image", "anubhavmishra/tiny-tools", "--", "dig", fmt.Sprintf("@%s-consul-dns", releaseName), "consul.service.consul",
+			}
+			logs, err := helpers.RunKubectlAndGetOutputE(t, context.KubectlOptions(), dnsTestPodArgs...)
+			require.NoError(t, err)
+
+			helpers.Cleanup(t, suite.Config().NoCleanupOnFailure, func() {
+				// Note: this delete command won't wait for pods to be fully terminated.
+				// This shouldn't cause any test pollution because the underlying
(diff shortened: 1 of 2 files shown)


test/acceptance/tests/consul-dns/consul_dns_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/hashicorp/consul-helm/test/acceptance/framework"
+	"github.com/hashicorp/consul-helm/test/acceptance/helpers"
+	"github.com/hashicorp/consul/sdk/testutil/retry"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const PODNAME = "dns-pod"
```
ishustava, line 16: Let's use camelcase for constants. I think this is the convention in golang.
```suggestion
const podName = "dns-pod"
```

ishustava (approved):
Looks great! A couple of minor comments, but nothing blocking. Great work!

test/acceptance/tests/consul-dns/consul_dns_test.go
```diff
@@ -0,0 +1,79 @@
...
+	helpers.Cleanup(t, suite.Config().NoCleanupOnFailure, func() {
+		// Note: this delete command won't wait for pods to be fully terminated.
+		// This shouldn't cause any test pollution because the underlying
+		// objects are deployments, and so when other tests create these
+		// they should have different pod names.
+		helpers.RunKubectl(t, context.KubectlOptions(), "delete", "pod", "dns-pod")
+	})
+}
+
+func testDNS(t *testing.T, context framework.TestContext, dnsIP string) {
```
lkysow, line 67: I think I'd like to see this inside the `TestConsulDNS` function because this isn't a helper but the actual meat of the test.

test/acceptance/tests/consul-dns/consul_dns_test.go
```diff
@@ -0,0 +1,79 @@
...
+	retrier := &retry.Timer{
+		Timeout: 20 * time.Second,
+		Wait:    500 * time.Millisecond,
+	}
+	retry.RunWith(retrier, t, func(r *retry.R) {
+		logs, err := helpers.RunKubectlAndGetOutputE(t, context.KubectlOptions(), "logs", podName)
+		require.NoError(r, err)
+
+		require.Contains(r, logs, fmt.Sprintf("SERVER: %s", dnsIP))
+		require.Contains(r, logs, "ANSWER SECTION:")
```
lkysow, line 77: Should we verify that this section contains the server pod ips?

test/acceptance/tests/consul-dns/consul_dns_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	k8sClient := context.KubernetesClient(t)
+	contextNamespace := context.KubectlOptions().Namespace
+
+	dnsService, err := k8sClient.CoreV1().Services(contextNamespace).Get(fmt.Sprintf("%s-%s", releaseName, "consul-dns"), metav1.GetOptions{})
+	require.NoError(t, err)
+
+	dnsIP := dnsService.Spec.ClusterIP
+
+	createdDNSPod(t, releaseName, context)
```
lkysow, line 34:
```suggestion
	createDNSTestPod(t, releaseName, context)
```

test/acceptance/tests/consul-dns/consul_dns_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	testDNS(t, context, dnsIP)
+}
+
+func createdDNSPod(t *testing.T, releaseName string, context framework.TestContext) {
```
lkysow, line 39:
```suggestion
func createDNSestPod(t *testing.T, releaseName string, context framework.TestContext) {
```

lkysow, line 39:
```suggestion
func createdDNSPod(t *testing.T, context framework.TestContext, releaseName string) {
```

Looks like we pass context first in `framework.NewHelmCluster()` so we could be consistent with that here. I haven't looked elsewhere so maybe this isn't always the pattern.

test/acceptance/tests/consul-dns/consul_dns_test.go
```diff
@@ -0,0 +1,79 @@
...
+				{
+					Name:    "dns",
+					Image:   "anubhavmishra/tiny-tools",
+					Command: []string{"dig", fmt.Sprintf("@%s-consul-dns", releaseName), "consul.service.consul"},
+				},
+			},
+			RestartPolicy: corev1.RestartPolicyNever,
+		},
+	}
+	_, err := context.KubernetesClient(t).CoreV1().Pods(context.KubectlOptions().Namespace).Create(&dnsPod)
```
lkysow, line 55: For our other "test" containers like the bounce pod we'd used a YAML fixture. This one seems like the same use-case so I think we should be consistent. I think I like the YAML because it's easier to debug the tests ourselves.

test/acceptance/tests/consul-dns/consul_dns_test.go
```diff
@@ -0,0 +1,79 @@
...
+	require.NoError(t, err)
+
+	helpers.Cleanup(t, suite.Config().NoCleanupOnFailure, func() {
+		// Note: this delete command won't wait for pods to be fully terminated.
+		// This shouldn't cause any test pollution because the underlying
+		// objects are deployments, and so when other tests create these
+		// they should have different pod names.
+		helpers.RunKubectl(t, context.KubectlOptions(), "delete", "pod", "dns-pod")
```
lkysow, line 63:
```suggestion
		helpers.RunKubectl(t, context.KubectlOptions(), "delete", "pod", podName)
```

test/acceptance/tests/consul-dns/consul_dns_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/hashicorp/consul-helm/test/acceptance/helpers"
+	"github.com/hashicorp/consul/sdk/testutil/retry"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const podName = "dns-pod"
+
+func TestConsulDNS(t *testing.T) {
```
lkysow, line 18: What about testing in secure mode? With ACLs enabled we actually need to create an anonymous policy to ensure DNS works so I do think it's worth testing.

lkysow (commented):
I love how little code these are ending up needing. Nice work on your first test. I've got a couple comments but defer to @ishustava if she thinks they don't make sense.

test/acceptance/tests/consul-dns/consul_dns_test.go
ashwin-venkatesh: My initial intention was to do that as well, but the name of the `consul-dns` service is dependent on the release name. See L49. I could potentially create a YAML template for the pod which I could inject the release name into, but it felt like overkill and also isn't something that can be kubectl apply-ied. Is there a work-around?

test/acceptance/tests/consul-dns/consul_dns_test.go
ashwin-venkatesh: Aah. I was unaware of the anonymous policy that would be needed to do the same. I'll add a test to test the secure mode installation in the case. Good catch.

ishustava: That's a good point. I completely forgot about the anonymous policy and recommended that Ashwin doesn't test in a secure mode, thinking only of TLS 🤦‍♀️ . @ashwin-venkatesh would you be OK with adding a test for a secure install? I think you could just convert this test to a table test, providing extra helm values for secure config, but keeping all assertions the same. You could take a look at the [basic test](https://github.com/hashicorp/consul-helm/blob/2cdfe72da3c1a0519546d0aa5dddc6c3bb97fa13/test/acceptance/tests/basic/basic_test.go) for reference; I think this one should be pretty similar.

test/acceptance/tests/consul-dns/consul_dns_test.go
ishustava: I just had another idea of how to simplify this such that we don't need the fixture at all.

You could create a pod and get the output of `dig` all in one command:

```shell
kubectl run  -it test-dns --image anubhavmishra/tiny-tools -- dig @iryna-consul-dns consul.service.consul
```

In code it will look like:

```go
runTestPodArgs := []string{
  "run", "-it", podName, "--image", "anubhavmishra/tiny-tools", "--", "dig", fmt.Sprintf("@%s-consul-dns", releaseName), "consul.service.consul",
}
output, err := helpers.RunKubectlAndGetOutputE(t, context.KubectlOptions(), runTestPodArgs...)
require.NoError(t, err)
// cleanup
// assertions
```

test/acceptance/tests/consul-dns/consul_dns_test.go
ashwin-venkatesh: :) done!!

ishustava (approved):
Looks great! thanks for making the changes!

test/acceptance/tests/consul-dns/consul_dns_test.go
```diff
@@ -0,0 +1,81 @@
...
+				// This shouldn't cause any test pollution because the underlying
+				// objects are deployments, and so when other tests create these
+				// they should have different pod names.
+				helpers.RunKubectl(t, context.KubectlOptions(), "delete", "pod", podName)
+			})
+
+			require.Contains(t, logs, fmt.Sprintf("SERVER: %s", dnsIP))
+			require.Contains(t, logs, "ANSWER SECTION:")
+			for _, ip := range serverIPs {
+				require.Contains(t, logs, fmt.Sprintf("consul.service.consul.\t0\tIN\tA\t%s", ip))
```
lkysow, line 77: Can you add a comment above here that shows what this output should look like so this string makes more sense.

lkysow (commented):
So I realized that right now we're testing that Consul server DNS works when querying it directly. Does it make more sense to test our documented DNS integration: https://www.consul.io/docs/k8s/dns where we've configured kube DNS to forward to Consul?

ishustava:
> So I realized that right now we're testing that Consul server DNS works when querying it directly. Does it make more sense to test our documented DNS integration: https://www.consul.io/docs/k8s/dns where we've configured kube DNS to forward to Consul?

I thought it makes sense to test it directly because I don't see that it'll provide a lot of extra confidence/value to also configure kubernetes DNS. Plus, DNS config could vary from distro to distro, so it seemed easier to test directly without the loss of generality. Do you think otherwise?

lkysow (approved):

lkysow:
> > So I realized that right now we're testing that Consul server DNS works when querying it directly. Does it make more sense to test our documented DNS integration: https://www.consul.io/docs/k8s/dns where we've configured kube DNS to forward to Consul?
> 
> I thought it makes sense to test it directly because I don't see that it'll provide a lot of extra confidence/value to also configure kubernetes DNS. Plus, DNS config could vary from distro to distro, so it seemed easier to test directly without the loss of generality. Do you think otherwise?

Sounds good.

PR #571, opened 2020-08-12
Add terminating gateway tests without ACL/TLS

kschoche:
This PR adds base terminating gateways acceptance tests without ACL/TLS support, the test creates a term gw, registers the services and deploys an application which we test for connectivity.

It largely follows the workflow described in the terminating gateway docs : https://www.consul.io/docs/k8s/connect/terminating-gateways

### DIFF
--- a/test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
+++ b/test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
@@ -33,15 +33,17 @@ func TestTerminatingGateway(t *testing.T) {
 
 	// Register the external service
 	t.Log("registering the external service")
-	err := consulClient.Agent().ServiceRegister(&api.AgentServiceRegistration{
-		Address:           "example.com",
-		Meta:              map[string]string{"external-node": "true", "external-probe": "true"},
-		ID:                "example-http",
-		Name:              "example-http",
-		Port:              80,
-		TaggedAddresses:   nil,
-		EnableTagOverride: false,
-	})
+	_, err := consulClient.Catalog().Register(&api.CatalogRegistration{
+		Node: "legacy_node",
+		//ID:       "example-http",
+		Address:  "example.com",
+		NodeMeta: map[string]string{"external-node": "true", "external-probe": "true"},
+		Service: &api.AgentService{
+			ID:      "example-http",
+			Service: "example-http",
+			Port:    80,
+		},
+	}, &api.WriteOptions{})
 	require.NoError(t, err)
 
 	// Create the config entry for the terminating gateway
--- a/test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
+++ b/test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
@@ -44,6 +44,7 @@ func TestTerminatingGateway(t *testing.T) {
 			Port:              80,
 			TaggedAddresses:   nil,
 			EnableTagOverride: false,
+		})
 		require.NoError(t, err)
 
 		// Create the config entry for the terminating gateway
--- a/test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
+++ b/test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
@@ -54,7 +54,7 @@ func TestTerminatingGateway(t *testing.T) {
 			Services: []api.LinkedService{{Name: "example-http"}},
 		}, nil)
 		require.NoError(t, err)
-		require.Equal(t, true, created, "config entry failed")
+		require.True(t, created, "config entry failed")
 
 		k8sClient := env.DefaultContext(t).KubernetesClient(t)
 		k8sOptions := env.DefaultContext(t).KubectlOptions()
--- a/test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
+++ b/test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
@@ -36,7 +36,7 @@ func TestTerminatingGateway(t *testing.T) {
 
 		// Register the external service
 		t.Log("registering the service")
-		if err := consulClient.Agent().ServiceRegister(&api.AgentServiceRegistration{
+		err := consulClient.Agent().ServiceRegister(&api.AgentServiceRegistration{
 			Address:           "example.com",
 			Meta:              map[string]string{"external-node": "true", "external-probe": "true"},
 			ID:                "example",
@@ -44,9 +44,7 @@ func TestTerminatingGateway(t *testing.T) {
 			Port:              80,
 			TaggedAddresses:   nil,
 			EnableTagOverride: false,
-		}); err != nil {
-			require.NoError(t, err)
-		}
+		require.NoError(t, err)
 
 		// Create the config entry for the terminating gateway
 		t.Log("creating config entry")
--- a/test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
+++ b/test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
@@ -26,14 +26,11 @@ func TestTerminatingGateway(t *testing.T) {
 			"terminatingGateways.gateways[0].replicas": "1",
 		}
 
+		t.Log("creating consul cluster")
 		releaseName := helpers.RandomName()
 		consulCluster := framework.NewHelmCluster(t, helmValues, env.DefaultContext(t), suite.Config(), releaseName)
-
 		consulCluster.Create(t)
 
-		/*t.Log("creating server")
-		createServer(t, suite.Config(), env.DefaultContext(t).KubectlOptions())
-		*/
 		// Once the cluster is up register the external service, then create the config entry.
 		consulClient := consulCluster.SetupConsulClient(t, false)
 
@@ -42,7 +39,8 @@ func TestTerminatingGateway(t *testing.T) {
 		if err := consulClient.Agent().ServiceRegister(&api.AgentServiceRegistration{
 			Address:           "example.com",
 			Meta:              map[string]string{"external-node": "true", "external-probe": "true"},
-			ID:                "example-http",
+			ID:                "example",
+			Name:              "example-http",
 			Port:              80,
 			TaggedAddresses:   nil,
 			EnableTagOverride: false,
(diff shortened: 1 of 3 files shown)


test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/hashicorp/consul/sdk/testutil/retry"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// Test that terminating gateways work in a default installation.
+func TestTerminatingGateway(t *testing.T) {
+	testName := "Terminating GW non-acl"
+	t.Run(testName, func(t *testing.T) {
```
lkysow, line 19: Are you running this in a `t.Run` to make way for the secure permutation?

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -0,0 +1,94 @@
...
+			Address:           "example.com",
+			Meta:              map[string]string{"external-node": "true", "external-probe": "true"},
+			ID:                "example",
+			Name:              "example-http",
+			Port:              80,
+			TaggedAddresses:   nil,
+			EnableTagOverride: false,
+		}); err != nil {
+			require.NoError(t, err)
+		}
```
lkysow, line 49:
```suggestion
		err := consulClient.Agent().ServiceRegister(&api.AgentServiceRegistration{
			Address:           "example.com",
			Meta:              map[string]string{"external-node": "true", "external-probe": "true"},
			ID:                "example",
			Name:              "example-http",
			Port:              80,
			TaggedAddresses:   nil,
			EnableTagOverride: false,
		require.NoError(t, err)
```

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+		// Create the config entry for the terminating gateway
+		t.Log("creating config entry")
+		created, _, err := consulClient.ConfigEntries().Set(&api.TerminatingGatewayConfigEntry{
+			Kind:     api.TerminatingGateway,
+			Name:     "terminating-gateway",
+			Services: []api.LinkedService{{Name: "example-http"}},
+		}, nil)
+		require.NoError(t, err)
+		require.Equal(t, true, created, "config entry failed")
```
lkysow, line 59:
```suggestion
		require.True(t, created, "config entry failed")
```

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -0,0 +1,94 @@
...
+		consulCluster := framework.NewHelmCluster(t, helmValues, env.DefaultContext(t), suite.Config(), releaseName)
+		consulCluster.Create(t)
+
+		// Once the cluster is up register the external service, then create the config entry.
+		consulClient := consulCluster.SetupConsulClient(t, false)
+
+		// Register the external service
+		t.Log("registering the service")
+		if err := consulClient.Agent().ServiceRegister(&api.AgentServiceRegistration{
+			Address:           "example.com",
```
lkysow, line 40: Should we instead create a static-server pod (non connect injected) and a kube service and use that instead? Wondering if this will be flaky due to depending on an external website (example.com).

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -0,0 +1,94 @@
...
+		deployStaticClient(t, suite.Config(), env.DefaultContext(t).KubectlOptions())
+
+		// Test that we can make a call to the terminating gateway
+		t.Log("trying calls to terminating gateway")
+		checkConnection(t, k8sOptions, k8sClient)
+	})
+}
+
+// checkConnection checks if static-client can connect through the terminating gateway.
+func checkConnection(t *testing.T, options *k8s.KubectlOptions, client kubernetes.Interface) {
```
lkysow, line 75: @ishustava should we make this a helper? I see the same method in every test and they're diverging. Some use the curl options `-vvvs`, others `-vvvsSf`, `-vvvsSs`.

lkysow (commented):
Couple questions but otherwise looks great!

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
kschoche: Yup!  That was the intention, it gets a bit ugly because the workflow for secure is very different so maybe it warrants a second test instead?

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -0,0 +1,94 @@
...
+		releaseName := helpers.RandomName()
+		consulCluster := framework.NewHelmCluster(t, helmValues, env.DefaultContext(t), suite.Config(), releaseName)
+		consulCluster.Create(t)
+
+		// Once the cluster is up register the external service, then create the config entry.
+		consulClient := consulCluster.SetupConsulClient(t, false)
+
+		// Register the external service
+		t.Log("registering the service")
+		if err := consulClient.Agent().ServiceRegister(&api.AgentServiceRegistration{
```
ishustava: Should we use the `Catalog` api as we do in the docs?

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -0,0 +1,92 @@
...
+		releaseName := helpers.RandomName()
+		consulCluster := framework.NewHelmCluster(t, helmValues, env.DefaultContext(t), suite.Config(), releaseName)
+		consulCluster.Create(t)
+
+		// Once the cluster is up register the external service, then create the config entry.
+		consulClient := consulCluster.SetupConsulClient(t, false)
+
+		// Register the external service
+		t.Log("registering the service")
+		err := consulClient.Agent().ServiceRegister(&api.AgentServiceRegistration{
```
ishustava, line 39: Should we use the `Catalog` api here as we do in the docs?

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -0,0 +1,92 @@
...
+			Address:           "example.com",
+			Meta:              map[string]string{"external-node": "true", "external-probe": "true"},
+			ID:                "example",
```
ishustava, line 42: Let's make the ID here the same as the name. That way it's more resilient for running these tests in parallel in the future.

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -0,0 +1,92 @@
...
+		t.Log("creating consul cluster")
+		releaseName := helpers.RandomName()
+		consulCluster := framework.NewHelmCluster(t, helmValues, env.DefaultContext(t), suite.Config(), releaseName)
+		consulCluster.Create(t)
+
+		// Once the cluster is up register the external service, then create the config entry.
+		consulClient := consulCluster.SetupConsulClient(t, false)
+
+		// Register the external service
+		t.Log("registering the service")
```
ishustava, line 38:
```suggestion
		t.Log("registering the external service")
```

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
ishustava: yeah, I think we should! I'm Ok with making this refactor outside of this PR though.

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -0,0 +1,92 @@
...
+		t.Log("deploying static client")
+		deployStaticClient(t, suite.Config(), env.DefaultContext(t).KubectlOptions())
+
+		// Test that we can make a call to the terminating gateway
+		t.Log("trying calls to terminating gateway")
+		checkConnection(t, k8sOptions, k8sClient)
+	})
+}
+
+// checkConnection checks if static-client can connect through the terminating gateway.
```
ishustava, line 73:
```suggestion
// checkConnection checks if static-client can connect through the external service.
```

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -0,0 +1,92 @@
...
+		// Register the external service
+		t.Log("registering the service")
+		err := consulClient.Agent().ServiceRegister(&api.AgentServiceRegistration{
+			Address:           "example.com",
+			Meta:              map[string]string{"external-node": "true", "external-probe": "true"},
+			ID:                "example",
+			Name:              "example-http",
+			Port:              80,
+			TaggedAddresses:   nil,
+			EnableTagOverride: false,
```
kschoche, line 46:
```suggestion
			EnableTagOverride: false,
		})
```

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
lkysow: Yeah not sure, I think if it might not be in this same flow then you could leave this out.

ishustava (approved):
Looks good! Great work! I left a couple of comments, but nothing blocking.

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
ishustava: I agree it's probably best to leave it out if it's not required for this test.

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
lkysow: @ishustava do you have an opinion on this?

ishustava: 🤔 Looking at the docs, it seems that the primary purpose of this is documentation. This means that we'd be abusing this in our tests, even if it's not going to cause flakiness. I do like that we're using an actual external-to-k8s service rather than a static server. Is there a public service that would be ok to use for this purpose? If not, then I think using a static-server service is better.

lkysow (approved):

PR #578, opened 2020-08-19
Acceptance tests: Refactor common connect functions to helpers

ishustava:
* Only run mesh gateway tests when `-enable-multi-cluster` is set
* Use helpers.Deploy for deploying deployments from fixtures
* Use helpers.CheckConnection for checking connections for connect
* Mark helper methods with t.Helper()
* Use static-server in the terminating gateway test instead of the example.com website

### DIFF
--- a/test/acceptance/helpers/helpers.go
+++ b/test/acceptance/helpers/helpers.go
@@ -88,11 +88,13 @@ func Deploy(t *testing.T, options *k8s.KubectlOptions, noCleanupOnFailure bool, 
 	RunKubectl(t, options, "wait", "--for=condition=available", fmt.Sprintf("deploy/%s", deployment.Name))
 }
 
-// checkConnection execs into a pod of the deployment given by deploymentName
+// CheckStaticServerConnection execs into a pod of the deployment given by deploymentName
 // and runs a curl command with the provided curlArgs.
+// This function assumes that the connection is made to the static-server and expects the output
+// to be "hello world" in a case of success.
 // If expectSuccess is true, it will expect connection to succeed,
 // otherwise it will expect failure due to intentions.
-func CheckConnection(t *testing.T, options *k8s.KubectlOptions, deploymentName string, expectSuccess bool, curlArgs ...string) {
+func CheckStaticServerConnection(t *testing.T, options *k8s.KubectlOptions, deploymentName string, expectSuccess bool, curlArgs ...string) {
 	t.Helper()
 
 	retrier := &retry.Timer{Timeout: 20 * time.Second, Wait: 500 * time.Millisecond}
--- a/test/acceptance/helpers/helpers.go
+++ b/test/acceptance/helpers/helpers.go
@@ -9,10 +9,13 @@ import (
 	"testing"
 	"time"
 
+	"github.com/gruntwork-io/terratest/modules/k8s"
 	"github.com/gruntwork-io/terratest/modules/random"
 	"github.com/hashicorp/consul/sdk/testutil/retry"
 	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/kubernetes"
 )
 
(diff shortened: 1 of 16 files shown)


test/acceptance/tests/ingress-gateway/ingress_gateway_test.go
```diff
@@ -1,25 +1,22 @@
-package connect
+package ingressgateway
```
lkysow, line 1: 🍺

test/acceptance/helpers/helpers.go
```diff
@@ -58,6 +63,55 @@ func WaitForAllPodsToBeReady(t *testing.T, client kubernetes.Interface, namespac
...
+	retrier := &retry.Timer{Timeout: 20 * time.Second, Wait: 500 * time.Millisecond}
+
+	args := []string{"exec", "deploy/" + deploymentName, "-c", deploymentName, "--", "curl", "-vvvsSf"}
+	args = append(args, curlArgs...)
+
+	retry.RunWith(retrier, t, func(r *retry.R) {
+		output, err := RunKubectlAndGetOutputE(t, options, args...)
+		if expectSuccess {
+			require.NoError(r, err)
+			require.Contains(r, output, "hello world")
```
kschoche, line 107: Great work pulling this one in!
Can we modify this to pass in some form of expected output?
"hello world" string is not going to always be in every test, I know that in the deployment used by terminating gateways it looks for a different string.

Also I wonder if the output for failure cases will vary based on the type of test?

test/acceptance/helpers/helpers.go
```diff
@@ -58,6 +63,55 @@ func WaitForAllPodsToBeReady(t *testing.T, client kubernetes.Interface, namespac
...
+	})
+
+	RunKubectl(t, options, "wait", "--for=condition=available", fmt.Sprintf("deploy/%s", deployment.Name))
+}
+
+// checkConnection execs into a pod of the deployment given by deploymentName
+// and runs a curl command with the provided curlArgs.
+// If expectSuccess is true, it will expect connection to succeed,
+// otherwise it will expect failure due to intentions.
+func CheckConnection(t *testing.T, options *k8s.KubectlOptions, deploymentName string, expectSuccess bool, curlArgs ...string) {
```
lkysow, line 95: Can we mention here that we're asserting on "hello world" output?

lkysow (approved):

test/acceptance/helpers/helpers.go
lkysow: We could also call this `CheckStaticServerConnection`. Then it's clear that we're talking to the static-server and its safe to assume its output.

test/acceptance/tests/ingress-gateway/ingress_gateway_test.go
```diff
@@ -1,25 +1,22 @@
...
-	"k8s.io/client-go/kubernetes"
 )
 
 // Test that ingress gateways work in a default installation and a secure installation.
 func TestIngressGateway(t *testing.T) {
 	for _, secure := range []bool{false, true} {
 		testName := fmt.Sprintf("secure: %t", secure)
 		t.Run(testName, func(t *testing.T) {
-			env := suite.Environment()
+			ctx := suite.Environment().DefaultContext(t)
```
kschoche, line 18: this change is much easier on the eyes! 💯

test/acceptance/helpers/helpers.go
ishustava: Yeah, let me change the name. So far, all tests are using the static server and so the output is the same. We can make this more generic later.

test/acceptance/helpers/helpers.go
ishustava: Done!

test/acceptance/helpers/helpers.go
ishustava: Renamed! Good catch!

kschoche (approved):
Everything looks great, I really like the cleanup you've done here.

Just a couple non-blocking comments:
1. I wonder if we have enough duplication of yaml files to necessitate common fixtures location?
2. a couple comments left in the code regarding output parsing in checkConnection, I do see now that you switched termgw test to use static-server instead so it should still work, but as a helper function maybe it would be nice to be able to specify expectations.
👍🏽

ishustava:
> I wonder if we have enough duplication of yaml files to necessitate common fixtures location?

That's a great point! I'll look into it but won't do it in this PR. I'm thinking that maybe we can re-use one static-server fixture everywhere but find a way to dynamically change annotations based on a test. That seems to be the main difference between fixtures. It's probably the same for the static-client fixture too.

> a couple comments left in the code regarding output parsing in checkConnection, I do see now that you switched termgw test to use static-server instead so it should still work, but as a helper function maybe it would be nice to be able to specify expectations.

Added a comment to the function, and hoping that the rename Luke suggested will help with that.

PR #580, opened 2020-08-19
Add support for changing the Consul node name where services are sync'd

adilyse:


values.yaml
```diff
@@ -730,6 +730,14 @@ syncCatalog:
   # (Kubernetes -> Consul sync)
   k8sTag: null
 
+  # consulNodeName defines the Consul synthetic node that all services
+  # will be registered to.
+  # NOTE: Changing the node name and upgrading the Helm chart will leave
+  # all of the previously sync'd services registered with Consul and
+  # register them again under the new Consul node name. The out-of-date
+  # registrations will need to be explicitly removed.
```
kschoche, line 738:
```suggestion
  # registrations will need to be manually removed.
```

kschoche (approved):
This looks awesome, great work!
One minor suggestion that is not blocking.

values.yaml
adilyse: I'd originally phrased it this way, then wondered if it would stop folks from creating scripts to automate the cleanup process when necessary.

kschoche: That makes complete sense actually! 👍🏽

lkysow (approved):
Looks good! I tested it out with ACLs and the new k8s docker image.

PR #581, opened 2020-08-19
Add auto-encrypt and secure terminating gateway tests

ishustava:


test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -11,61 +12,148 @@ import (
...
+			t.Log("deploying static client")
+			helpers.Deploy(t, ctx.KubectlOptions(), cfg.NoCleanupOnFailure, "fixtures/static-client.yaml")
+
+			// If ACLs are enabled, test that intentions prevent connections.
+			if c.secure {
+				// With the terminating gateway up, we test that we can make a call to it
+				// via the static-server. It should fail to connect with the
+				// static-server pod because of intentions.
+				t.Log("testing intentions prevent connections through the terminating gateway")
+				helpers.CheckStaticServerConnection(t,
```
ashwin-venkatesh, line 130: I know this method isn't a part of this PR, but can we move the expectedResult value to the be the last argument (source, expectedResult,destination) -> (source, destination, expectedResult). IMO it reads a little better as an assertion. I wouldnt mind (expectedResult, source, destination) either.

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -11,61 +12,148 @@ import (
...
+					ctx.KubectlOptions(),
+					"static-client",
+					false,
+					"http://localhost:1234")
+
+				// Now we create the allow intention.
+				t.Log("creating static-client => static-server intention")
+				_, _, err = consulClient.Connect().IntentionCreate(&api.Intention{
+					SourceName:      "static-client",
```
ashwin-venkatesh, line 139: Constantize "static-server" and "static-client"? We seem to use them all over the place

test/acceptance/tests/connect/connect_inject_test.go
```diff
@@ -34,37 +34,55 @@ func TestConnectInjectDefault(t *testing.T) {
...
+			consulCluster := framework.NewHelmCluster(t, helmValues, ctx, cfg, releaseName)
+
+			consulCluster.Create(t)
+
+			t.Log("creating static-server and static-client deployments")
+			helpers.Deploy(t, ctx.KubectlOptions(), cfg.NoCleanupOnFailure, "fixtures/static-server.yaml")
+			helpers.Deploy(t, ctx.KubectlOptions(), cfg.NoCleanupOnFailure, "fixtures/static-client.yaml")
+
+			t.Log("checking that the connection is not successful because there's no intention")
+			helpers.CheckStaticServerConnection(t, ctx.KubectlOptions(), "static-client", false, "http://localhost:1234")
```
ashwin-venkatesh, line 72: same thing about constants for "static-client" and "static-server". Also, does it make sense to add a comment about why we are testing the connection against "http://localhost:1234"?

thisisnotashwin (commented):
This PR looks great. Have left a few comments but they are cosmetic for the most part. I keep going back and forth on `enableAutoEncrypt` being a string field that has bool values. In particular:
```
{
			"default",
			false,
			"false",
		},
```
the false, "false".  I dont have a better idea right now, other than making it a bool and using string(enableAutoEncrypt) for the helm values which im not a big fan of either. I don't want that to be a blocker though.

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
ishustava: I can't make `expectedResult` last because the last argument of that function is variadic (takes arbitrary number of additional curl args). I can change it to `(expectedResult, source, destination)` though. Do you think that will read better?

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
ishustava: Do you mean make them exported constants to be shared by all tests? or just within this test?

ashwin-venkatesh: That is an excellent question 😛 
I think per file is alright for now. We can think of global constants if we see an explosion of them.

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
ashwin-venkatesh: oh that makes sense. Sorry. Did not look at the actual signature. `(expectedResult, source, destination)` is perfect. It reads like an assertion if that makes sense. Thanks for indulging these requests 😁

test/acceptance/tests/ingress-gateway/ingress_gateway_test.go
```diff
@@ -12,9 +12,29 @@ import (
 
 // Test that ingress gateways work in a default installation and a secure installation.
 func TestIngressGateway(t *testing.T) {
-	for _, secure := range []bool{false, true} {
-		testName := fmt.Sprintf("secure: %t", secure)
-		t.Run(testName, func(t *testing.T) {
+	cases := []struct {
+		name              string
+		secure            bool
+		enableAutoEncrypt string
```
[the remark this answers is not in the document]
lkysow: Could use `fmt.Sprintf("%t")` in the helm values map.

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
```diff
@@ -11,61 +12,148 @@ import (
...
+			_, err := consulClient.Catalog().Register(&api.CatalogRegistration{
+				Node:     "legacy_node",
+				Address:  "static-server",
+				NodeMeta: map[string]string{"external-node": "true", "external-probe": "true"},
+				Service: &api.AgentService{
+					ID:      "static-server",
+					Service: "static-server",
+					Port:    80,
+				},
+			}, &api.WriteOptions{})
```
lkysow, line 77:
```suggestion
			}, nil)
```

nit

lkysow (approved):
Looks good apart from Ashwin's comments!

thisisnotashwin (approved):

test/acceptance/tests/ingress-gateway/ingress_gateway_test.go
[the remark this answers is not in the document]
ishustava: Updated! I also removed the name and instead used secure and autoEncrypt values to derive it similarly to how the ingress gw test did it before.

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
ishustava: Updated!

test/acceptance/tests/terminating-gateway/terminating_gateway_test.go
ishustava: Constantized!

PR #582, opened 2020-08-20
Update `config-checksum` annotation on config changes

### LINKED ISSUE #106: Upgrade after extra config changes
Changes in `server.extraConfig` or `client.extraConfig` alone don't cause my Consul pods to upgrade. From Kubernetes perspective my definitions don't change, because config map changes are not tracked.

I used to use this approach in my charts before switching to `consul-helm`: [Automatically Roll Deployments When ConfigMaps or Secrets change](https://github.com/helm/helm/blob/master/docs/charts_tips_and_tricks.md#automatically-roll-deployments-when-configmaps-or-secrets-change) from the Helm docs. Consul-helm doesn't use it. Note: my release name is set to a constant value.

What is the recommended way of ensuring that Consul pods get upgraded with changes in extra config? Should my release name change on every change to params?

### LINKED ISSUE #579: Cannot upgrade and turn on ACLs
```
post-install-job
2020-08-19T17:43:55.513Z [ERROR] Failure: bootstrapping ACLs - PUT /v1/acl/bootstrap: err="Unexpected response code: 401 (ACL support disabled)"
```

Getting the above error in acl-init pod after applying update to enable ACLs with
`global.acls.manageSystemACLs: true`

thisisnotashwin:
The `config-checksum` annotation on the server-stateful and client-daemonset would previously only updated on changes to their respective extraConfig. This would fail to restart the server when ACL config was updated. The behaviour has been updated to hash the entire contents of the server/client config configmap so that any changes to the configmap will force a recreate of the server/client pods. This will allows consul upgrades to update an insecure install to a secure one.

Fixes https://github.com/hashicorp/consul-helm/issues/579
Fixes #106 

Testing:
helm install consul with `global.acls.manageSystemACLs: false`
update `global.acls.manageSystemACLs: true` and run helm upgrade
upgrades successfully

The above behavior should hold true for all config that is maintained in the `server-config-configmap` and `client-config-configmap`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ IMPROVEMENTS:
 * Updated the default Consul image to `hashicorp/consul:1.9.0`.
 * Updated the default consul-k8s image to `hashicorp/consul-k8s:0.21.0`.
 * Updated the default envoy image to `envoyproxy/envoy-alpine:v1.16.0`.
+* Add Server ConfigMap and Client ConfigMap values as hashes on Server StatefulSet and Client DaemonSet annotations respectively. This recreates
+  the Server/Client Pod when the Server/Client ConfigMap is updated via `helm upgrade`. This updates the previously hashed values of the extraConfig. [[GH-550](https://github.com/hashicorp/consul-helm/pull/550)]
 
 ## 0.26.0 (Nov 12, 2020)
 
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -32,7 +32,7 @@ spec:
         hasDNS: "true"
       annotations:
         "consul.hashicorp.com/connect-inject": "false"
-        "consul.hashicorp.com/config-checksum": {{ .Values.client.extraConfig | sha256sum }}
+        "consul.hashicorp.com/config-checksum": {{ include (print $.Template.BasePath "/client-config-configmap.yaml") . | sha256sum }}
         {{- if .Values.client.annotations }}
           {{- tpl .Values.client.annotations . | nindent 8 }}
         {{- end }}
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -44,7 +44,7 @@ spec:
         {{- end }}
       annotations:
         "consul.hashicorp.com/connect-inject": "false"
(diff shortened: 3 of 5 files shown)


ashwin-venkatesh:
Holding merge of this PR until we have better documentation describing safely upgrading the consul cluster. Merging this PR in the absence of docs could potentially lead to a situation where this change gets merged and released before the docs arrive which could lead to users having their workloads broken.

lawliet89:
Hey @ashwin-venkatesh I am interested in seeing this fix out ASAP. May I know which upgrade documentation you are referring to? Do you mean the ones added https://github.com/hashicorp/consul/pull/8678?

lkysow:
Hi @lawliet89, yes that was the docs PR! We still need to figure out if that's enough docs so that folks don't shoot themselves in the foot by automatically upgrading consul servers (and potentially causing downtime).

On the one hand, it's a poor UX if we don't auto-rollout changes because you make a helm change and nothing happens, on the other hand it could be outage-causing in production if folks make a small helm change that ends up rolling out new consul servers unexpectedly.

What do you think for your use-case?

lawliet89:
Thanks for your response. I am trying to enable ACL post-bootstrap and it is quite expected that there will be a rolling update to the servers' `StatefulSet` as a result. Still, the docs are very helpful when figuring out how to make changes.

Throckmortra:
Hi again, after having a production issue that DDoS'd our consul with blocking queries, these ACL issues have returned. Helm chart based consul deployments seem to have ongoing issues with ACLs. extraConfig for restarting doesn't seem to help

lkysow:
> Hi again, after having a production issue that DDoS'd our consul with blocking queries, these ACL issues have returned. Helm chart based consul deployments seem to have ongoing issues with ACLs. extraConfig for restarting doesn't seem to help

Hi Aaron, really sorry to hear about your prod issues 😞 .  The ACL logs in https://github.com/hashicorp/consul-helm/issues/650 might be due to the servers/agents not being able to communicate with one another and therefore verify ACLs. We need to look into this further.

lkysow (approved):
Needs changelog entry too

ndhanushkodi (approved):
This looks great!

PR #584, opened 2020-08-25
Add acceptance tests instructions to Contributing

ishustava:
This PR contains the following changes:

* Adds instructions on how to run and write acceptance tests to contributing
* Fixes a race condition in terraform templates. When more than one gke cluster is created, the local-exec commands that add the cluster to kubeconfig can run together, racing each other. I have changed it to instead use a separate kubeconfig file for each cluster to avoid this problem.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,11 +126,11 @@ jobs:
                 echo "Skipping acceptance tests for forked PRs; marking step successful."
                 circleci step halt
             fi
-            primary=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .cluster_names.value[0])
-            secondary=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .cluster_names.value[1])
+            eval "$(echo export primary_kubeconfig=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .kubeconfigs.value[0]))"
+            eval "$(echo export secondary_kubeconfig=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .kubeconfigs.value[1]))"
 
             gotestsum --junitfile $TEST_RESULTS/gotestsum-report.xml -- ./... -p 1 -timeout 20m -enable-multi-cluster \
-              -kubeconfig=$HOME/.kube/$primary -secondary-kubeconfig=$HOME/.kube/$secondary
+              -kubeconfig=$HOME/.kube/$primary_kubeconfig -secondary-kubeconfig=$HOME/.kube/$secondary_kubeconfig
 
       - store_test_results:
           path: /tmp/test-results
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,8 +129,8 @@ jobs:
             eval "$(echo export primary_kubeconfig=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .kubeconfigs.value[0]))"
             eval "$(echo export secondary_kubeconfig=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .kubeconfigs.value[1]))"
 
-            gotestsum --junitfile $TEST_RESULTS/gotestsum-report.xml -- ./... -p 1 -timeout 20m -enable-multi-cluster \
-              -kubeconfig=$HOME/.kube/$primary_kubeconfig -secondary-kubeconfig=$HOME/.kube/$secondary_kubeconfig
+            gotestsum --junitfile "$TEST_RESULTS/gotestsum-report.xml" -- ./... -p 1 -timeout 20m -enable-multi-cluster \
+              -kubeconfig="$primary_kubeconfig" -secondary-kubeconfig="$secondary_kubeconfig"
 
       - store_test_results:
           path: /tmp/test-results
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,11 +126,11 @@ jobs:
                 echo "Skipping acceptance tests for forked PRs; marking step successful."
                 circleci step halt
             fi
-            primary_context=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .context_names.value[0])
-            secondary_context=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .context_names.value[1])
-            kubectl config get-contexts
+            primary=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .cluster_names.value[0])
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -126,10 +126,11 @@ jobs:
...
                 circleci step halt
             fi
-            primary_context=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .context_names.value[0])
-            secondary_context=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .context_names.value[1])
+            primary=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .cluster_names.value[0])
+            secondary=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .cluster_names.value[1])
 
-            gotestsum --junitfile /tmp/gotestsum-report.xml -- ./... -p 1 -timeout 20m -enable-multi-cluster -kubecontext=$primary_context -secondary-kubecontext=$secondary_context
+            gotestsum --junitfile $TEST_RESULTS/gotestsum-report.xml -- ./... -p 1 -timeout 20m -enable-multi-cluster \
+              -kubeconfig=$HOME/.kube/$primary -secondary-kubeconfig=$HOME/.kube/$secondary
```
lkysow, line 133:
```suggestion
gotestsum --junitfile "$TEST_RESULTS/gotestsum-report.xml" -- ./... -p 1 -timeout 20m -enable-multi-cluster \
  -kubeconfig="$HOME/.kube/$primary" -secondary-kubeconfig="$HOME/.kube/$secondary"
```

should quote the args

CONTRIBUTING.md
```diff
@@ -175,3 +222,146 @@ Here are some examples of common test patterns:
...
+If you are adding a feature that fits thematically with one of the existing test suites,
+then you need to add your test cases to the existing test files.
+Otherwise, you will need to create a new test suite.
+
+#### Adding Test Suites
+
+To add a test suite, create a new package directory and add a `main_test.go` file.
+At a minimum, this file needs to contain the following:
+
+```go
```
lkysow, line 237: What do you think about actually creating a directory in tests/acceptance called `template` that has these files we can copy/paste?

My reasoning is that folks are more likely to copy/paste from there than from this doc.

test/terraform/gke/outputs.tf
```diff
@@ -7,5 +7,9 @@ output "cluster_names" {
 }
 
 output "context_names" {
-  value = [for cl in google_container_cluster.cluster : format("gke_%s_%s_%s", var.project, var.zone, cl.name) ]
+  value = [for cl in google_container_cluster.cluster : format("gke_%s_%s_%s", var.project, var.zone, cl.name)]
+}
+
+output "kubeconfigs" {
+  value = [for cl in google_container_cluster.cluster : format("$HOME/.kube/%s", cl.name)]
```
lkysow, line 14: fancy!

test/terraform/gke/outputs.tf
```diff
@@ -7,5 +7,9 @@ output "cluster_names" {
 }
 
 output "context_names" {
-  value = [for cl in google_container_cluster.cluster : format("gke_%s_%s_%s", var.project, var.zone, cl.name) ]
+  value = [for cl in google_container_cluster.cluster : format("gke_%s_%s_%s", var.project, var.zone, cl.name)]
+}
+
+output "kubeconfigs" {
```
lkysow, line 13: Can you add a note that if this output changes, also update .circleci/config.yml?

lkysow (commented):
This looks amazing! My one big question is what you think about moving some of the test instructions into actual code (a "template" test) that folks can copy when they're starting a new test suite.

CONTRIBUTING.md
```diff
@@ -48,21 +52,64 @@ It's expected that the version of `helm` in your path is Helm 3.
 
 In our CI/CD the tests are run against both Helm 2 and Helm 3.
 
+**Note:** Acceptance tests require Helm 3.
```
kschoche, line 55: 👍🏽

CONTRIBUTING.md
kschoche: I like this idea as well.

kschoche (approved):
This looks awesome!
Not a blocker by any means but I think we could reorg the sample code bits into templates files as luke suggested.

ishustava:
@lkysow 

> My one big question is what you think about moving some of the test instructions into actual code (a "template" test) that folks can copy when they're starting a new test suite.

That's a cool idea! I took a stab at it [here](https://github.com/hashicorp/consul-helm/tree/7c84ad01898078e2b27beb45559404b0b3b6edfc/test/acceptance/tests/example). Let me know what you think, and I'll update the docs.

lkysow:
> @lkysow
> 
> > My one big question is what you think about moving some of the test instructions into actual code (a "template" test) that folks can copy when they're starting a new test suite.
> 
> That's a cool idea! I took a stab at it [here](https://github.com/hashicorp/consul-helm/tree/7c84ad01898078e2b27beb45559404b0b3b6edfc/test/acceptance/tests/example). Let me know what you think, and I'll update the docs.

Looks awesome

test/terraform/gke/outputs.tf
ishustava: This actually is a bad thing, so I've modified the circleci config to not have this implicit coupling and use this output directly instead.

.circleci/config.yml
ishustava: done!

.circleci/config.yml
```diff
@@ -126,10 +126,11 @@ jobs:
                 echo "Skipping acceptance tests for forked PRs; marking step successful."
                 circleci step halt
             fi
-            primary_context=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .context_names.value[0])
-            secondary_context=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .context_names.value[1])
+            eval "$(echo export primary_kubeconfig=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .kubeconfigs.value[0]))"
```
lkysow, line 129: why do you need eval here? my bash foo is lacking

test/terraform/gke/outputs.tf
```diff
@@ -7,5 +7,9 @@ output "cluster_names" {
 }
 
 output "context_names" {
```
lkysow, line 9: do you still need this?

lkysow (approved):

.circleci/config.yml
ishustava: haha, it's because the output of terraform contains another env variable (`$HOME`). If I don't eval it, it'll just be passed as a string because it's stored inside another variable.

test/terraform/gke/outputs.tf
ishustava: nope, will remove!

PR #585, opened 2020-08-26
envoy versioning is now set at the global level

kschoche:
Remove meshGateway.imageEnvoy and connectInject.imageEnvoy in favour of global.imageEnvoy which is what ingress and terminating gateways use, this will standardize imageEnvoy across all the gateways and connectInject.

Note : we are removing the option to leave `connectInject.imageEnvoy:  "null"`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ IMPROVEMENTS:
 * Catalog Sync: Can now be run when Consul clients are disabled. It will make API calls to the Consul servers instead. [[GH-570](https://github.com/hashicorp/consul-helm/pull/570)]
 
 BREAKING CHANGES:
-* `connectInject.imageEnvoy` and `meshGateway.imageEnvoy` have been removed and now inherit `global.imageEnvoy`
+* `connectInject.imageEnvoy` and `meshGateway.imageEnvoy` have been removed and now inherit from `global.imageEnvoy`
   which is now standardized across terminating/ingress/mesh gateways and connectInject.
   `global.imageEnvoy` is now a required parameter. [GH-585](https://github.com/hashicorp/consul-helm/pull/585)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ IMPROVEMENTS:
 * Add `dns.type` and `dns.additionalSpec` settings for changing the DNS service type and adding additional spec. [[GH-555](https://github.com/hashicorp/consul-helm/pull/555)]
 * Catalog Sync: Can now be run when Consul clients are disabled. It will make API calls to the Consul servers instead. [[GH-570](https://github.com/hashicorp/consul-helm/pull/570)]
 
+BREAKING CHANGES:
+* `connectInject.imageEnvoy` and `meshGateway.imageEnvoy` have been removed and now inherit `global.imageEnvoy`
+  which is now standardized across terminating/ingress/mesh gateways and connectInject.
+  `global.imageEnvoy` is now a required parameter. [GH-585](https://github.com/hashicorp/consul-helm/pull/585)
+
 ## 0.24.1 (Aug 10, 2020)
 
 BUG FIXES:
--- a/templates/connect-inject-deployment.yaml
+++ b/templates/connect-inject-deployment.yaml
@@ -1,7 +1,7 @@
 {{- if (or (and (ne (.Values.connectInject.enabled | toString) "-") .Values.connectInject.enabled) (and (eq (.Values.connectInject.enabled | toString) "-") .Values.global.enabled)) }}
 {{- if not (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}{{ fail "clients must be enabled for connect injection" }}{{ end }}
 {{- if not .Values.client.grpc }}{{ fail "client.grpc must be true for connect injection" }}{{ end }}
-{{ if .Values.connectInject.imageEnvoy -}} {{ fail "connectInject.imageEnvoy must be specified in global.imageEnvoy" -}} {{ end }}
+{{- if .Values.connectInject.imageEnvoy }}{{ fail "connectInject.imageEnvoy must be specified in global.imageEnvoy" }}{{ end }}
 # The deployment for running the Connect sidecar injector
 apiVersion: apps/v1
 kind: Deployment
--- a/templates/connect-inject-deployment.yaml
+++ b/templates/connect-inject-deployment.yaml
@@ -81,9 +81,7 @@ spec:
               consul-k8s inject-connect \
                 -default-inject={{ .Values.connectInject.default }} \
                 -consul-image="{{ default .Values.global.image .Values.connectInject.imageConsul }}" \
-                {{ if .Values.global.imageEnvoy -}}
                 -envoy-image="{{ .Values.global.imageEnvoy }}" \
-                {{ end -}}
                 -consul-k8s-image="{{ default .Values.global.imageK8S .Values.connectInject.image }}" \
                 -listen=:8080 \
                 {{- if .Values.connectInject.overrideAuthMethodName }}
--- a/templates/connect-inject-deployment.yaml
+++ b/templates/connect-inject-deployment.yaml
@@ -1,7 +1,7 @@
 {{- if (or (and (ne (.Values.connectInject.enabled | toString) "-") .Values.connectInject.enabled) (and (eq (.Values.connectInject.enabled | toString) "-") .Values.global.enabled)) }}
 {{- if not (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}{{ fail "clients must be enabled for connect injection" }}{{ end }}
 {{- if not .Values.client.grpc }}{{ fail "client.grpc must be true for connect injection" }}{{ end }}
-{{ if .Values.connectInject.imageEnvoy -}} {{ fail "connectInject.imageEnvoy must be specified in global" -}} {{ end }}
+{{ if .Values.connectInject.imageEnvoy -}} {{ fail "connectInject.imageEnvoy must be specified in global.imageEnvoy" -}} {{ end }}
 # The deployment for running the Connect sidecar injector
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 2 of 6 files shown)


templates/connect-inject-deployment.yaml
```diff
@@ -1,6 +1,7 @@
 {{- if (or (and (ne (.Values.connectInject.enabled | toString) "-") .Values.connectInject.enabled) (and (eq (.Values.connectInject.enabled | toString) "-") .Values.global.enabled)) }}
 {{- if not (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}{{ fail "clients must be enabled for connect injection" }}{{ end }}
 {{- if not .Values.client.grpc }}{{ fail "client.grpc must be true for connect injection" }}{{ end }}
+{{ if .Values.connectInject.imageEnvoy -}} {{ fail "connectInject.imageEnvoy must be specified in global" -}} {{ end }}
```
ishustava, line 4:
```suggestion
{{ if .Values.connectInject.imageEnvoy -}} {{ fail "connectInject.imageEnvoy must be specified in global.imageEnvoy" -}} {{ end }}
```

values.yaml
```diff
@@ -232,6 +228,13 @@ global:
         memory: "50Mi"
         cpu: "20m"
 
+  # The Docker image for envoy to use as the proxy sidecar when performing
+  # Connect injection or using the mesh/terminating/ingress gateways.
```
ishustava, line 232: Could we rephrase or break up this sentence? I think it now reads a bit confusing. Maybe something like:

```
imageEnvoy is the name (and tag) of the Envoy Docker image used for the
connect-injected sidecar proxies and the mesh, terminating, and ingress gateways.
```

values.yaml
```diff
@@ -232,6 +228,13 @@ global:
...
+  #. If using Consul 1.7+, the envoy version must be 1.13+.
```
ishustava, line 233: I think we should point to the consul-envoy compatibility [docs](https://www.consul.io/docs/connect/proxies/envoy).

templates/connect-inject-deployment.yaml
```diff
@@ -80,8 +81,8 @@ spec:
               consul-k8s inject-connect \
                 -default-inject={{ .Values.connectInject.default }} \
                 -consul-image="{{ default .Values.global.image .Values.connectInject.imageConsul }}" \
-                {{ if .Values.connectInject.imageEnvoy -}}
-                -envoy-image="{{ .Values.connectInject.imageEnvoy }}" \
+                {{ if .Values.global.imageEnvoy -}}
```
ishustava, line 84: should we always require it? that way we can eventually get rid of the default in consul-k8s?

templates/mesh-gateway-deployment.yaml
```diff
@@ -2,6 +2,7 @@
 {{- if not .Values.connectInject.enabled }}{{ fail "connectInject.enabled must be true" }}{{ end -}}
 {{- if not .Values.client.grpc }}{{ fail "client.grpc must be true" }}{{ end -}}
 {{- if and .Values.global.acls.manageSystemACLs (ne .Values.meshGateway.consulServiceName "") (ne .Values.meshGateway.consulServiceName "mesh-gateway") }}{{ fail "if global.acls.manageSystemACLs is true, meshGateway.consulServiceName cannot be set" }}{{ end -}}
+{{ if .Values.meshGateway.imageEnvoy -}} {{ fail "meshGateway.imageEnvoy must be specified in global" -}} {{ end }}
```
ishustava, line 5:
```suggestion
{{ if .Values.meshGateway.imageEnvoy -}} {{ fail "meshGateway.imageEnvoy must be specified in global.imageEnvoy" -}} {{ end }}
```

ishustava (commented):
That's awesome! Thanks for making this change 🙏 

I left a couple of edits. I'm also now thinking about whether we should keep the behavior when you set `imageEnvoy` to `null` for connect? Since we're consolidating it into one property, I think it'd make sense for it to behave the same way across connect and gateways. What do you think?

I'm also curious if you have thought about what to do with the default in consul-k8s. Since we're making this change and it'll be a breaking change, would it makes sense to also make a breaking change in consul-k8s and remove the default?

values.yaml
kschoche:
```suggestion
  # imageEnvoy is the name (and tag) of the Envoy Docker image used for the
  # connect-injected sidecar proxies and the mesh, terminating, and ingress gateways.
```

templates/connect-inject-deployment.yaml
kschoche: my plan was to take up the consul-k8s work as well since it's fairly easy and raise a PR there to make it mandatory and remove the defaults.
I suppose it makes sense to hold off on merging this until that work is done?

lkysow: Yeah let's not support `global.imageEnvoy: null` since from the beginning this value has had a default.

Re consul-k8s, I don't think you need to wait on merging this PR for that consul-k8s change.

templates/connect-inject-deployment.yaml
```diff
@@ -1,6 +1,7 @@
 {{- if (or (and (ne (.Values.connectInject.enabled | toString) "-") .Values.connectInject.enabled) (and (eq (.Values.connectInject.enabled | toString) "-") .Values.global.enabled)) }}
 {{- if not (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}{{ fail "clients must be enabled for connect injection" }}{{ end }}
 {{- if not .Values.client.grpc }}{{ fail "client.grpc must be true for connect injection" }}{{ end }}
+{{ if .Values.connectInject.imageEnvoy -}} {{ fail "connectInject.imageEnvoy must be specified in global.imageEnvoy" -}} {{ end }}
```
lkysow, line 4:
```suggestion
{{- if .Values.connectInject.imageEnvoy }}{{ fail "connectInject.imageEnvoy must be specified in global.imageEnvoy" }}{{ end }}
```

For consistency with the above fails.

templates/mesh-gateway-deployment.yaml
```diff
@@ -2,6 +2,7 @@
 {{- if not .Values.connectInject.enabled }}{{ fail "connectInject.enabled must be true" }}{{ end -}}
 {{- if not .Values.client.grpc }}{{ fail "client.grpc must be true" }}{{ end -}}
 {{- if and .Values.global.acls.manageSystemACLs (ne .Values.meshGateway.consulServiceName "") (ne .Values.meshGateway.consulServiceName "mesh-gateway") }}{{ fail "if global.acls.manageSystemACLs is true, meshGateway.consulServiceName cannot be set" }}{{ end -}}
+{{ if .Values.meshGateway.imageEnvoy -}} {{ fail "meshGateway.imageEnvoy must be specified in global.imageEnvoy" -}} {{ end }}
```
lkysow, line 5:
```suggestion
{{- if .Values.meshGateway.imageEnvoy }}{{ fail "meshGateway.imageEnvoy must be specified in global.imageEnvoy" }}{{ end -}}
```

(looks like these ones have `end -` and the ones in connect injector have just `end`, guess it doesn't make a difference so we could just keep consistency within each file or change all of them to one pattern. Fine either way.)

values.yaml
```diff
@@ -232,6 +228,13 @@ global:
         memory: "50Mi"
         cpu: "20m"
 
+  # imageEnvoy is the name (and tag) of the Envoy Docker image used for the
+  # connect-injected sidecar proxies and the mesh, terminating, and ingress gateways.
+  #. If using Consul 1.7+, the envoy version must be 1.13+.
```
lkysow, line 233:
```suggestion
  # If using Consul 1.7+, the envoy version must be 1.13+.
```

values.yaml
```diff
@@ -232,6 +228,13 @@ global:
...
+  # If not set, the image used depends on the consul-k8s version. For
+  # consul-k8s 0.12.0 the default is envoyproxy/envoy-alpine:v1.13.0.
```
lkysow, line 235: Reminder to delete this comment since not setting this isn't supported.

lkysow (commented):
Looks great! Needs changelog entry.

values.yaml
kschoche:
```suggestion
```

values.yaml
```diff
@@ -232,6 +228,11 @@ global:
         memory: "50Mi"
         cpu: "20m"
 
+  # imageEnvoy is the name (and tag) of the Envoy Docker image used for the
+  # connect-injected sidecar proxies and the mesh, terminating, and ingress gateways.
+  # If using Consul 1.7+, the envoy version must be 1.13+.
```
kschoche, line 233:
```suggestion
  # See https://www.consul.io/docs/connect/proxies/envoy for full compatibility matrix between Consul and Envoy.
```

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ IMPROVEMENTS:
 * Add `dns.type` and `dns.additionalSpec` settings for changing the DNS service type and adding additional spec. [[GH-555](https://github.com/hashicorp/consul-helm/pull/555)]
 * Catalog Sync: Can now be run when Consul clients are disabled. It will make API calls to the Consul servers instead. [[GH-570](https://github.com/hashicorp/consul-helm/pull/570)]
 
+BREAKING CHANGES:
+* `connectInject.imageEnvoy` and `meshGateway.imageEnvoy` have been removed and now inherit `global.imageEnvoy`
```
lkysow, line 9:
```suggestion
* `connectInject.imageEnvoy` and `meshGateway.imageEnvoy` have been removed and now inherit from `global.imageEnvoy`
```

lkysow (approved):
Nice! Love the `fail` messages as well, nice touch for users that miss the changelog.

templates/connect-inject-deployment.yaml
ishustava: Cool, sounds good. Yeah, I agree, no need to block this PR on the consul-k8s work. 

If we require it, then we should remove this `if` condition. Having the `if` means that you can still set it to `null` in the helm chart and use consul-k8s default.

ishustava (approved):
Looks good! 

The only minor change is that we need to remove `{{ if .Values.global.imageEnvoy -}}` from the connect inject deployment since we're now requiring this value to be set. Once that's done, it's good to go.

PR #586, opened 2020-08-28
Write debug info when tests fail

ishustava:
This PR proposes the following changes: 

* Adds a new flag `-debug-directory`.
* When the flag is not set, the tests will create a temporary directory.
* On cleanup, the consul cluster and
helpers.Deploy functions will write pod info and logs
to the debug directory. 

For the time being, we will only support these functions, but if more support is needed, it might make sense to provide this more natively within the framework for each test.

See [this build](https://app.circleci.com/pipelines/github/hashicorp/consul-helm/1203/workflows/fb420d14-436d-457a-a192-6803bbcd1d7a/jobs/1909/artifacts) for an example of what is written when tests fail.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -90,6 +90,8 @@ Below is the list of available flags:
     The Consul image to use for all tests.
 -consul-k8s-image string
     The consul-k8s image to use for all tests.
+-debug-directory
+    The directory where to dump debug information about test runs, such as logs, pod definitions etc. If not provided, a temporary directory will be created by the tests.
 -enable-multi-cluster
     If true, the tests that require multiple Kubernetes clusters will be run. At least one of -secondary-kubeconfig or -secondary-kubecontext is required when this flag is used.
 -kubeconfig string
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -91,7 +91,7 @@ Below is the list of available flags:
 -consul-k8s-image string
     The consul-k8s image to use for all tests.
 -debug-directory
-    The directory where to write debug information about test runs, uch as logs and pod definitions. If not provided, a temporary directory will be created by the tests.
+    The directory where to write debug information about failed test runs, such as logs and pod definitions. If not provided, a temporary directory will be created by the tests.
 -enable-multi-cluster
     If true, the tests that require multiple Kubernetes clusters will be run. At least one of -secondary-kubeconfig or -secondary-kubecontext is required when this flag is used.
 -kubeconfig string
@@ -101,7 +101,7 @@ Below is the list of available flags:
 -namespace string
     The Kubernetes namespace to use for tests. (default "default")
 -no-cleanup-on-failure
-    If true, the tests will not cleanup resources they create when they finish running.Note this flag must be run with -failfast flag, otherwise subsequent tests will fail.
+    If true, the tests will not cleanup Kubernetes resources they create when they finish running.Note this flag must be run with -failfast flag, otherwise subsequent tests will fail.
 -secondary-kubeconfig string
     The path to a kubeconfig file of the secondary k8s cluster. If this is blank, the default kubeconfig path (~/.kube/config) will be used.
 -secondary-kubecontext string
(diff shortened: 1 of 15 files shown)


test/acceptance/helpers/helpers.go
```diff
@@ -148,3 +152,95 @@ func Cleanup(t *testing.T, noCleanupOnFailure bool, cleanup func()) {
...
+	if t.Failed() {
+		// Create k8s client from kubectl options
+		client := KubernetesClientFromOptions(t, kubectlOptions)
+
+		contextName := kubernetesContextFromOptions(t, kubectlOptions)
+
+		// Create a temporary directory if one isn't provided.
+		if debugDirectory == "" {
+			var err error
+			debugDirectory, err = ioutil.TempDir("", "consul-test")
```
kschoche, line 170: Not a blocker by any means, but I can imagine that this will overwrite the last test's results, and maybe it would be helpful to be able to compare the two while debugging something?
Would it make sense to always use `consul-test` or should we salt the name with some sort of identifier so we don't overwrite the previous test results?

kschoche: Actually it looks like you're doing that when generating testDebugDirectory() .. nevermind! 👍🏽

kschoche (approved):
This looks awesome, fantastic work! 👍🏽 💯 🥇 

One small comment : 
Do you envision a case where there might be a clash between the settings for noCleanupOnFailure and debugDirectory ?  i.e. 
`-debug-directory=/tmp/abcd -no-cleanup-on-failure=false`
 when there is a failure does the cleanup also imply cleanup of the debug logs?

It seems natural to not cleanup debug logs if it's set (no changes required).

test/acceptance/framework/environment.go
```diff
@@ -87,22 +91,22 @@ func (k kubernetesContext) KubernetesClient(t *testing.T) kubernetes.Interface {
...
-	config, err := k8s.LoadApiClientConfigE(configPath, k.contextName)
+	t.Logf("Creating client from config path at %s for context %s", configPath, k.kubeContextName)
+	config, err := k8s.LoadApiClientConfigE(configPath, k.kubeContextName)
 	require.NoError(t, err)
 
 	client, err := kubernetes.NewForConfig(config)
 	require.NoError(t, err)
 
-	k.client = client
+	k.client = helpers.KubernetesClientFromOptions(t, k.KubectlOptions())
```
ashwin-venkatesh, line 101: Is there a difference between these two client? Can we get away with using the client from L98 here? That way, we could make `helpers.KubernetesClientFromOptions` package private

thisisnotashwin (commented):
Really love the addition of this feature. Have a suggestions and a quick question regarding how the Kube Client is used.

test/acceptance/framework/environment.go
ishustava: 🤦‍♀️ they are the same, but the goal was to refactor this to helpers. So all the code above line 101 should not be there. Let me update.

ashwin-venkatesh: Using the helper package makes more sense!! 🚢

CONTRIBUTING.md
```diff
@@ -90,6 +90,8 @@ Below is the list of available flags:
     The Consul image to use for all tests.
 -consul-k8s-image string
     The consul-k8s image to use for all tests.
+-debug-directory
+    The directory where to dump debug information about test runs, such as logs, pod definitions etc. If not provided, a temporary directory will be created by the tests.
```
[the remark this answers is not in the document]
ishustava: I've rephrased that a bit.

test/acceptance/framework/environment.go
ishustava: Updated!

ishustava:
@kschoche 

> Do you envision a case where there might be a clash between the settings for noCleanupOnFailure and debugDirectory ? i.e.
-debug-directory=/tmp/abcd -no-cleanup-on-failure=false
when there is a failure does the cleanup also imply cleanup of the debug logs?

I don't see a clash: one is controlling whether we should delete resources after tests fail and the other will write debug information. Cleanup only applies to resources created by the tests in the kube cluster. I've clarified this in the `-no-cleanup-on-failure` flag definition.

thisisnotashwin (approved):
🎉

kschoche (approved):
Looks great!

PR #587, opened 2020-08-31
Support ACLs for controller

lkysow:
This PR adds support for ACLs. It only works for OSS. Ent will be done in another PR.

Prereqs:
```
```

Testing:
```
helm install consul ./ \
  --set controller.enabled=true \
  --set global.imageK8S="lkysow/consul-k8s-dev:aug31-ctrl-acl" \
  --set server.replicas=1 \
  --set server.bootstrapExpect=1 \
  --set global.acls.manageSystemACLs=true \
  --wait

kind: ServiceDefaults
metadata:
  name: servicedefaults-sample
spec:
  protocol: "http"
EOF

```

Cleanup:

Companion with https://github.com/hashicorp/consul-k8s/pull/317

### DIFF
--- a/templates/controller-clusterrole.yaml
+++ b/templates/controller-clusterrole.yaml
@@ -10,32 +10,41 @@ metadata:
     release: {{ .Release.Name }}
     component: controller
 rules:
-  - apiGroups:
-      - consul.hashicorp.com
-    resources:
-      - servicedefaults
-    verbs:
-      - create
-      - delete
-      - get
-      - list
-      - patch
-      - update
-      - watch
-  - apiGroups:
-      - consul.hashicorp.com
-    resources:
-      - servicedefaults/status
-    verbs:
-      - get
-      - patch
-      - update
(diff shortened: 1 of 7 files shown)


test/unit/sync-catalog-clusterrole.bats
```diff
@@ -63,7 +63,7 @@ load _helpers
 #--------------------------------------------------------------------
 # global.acls.manageSystemACLs
 
-@test "syncCatalog/ClusterRole: allows secret access with global.bootsrapACLs=true" {
+@test "syncCatalog/ClusterRole: allows secret access with global.acls.manageSystemACLs=true" {
```
lkysow, line 66: typo

templates/controller-clusterrole.yaml
```diff
@@ -10,24 +10,33 @@ metadata:
     release: {{ .Release.Name }}
     component: controller
 rules:
-  - apiGroups:
```
lkysow, line 13: fixing indentation here

PR #589, opened 2020-08-31
Support TLS and TLS auto-encrypt for controller

lkysow:
This PR adds support for TLS and TLS auto-encrypt.

Prereqs:
```
```

### Testing

TLS
```yaml
controller:
  enabled: true
global:
  imageK8S: lkysow/consul-k8s-dev:aug31-ctrl-acl
  tls:
    enabled: true
server:
  replicas: 1
  bootstrapExpect: 1
```

TLS Autoencrypt
```yaml
controller:
  enabled: true
global:
  imageK8S: lkysow/consul-k8s-dev:aug31-ctrl-acl
  tls:
    enabled: true
    enableAutoEncrypt: true
server:
  replicas: 1
  bootstrapExpect: 1
```

Once installed, create the CRD and see that it is created in consul:
```
kind: ServiceDefaults
metadata:
  name: servicedefaults-sample
spec:
  protocol: "http"
EOF

```

Cleanup:

### DIFF
--- a/templates/connect-inject-deployment.yaml
+++ b/templates/connect-inject-deployment.yaml
@@ -205,7 +205,7 @@ spec:
               mountPath: /consul/tls/ca
               readOnly: true
             {{- end }}
-          {{- end }}
+            {{- end }}
           {{- with .Values.connectInject.resources }}
           resources:
             {{- toYaml . | nindent 12 }}
--- a/templates/controller-deployment.yaml
+++ b/templates/controller-deployment.yaml
@@ -27,8 +27,9 @@ spec:
         release: {{ .Release.Name }}
         component: controller
     spec:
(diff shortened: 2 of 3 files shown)


templates/connect-inject-deployment.yaml
```diff
@@ -206,7 +206,7 @@ spec:
               mountPath: /consul/tls/ca
               readOnly: true
             {{- end }}
-          {{- end }}
+            {{- end }}
```
lkysow, line 209: Just fixing identation to line up with opening `if` on line 200

kschoche (approved):
Looks great!

thisisnotashwin (approved):
These helm if conditions make my head hurt, but this is 🥇

PR #591, opened 2020-08-31
Crd controller psps

lkysow:
This PR adds support for pod security policies.

Testing this is a bit annoying because cert-manager doesn't support pod security policies. We're getting rid of cert manager soon but until then you need to install cert-manager with psps *disabled* and then enable them before installing the controller.

Using a GKE cluster (where it's easiest to enable psps in my experience).

Prereqs (**with psps disabled**):
```
```

Enable PSPs:
```
gcloud beta container clusters update <your-cluster-name> --enable-pod-security-policy --zone us-central1-c
```

Install with:
```yaml
global:
  enablePodSecurityPolicies: true
  imageK8S: lkysow/consul-k8s-dev:aug31-ctrl-acl
controller:
  enabled: true
server:
  replicas: 1
  bootstrapExpect: 1
```

Once installed, create the CRD and see that it is created in consul:
```
kind: ServiceDefaults
metadata:
  name: servicedefaults-sample
spec:
  protocol: "http"
EOF

```

Cleanup:

### DIFF
--- a/templates/controller-podsecuritypolicy.yaml
+++ b/templates/controller-podsecuritypolicy.yaml
@@ -1,1 +1,39 @@
-# todo
+{{- if and .Values.controller.enabled .Values.global.enablePodSecurityPolicies }}
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: {{ template "consul.fullname" . }}-controller
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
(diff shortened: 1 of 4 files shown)


templates/controller-podsecuritypolicy.yaml
```diff
@@ -1 +1,39 @@
-# todo
+{{- if and .Values.controller.enabled .Values.global.enablePodSecurityPolicies }}
+apiVersion: policy/v1beta1
```
lkysow, line 2: This is using our minimal psp we use elsewhere.

ishustava (approved):
Looks great!

thisisnotashwin (approved):
💥

PR #592, opened 2020-09-01
Refactor common fixtures to use kustomize

ishustava:
This reduces the duplication between fixtures and also makes it easier to add new fixtures that we will need for namespace tests.


thisisnotashwin (approved):
💥

test/acceptance/tests/fixtures/cases/static-client-inject/patch.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: static-client
+spec:
+  selector:
+    matchLabels:
+      app: static-client
```
lkysow, line 8: Do you need these lines? I briefly tested without them and it generated the same?

lkysow (approved):

test/acceptance/tests/fixtures/cases/static-client-inject/patch.yaml
ishustava: Interesting,  I thought I did but that could have been one of many variations I tried that I don't remember now 😄 Let me try.

ishustava: Yup, didn't need those, thanks!

PR #593, opened 2020-09-02
Acceptance tests: add enterprise support and namespace tests

ishustava:
This PR adds support for running tests with consul enterprise and adds tests for consul namespaces across connect, sync, ingress, and terminating gateways.

Note that in CI we're not providing the enterprise license and just using the temporary 30min license the consul binary includes. This works for the tests since they take less than 30min, so, for the time being, we will not test it with enterprise license. This will be left as a future improvement.


test/acceptance/tests/connect/connect_inject_namespaces_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/hashicorp/consul-helm/test/acceptance/framework"
+	"github.com/hashicorp/consul-helm/test/acceptance/helpers"
+	"github.com/hashicorp/consul/api"
+	"github.com/stretchr/testify/require"
+)
+
+const staticServerNamespace = "ns1"
+const staticClientNamespace = "ns2"
+
+// Test that Connect works in a default installation
```
lkysow, line 17:
```suggestion
// Test that Connect works with Consul Enterprise namespaces.
```

test/acceptance/tests/connect/connect_inject_namespaces_test.go
```diff
@@ -0,0 +1,147 @@
...
+				helpers.RunKubectl(t, ctx.KubectlOptions(), "delete", "ns", staticClientNamespace)
+			})
+
+			t.Log("creating static-server and static-client deployments")
+			helpers.DeployKustomize(t, staticServerOpts, cfg.NoCleanupOnFailure, cfg.DebugDirectory, "../fixtures/cases/static-server-inject")
+			helpers.DeployKustomize(t, staticClientOpts, cfg.NoCleanupOnFailure, cfg.DebugDirectory, "../fixtures/cases/static-client-namespaces")
+
+			consulClient := consulCluster.SetupConsulClient(t, c.secure)
+
+			// Make sure that services are registered in the correct namespace
```
lkysow, line 103:
```suggestion
			// Make sure that services are registered in the correct namespace.
			// If mirroring is enabled, we expect services to be registered in the
			// Consul namespace with the same name as their source
			// Kubernetes namespace.
			// If a single destination namespace is set, we expect all services
			// to be registered in that destination Consul namespace. 
```

test/acceptance/tests/connect/connect_inject_namespaces_test.go
```diff
@@ -0,0 +1,147 @@
...
+			true,
+			false,
+		},
+		{
+			"mirror k8s namespaces; secure",
+			staticServerNamespace,
+			true,
+			true,
+		},
+	}
```
lkysow, line 54: Do you want to make a note that we're not testing auto-encrypt? Presumably because the tests were too long?

test/acceptance/tests/fixtures/cases/static-client-inject/patch.yaml
```diff
@@ -3,6 +3,7 @@ kind: Deployment
 metadata:
   name: static-client
 spec:
+
```
lkysow, line 6: ?

test/acceptance/tests/ingress-gateway/ingress_gateway_namespaces_test.go
```diff
@@ -0,0 +1,263 @@
...
+			if !c.secure {
+				t.Logf("creating the %s namespace in Consul", testNamespace)
+				_, _, err := consulClient.Namespaces().Create(&api.Namespace{
+					Name: testNamespace,
+				}, nil)
+				require.NoError(t, err)
+			}
+
+			t.Log("upgrading with ingress gateways enabled")
+			consulCluster.Upgrade(t, map[string]string{
```
lkysow, line 69: Good thing this was implemented :D

test/acceptance/tests/connect/connect_inject_namespaces_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			ctx := suite.Environment().DefaultContext(t)
+
+			helmValues := map[string]string{
+				"global.enableConsulNamespaces":                             "true",
+				"connectInject.enabled":                                     "true",
+				"connectInject.consulNamespaces.consulDestinationNamespace": c.destinationNamespace,
```
lkysow, line 63: Worth noting that if both dest and mirroring are set that mirroring overrides?

test/acceptance/tests/terminating-gateway/terminating_gateway_namespaces_test.go
```diff
@@ -0,0 +1,232 @@
...
+			}
+
+			// Test that we can make a call to the terminating gateway
+			t.Log("trying calls to terminating gateway")
+			helpers.CheckStaticServerConnection(t, ns2K8SOptions, true, staticClientName, "http://localhost:1234")
+		})
+	}
+}
+
+const staticServerPolicyRulesNamespace = `namespace %s {
```
lkysow, line 229:
```suggestion
const staticServerPolicyRulesNamespace = `namespace %q {
```

Surprised this doesn't need to be quoted.

lkysow (approved):
Can we use mirror*ing*K8S to match the helm setting name?

test/acceptance/tests/connect/connect_inject_test.go
```diff
@@ -11,54 +13,28 @@ import (
 
 const staticClientName = "static-client"
```
ashwin-venkatesh, line 14: Can we create a const for `staticServerName` as well and use that across the tests?

thisisnotashwin (approved):
this looks great, aside from the comments left by Luke and a small suggestion. 🔥

test/acceptance/tests/terminating-gateway/terminating_gateway_namespaces_test.go
ishustava: It did need to be quoted, the tests failed 😄 Good catch, thank you 🙏

test/acceptance/tests/connect/connect_inject_namespaces_test.go
ishustava: Sure, I'll make a note. I just didn't think there are any special changes we needed to do for auto-encrypt and namespaces, that's why I didn't add a test for it. I thought having just one secure installation was good enough.

test/acceptance/tests/connect/connect_inject_test.go
ishustava: yes, definitely 😄

test/acceptance/tests/fixtures/cases/static-client-inject/patch.yaml
ishustava: Just checking if you're paying attention 😄

test/acceptance/tests/ingress-gateway/ingress_gateway_namespaces_test.go
ishustava: 🙏

PR #596, opened 2020-09-07
Giving the possibility to bring own client cert

### LINKED ISSUE #588: Failed to join <server IPs>: Remote state is encrypted and encryption is not configured
Hi,

I'm trying to deploy consul clients within kubernetes and got the servers outside. When I do so I receive this error `message: Failed to join <server IPs>: Remote state is encrypted and encryption is not configured`. I agree I haven't setup gossip encryption as I'm using mTLS so I supposed setting tls was enough but it doesn't seem to. Here is my values.yaml file:

```
server:
  enabled: false

externalServers:
  enabled: true
  hosts: ["<node1.my.domain>", "<node2.my.domain>", "<node3.my.domain>"]
  useSystemRoots: true

client:
  image: consul:1.7.0
  join:  ["<node1.my.domain>", "<node2.my.domain>", "<node3.my.domain>"]

global:
  domain: <my.domain>
  tls:
    enabled: true
    caCert:
      secretName: consul-agent
      secretKey: tls.crt
    caKey:
      secretName: consul-agent
      secretKey: tls.key
```

Might not be fully complete so that's why I'm asking for help

Additional information:
I'm using kubernetes 1.18 on a self hosted cluster, consul is also running internally on three VMs with version 1.7.0 thus tried to deploy 1.7.0 agents, deployed with helm 2 and version 0.24.1 of the helm chart

zonArt:
In order to bring own client certificate with a cluster of consul servers outside kubernetes. When using any other PKI and managing its own certificate and keys but without having a dedicated CA to give to consul agent pods
Fix #588

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -84,6 +84,7 @@ spec:
             - key: {{ default "tls.crt" .Values.global.tls.caCert.secretKey }}
               path: tls.crt
         {{ if not .Values.global.tls.enableAutoEncrypt }}
+        {{ if not .Values.global.tls.bringOwnKey }}
         - name: consul-ca-key
           secret:
             {{- if .Values.global.tls.caKey.secretName }}
@@ -99,6 +100,21 @@ spec:
             # We're using tmpfs here so that
             # client certs are not written to disk
             medium: "Memory"
+        {{- end }}
+        {{- end }}
+        {{ if .Values.global.tls.bringOwnKey }}
+        - name: consul-client-cert
+          secret:
+            {{- if .Values.global.tls.clientCert.secretName }}
+            secretName: {{ .Values.global.tls.clientCert.secretName }}
+            {{- else }}
+            secretName: {{ template "consul.fullname" . }}-cert
+            {{- end }}
+            items:
+            - key: {{ default "tls.crt" .Values.global.tls.clientCert.secretKey }}
+              path: tls.crt
+            - key: {{ default "tls.key" .Values.global.tls.clientKey.secretKey }}
+              path: tls.key
         {{- end }}
         {{- end }}
         {{- range .Values.client.extraVolumes }}
@@ -333,7 +349,7 @@ spec:
             memory: "25Mi"
             cpu: "50m"
       {{- end }}
-      {{- if and .Values.global.tls.enabled (not .Values.global.tls.enableAutoEncrypt) }}
+      {{- if and .Values.global.tls.enabled (not .Values.global.tls.enableAutoEncrypt) (not .Values.global.tls.bringOwnKey) }}
       - name: client-tls-init
         image: "{{ default .Values.global.image .Values.client.image }}"
         env:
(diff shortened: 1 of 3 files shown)


dynek (approved):
LGTM

zonArt:
@lkysow : Closing the issue was a good idea, I was going to do it anyway as you answered all my questions. As you suggested let continue the discussion here. What do you think about my proposal and what should I change to see it merged?

lkysow:
Hi @zonArt thanks for the PR! We wanted to dig into the use-case here first.

1. How come you can't give the CA cert/key to Kubernetes? "but without having a dedicated CA to give to consul agent pods"
1. Have you tried using the `client.extraVolumes` and `client.extraConfig` settings to mount in your own certs?

zonArt:
> Hi @zonArt thanks for the PR! We wanted to dig into the use-case here first.
> 
> 1. How come you can't give the CA cert/key to Kubernetes? "but without having a dedicated CA to give to consul agent pods"
> 2. Have you tried using the `client.extraVolumes` and `client.extraConfig` settings to mount in your own certs?

1. Because the CA is managed by Vault and we're mounting certificates as secret directly with cert-manager and vaultbot on our consul servers
2. Mmmmmh I saw the extraVolumes param but it mounts the cert at a hardcoded path, maybe extraConfig is the key here, I guess you can this way tell consul where to pick the certificates right? If that's the case, you're probably right, this PR might be fully useless and I'm sorry for that

zonArt:
@lkysow Just tested extraVolumes and extraConfig, feeling dumb right now... Sorry for the noise

lkysow:
No worries! We don't have this documented very well so it would have been hard to find. Let us know if there's any problems with this approach.

One thing that might be useful to look into is using auto-encrypt (https://learn.hashicorp.com/tutorials/consul/tls-encryption-secure). This allows Consul servers to push certs to consul clients. Then they don't need their own certs.

It can be enabled with
```
global:
  tls:
    enabled: true
    enableAutoEncrypt: true
```

zonArt:
Hi @lkysow,

I'm actually still having an issue with this approach yes. The issue I have is that I cannot enable tls because of the initContainer which is then created and fails each time. May I open another PR with only the flag "bringOwnKey" so that I can get rid of the initContainer but still be able to benefice from exposing port 8501 when choosing to enable tls?

zonArt:
Here it is: #603

ashwin-venkatesh:
Hey @zonArt,

Just picking this up from @lkysow here.

>  I cannot enable tls because of the initContainer which is then created and fails each time

Out of curiosity, did you see this occurring even after `enableAutoEncrypt` was set to true? If `server.enabled` is set to `false`, which i suspect is your case give the servers are external, the `tls-init` job should not run.

zonArt:
> Hey @zonArt,
> 
> Just picking this up from @lkysow here.
> 
> > I cannot enable tls because of the initContainer which is then created and fails each time
> 
> Out of curiosity, did you see this occurring even after `enableAutoEncrypt` was set to true? ~If `server.enabled` is set to `false`, which i suspect is your case give the servers are external, the `tls-init` job should not run~. Setting `enableAutoEncrypt` to true should ensure the `tls-init` job on the client does not run.

Here is what happens if I do as you suggest
  Warning  FailedMount  21s (x7 over 52s)  kubelet, node1.my.domain  MountVolume.SetUp failed for volume "consul-ca-cert" : secret "consul-consul-ca-cert" not found

PR #600, opened 2020-09-08
Openshift support

ishustava:
### Helm chart values and template changes
- Add a new Helm value `global.openshift.enabled`
- Add a `SecurityContextConstraint` for the consul clients when `global.openshift.enabled` is set to true
- Don't set fsGroup for the servers when `global.openshift.enabled` is set to true
- Remove `server.disableFsGroupSecurityContext` value and fail in the chart if someone tries to set it
- Increase memory limits and requests from `25Mi` to `50Mi` for all jobs and service-init containers in the terminating and ingress gateway deployments. This was mostly determined by running the tests. There are other containers that still have the `25Mi` memory request and limit, however those containers were not causing failures on OpenShift.

### Acceptance tests changes
- Add a new flag `-enable-openshift` to the framework which will set `global.openshift.enabled` to true for all helm installs/upgrades.
- Increase timeouts in various places because it takes longer for things to be created on OpenShift.
- Change consul-dns test to retry and to not use TTY since it's not always available.

### CI changes
- Add a new job to run acceptance tests against OpenShift. Note that it currently runs against pre-created Azure Red Hat OpenShift clusters. I was not able to get terraform running to create them before each run due to a variety of reasons, the main one being that it takes a not yet determined amount of time for the clusters to become ready. It has been hard to determine that time in a script/CI, which resulted in intermittent/non-deterministic test failures. We will have to address further automation of the OpenShift cluster creation at a later time.
- Add a new workflow to run acceptance tests nightly (currently commented out).
- Add Azure CLI and OpenShift CLI to the Docker image used in CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -156,6 +156,7 @@ jobs:
   acceptance-openshift:
     environment:
       - TEST_RESULTS: /tmp/test-results
+      # Primary and secondary Azure OpenShift clusters (created manually) that are used to run acceptance tests.
       - OC_PRIMARY_NAME: consul-helm-test-2757871175
       - OC_SECONDARY_NAME: consul-helm-test-3737660519
     docker:
@@ -243,27 +244,26 @@ workflows:
             - go-fmt-and-vet
       - unit-helm2
       - unit-helm3
-#      - acceptance:
-#          requires:
-#            - unit-helm2
-#            - unit-helm3
-#            - unit-acceptance-framework
+      - acceptance:
+          requires:
+            - unit-helm2
+            - unit-helm3
+            - unit-acceptance-framework
       - acceptance-openshift:
           requires:
             - unit-helm2
             - unit-helm3
             - unit-acceptance-framework
-# todo: uncomment before merging the PR
-#  nightly-acceptance-tests:
-#    triggers:
-#      - schedule:
-#          cron: "0 0 * * *"
-#          filters:
-#            branches:
-#              only:
-#                - master
-#    jobs:
-#      - acceptance-openshift
+  nightly-acceptance-tests:
+    triggers:
+      - schedule:
+          cron: "0 0 * * *"
+          filters:
+            branches:
+              only:
+                - master
+    jobs:
+      - acceptance-openshift
   update-helm-charts-index:
     jobs:
       - update-helm-charts-index:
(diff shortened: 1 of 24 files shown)


test/acceptance/tests/terminating-gateway/terminating_gateway_namespaces_test.go
```diff
@@ -211,7 +211,7 @@ func TestTerminatingGatewayNamespaceMirroring(t *testing.T) {
 			}
 
 			// Create the config entry for the terminating gateway
-			createTerminatingGatewayConfigEntry(t, consulClient, ctx.KubectlOptions(t).Namespace, testNamespace)
+			createTerminatingGatewayConfigEntry(t, consulClient, "", testNamespace)
```
ishustava, line 214: This is a bug because we're always running acceptance tests in the default namespace. On openshift, we don't but we need to create the config entry in the default namespace in consul.

.circleci/config.yml
```diff
@@ -191,11 +243,27 @@ workflows:
             - go-fmt-and-vet
       - unit-helm2
       - unit-helm3
-      - acceptance:
+#      - acceptance:
```
ishustava, line 246: todo: uncomment before merging

.circleci/config.yml
```diff
@@ -153,6 +153,58 @@ jobs:
             terraform destroy -var project=${CLOUDSDK_CORE_PROJECT} -auto-approve
           when: always
 
+  acceptance-openshift:
+    environment:
+      - TEST_RESULTS: /tmp/test-results
+      - OC_PRIMARY_NAME: consul-helm-test-2757871175
+      - OC_SECONDARY_NAME: consul-helm-test-3737660519
```
lkysow, line 160: What are these? Can we add a comment

lkysow (approved):
🎉 🦊  Looks simple enough 😆

.circleci/config.yml
ishustava: These are the pre-created openshift cluster names. Will add!

kschoche (approved):
Great work @ishustava!!
One quick question about the comments in values.yaml : 
Does it make sense to add a comment saying 
"this has been tested on openshift version X?"
so that users have an idea of what a supportable version would be?

ishustava:
> Does it make sense to add a comment saying
"this has been tested on openshift version X?"

yeah! I think that's a great idea!

PR #603, opened 2020-09-14
Giving the possibility to bring own client cert

zonArt:


david-yu:
@zonArt Thanks for the PR. Could you tell us why you chose to bring your own client cert?

lkysow:
@david-yu I think this comment explains it: https://github.com/hashicorp/consul-helm/pull/596#issuecomment-691939696
> I'm actually still having an issue with this approach yes. The issue I have is that I cannot enable tls because of the initContainer which is then created and fails each time. May I open another PR with only the flag "bringOwnKey" so that I can get rid of the initContainer but still be able to benefice from exposing port 8501 when choosing to enable tls?

zonArt:
I'm not sure to understand your suggestion to use enableAutoEncrypt as this also requires to mount a secret which I don't want and it fails if called and secret not mounted, thus this PR to get rid of the init-container but still being able to enable tls

lkysow:
So we've discussed this more internally and unfortunately this isn't something we want to support at this time. We're planning on eventually deprecating support for any TLS method other than auto-encrypt and so we don't want to support a third method (bring-your-own-client-certs) in addition to the two methods we already support (autoencrypt, non-autoencrypt with autogenerated client certs).

> I'm not sure to understand your suggestion to use enableAutoEncrypt as this also requires to mount a secret which I don't want and it fails if called and secret not mounted, thus this PR to get rid of the init-container and the mount of different secret which we don't care but still being able to enable tls

The secret it's complaining about is the CA cert (not the private key) which it needs to be able to make calls to the Consul servers and verify the connection.

lkysow:
I've created https://github.com/hashicorp/consul-helm/issues/623 to track interest. If that gets a lot of 👍 reactions then we may revisit this decision.

t-eckert:

PR #607, opened 2020-09-16
acceptance-tests: fix flakey mesh gateway test

ishustava:
Previously, we were checking that the federation is successful
by only looking for the number of WAN consul members. However, sometimes
those members could be unhealthy/not alive, which will cause the test to fail.
This change improves federation verification and checks that all members are
healthy from the perspective of both servers. It also checks that the ACL
replication is running in case ACLs are used.

### DIFF
--- a/test/acceptance/framework/consul_cluster.go
+++ b/test/acceptance/framework/consul_cluster.go
@@ -3,8 +3,6 @@ package framework
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
-	"os"
 	"strings"
 	"testing"
 
@@ -15,6 +13,7 @@ import (
 	"github.com/hashicorp/consul/api"
 	"github.com/hashicorp/consul/sdk/freeport"
 	"github.com/stretchr/testify/require"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
@@ -144,29 +143,25 @@ func (h *HelmCluster) SetupConsulClient(t *testing.T, secure bool) *api.Client {
 	remotePort := 8500 // use non-secure by default
 
 	if secure {
-		// overwrite remote port to HTTPS
+		// Overwrite remote port to HTTPS.
 		remotePort = 8501
 
-		// get the CA
-		caSecret, err := h.kubernetesClient.CoreV1().Secrets(namespace).Get(h.releaseName+"-consul-ca-cert", metav1.GetOptions{})
-		require.NoError(t, err)
-		caFile, err := ioutil.TempFile("", "")
-		require.NoError(t, err)
-		helpers.Cleanup(t, h.noCleanupOnFailure, func() {
-			require.NoError(t, os.Remove(caFile.Name()))
-		})
-
-		if caContents, ok := caSecret.Data["tls.crt"]; ok {
-			_, err := caFile.Write(caContents)
+		// Get the ACL token. First, attempt to read it from the bootstrap token.
+		// If the bootstrap token doesn't exist, try to read the replication token from the federation secret.
+		aclSecret, err := h.kubernetesClient.CoreV1().Secrets(namespace).Get(h.releaseName+"-consul-bootstrap-acl-token", metav1.GetOptions{})
+		if err != nil && errors.IsNotFound(err) {
+			federationSecret := fmt.Sprintf("%s-consul-federation", h.releaseName)
+			aclSecret, err = h.kubernetesClient.CoreV1().Secrets(namespace).Get(federationSecret, metav1.GetOptions{})
+			require.NoError(t, err)
+			config.Token = string(aclSecret.Data["replicationToken"])
+		} else if err == nil {
+			config.Token = string(aclSecret.Data["token"])
+		} else {
 			require.NoError(t, err)
 		}
 
-		// get the ACL token
-		aclSecret, err := h.kubernetesClient.CoreV1().Secrets(namespace).Get(h.releaseName+"-consul-bootstrap-acl-token", metav1.GetOptions{})
-		require.NoError(t, err)
-
-		config.TLSConfig.CAFile = caFile.Name()
-		config.Token = string(aclSecret.Data["token"])
+		// It's OK to skip TLS verification for local traffic.
+		config.TLSConfig.InsecureSkipVerify = true
 		config.Scheme = "https"
 	}
 
--- a/test/acceptance/framework/consul_cluster.go
+++ b/test/acceptance/framework/consul_cluster.go
@@ -146,8 +146,15 @@ func (h *HelmCluster) SetupConsulClient(t *testing.T, secure bool) *api.Client {
 		// Overwrite remote port to HTTPS.
 		remotePort = 8501
 
-		// Get the ACL token. First, attempt to read it from the bootstrap token.
-		// If the bootstrap token doesn't exist, try to read the replication token from the federation secret.
+		// It's OK to skip TLS verification for local traffic.
+		config.TLSConfig.InsecureSkipVerify = true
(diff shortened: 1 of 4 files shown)


test/acceptance/framework/consul_cluster.go
```diff
@@ -144,29 +143,25 @@ func (h *HelmCluster) SetupConsulClient(t *testing.T, secure bool) *api.Client {
...
-		caFile, err := ioutil.TempFile("", "")
-		require.NoError(t, err)
-		helpers.Cleanup(t, h.noCleanupOnFailure, func() {
-			require.NoError(t, os.Remove(caFile.Name()))
-		})
-
-		if caContents, ok := caSecret.Data["tls.crt"]; ok {
-			_, err := caFile.Write(caContents)
+		// Get the ACL token. First, attempt to read it from the bootstrap token.
+		// If the bootstrap token doesn't exist, try to read the replication token from the federation secret.
```
ashwin-venkatesh, line 150: Can we elaborate this comment a little bit as to why the token might not exists and potentially why it is ok for it to not be present.

thisisnotashwin (approved):
Looks great. Im so excited that it will not be flaky anymore!! 🏅

test/acceptance/framework/consul_cluster.go
```diff
@@ -144,29 +143,25 @@ func (h *HelmCluster) SetupConsulClient(t *testing.T, secure bool) *api.Client {
...
 
-		// get the ACL token
-		aclSecret, err := h.kubernetesClient.CoreV1().Secrets(namespace).Get(h.releaseName+"-consul-bootstrap-acl-token", metav1.GetOptions{})
-		require.NoError(t, err)
-
-		config.TLSConfig.CAFile = caFile.Name()
-		config.Token = string(aclSecret.Data["token"])
+		// It's OK to skip TLS verification for local traffic.
+		config.TLSConfig.InsecureSkipVerify = true
 		config.Scheme = "https"
```
lkysow, line 165: These lines make more sense to me to be grouped with `remotePort = 8501`

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
```diff
@@ -217,3 +217,33 @@ func TestMeshGatewaySecure(t *testing.T) {
 		})
 	}
 }
+
+const consulMemberStatusAlive = 1
```
lkysow, line 221: ... why 😂

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
```diff
@@ -217,3 +217,33 @@ func TestMeshGatewaySecure(t *testing.T) {
...
+
+// verifyFederation checks that the WAN federation between servers is successful
+// by first checking members are alive from the perspective of both servers.
+// If secure is true, it will also check that the ACL replication is running on the secondary server.
+func verifyFederation(t *testing.T, primaryClient, secondaryClient *api.Client, secure bool) {
```
lkysow, line 226: Love this.

lkysow (approved):
Constant improvement!! 🎉

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
ishustava: It's for @ashwin-venkatesh 😄 

But seriously, there are 4 integers values for member status, and it wasn't super clear to me which one is which (is 0 good or bad?). So this is more for clarity also around consul api, which doesn't have them as an enum.

test/acceptance/framework/consul_cluster.go
ishustava: Yes, will do. good call!

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
ishustava: I should prob move it inside the function though. It doesn't need to live here.

ashwin-venkatesh: The shade I receive because I ask for a few constants every now and then 😛

test/acceptance/framework/consul_cluster.go
ishustava: I've moved that whole section to be together with remotePort.

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
ishustava: I found out that these constants exist in serf, so I switched it to use `serf.StatusAlive`. That way we get the best of both worlds.

lkysow: No no! :D I like the constant. I was saying why is this an int in consul's API instead of a bool. But now I hear it's an enum so I guess that makes sense, although really a string is better IMO.

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
```diff
@@ -3,10 +3,13 @@ package meshgateway
 import (
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/consul-helm/test/acceptance/framework"
 	"github.com/hashicorp/consul-helm/test/acceptance/helpers"
 	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/serf/serf"
+	"github.com/hashicorp/serf/testutil/retry"
```
lkysow, line 12: Using serf's retry here

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
```diff
@@ -3,10 +3,13 @@ package meshgateway
 import (
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/consul-helm/test/acceptance/framework"
 	"github.com/hashicorp/consul-helm/test/acceptance/helpers"
 	"github.com/hashicorp/consul/api"
+	"github.com/hashicorp/serf/serf"
```
lkysow, line 11: I don't think we need the serf dep. It's being brought in accidentally in namespaces_test.

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
```diff
@@ -217,3 +218,31 @@ func TestMeshGatewaySecure(t *testing.T) {
...
+// by first checking members are alive from the perspective of both servers.
+// If secure is true, it will also check that the ACL replication is running on the secondary server.
+func verifyFederation(t *testing.T, primaryClient, secondaryClient *api.Client, secure bool) {
+	retrier := &retry.Timer{Timeout: 1 * time.Minute, Wait: 1 * time.Second}
+
+	retry.RunWith(retrier, t, func(r *retry.R) {
+		members, err := primaryClient.Agent().Members(true)
+		require.NoError(r, err)
+		require.Len(r, members, 2)
+		require.Equal(r, members[0].Status, serf.StatusAlive)
```
lkysow, line 232: I think go back to the const and fix namespaces_test and then we can ditch the serf dep.

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
ishustava: Oh, you're right. I was surprised when I brought in the serf const why go.mod wasn't modified! Let me correct!

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
ishustava: Done!

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
ishustava: Agree. I thought since we had it then why not, but it looks like we had it accidentally 😄

PR #609, opened 2020-09-18
[connectInject|syncCatalog] - Add variables for priorityClassName

Lord-Y:
Add variables for priorityClassName and add documentation regarding replicas/resources per ingress-gateway.

Signed-off-by: Lord-Y Lord-Y@users.noreply.github.com

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -1269,6 +1269,16 @@ ingressGateways:
   # case of annotations where both will be applied.
   gateways:
     - name: ingress-gateway
+      # Possibility to override default replicas per gateway
+      # replicas: 3
+      # Possibility to override default resources per gateway
+      # resources:
+      #   requests:
+      #     memory: "100Mi"
+      #     cpu: "100m"
+      #   limits:
+      #     memory: "500Mi"
+      #     cpu: "500m"
 
 # Configuration options for terminating gateways. Default values for all
 # terminating gateways are defined in `terminatingGateways.defaults`. Any of
--- a/values.yaml
+++ b/values.yaml
@@ -1269,16 +1269,6 @@ ingressGateways:
   # case of annotations where both will be applied.
   gateways:
     - name: ingress-gateway
-      # Possibility to override default replicas per gateway
-      # replicas: 3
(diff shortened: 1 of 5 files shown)


Lord-Y:
@ishustava here is my new clean PR :)

Lord-Y:
any news @ishustava @lkysow

values.yaml
```diff
@@ -1263,6 +1269,16 @@ ingressGateways:
...
+      # Possibility to override default replicas per gateway
+      # replicas: 3
+      # Possibility to override default resources per gateway
+      # resources:
+      #   requests:
+      #     memory: "100Mi"
+      #     cpu: "100m"
+      #   limits:
+      #     memory: "500Mi"
+      #     cpu: "500m"
```
ishustava, line 1281: This looks like it got here by accident 😃

ishustava (commented):
Looks good!! Tiny comment about what looks like an extra comment in the ingress gateway section, otherwise this looks good to me.

values.yaml
Lord-Y: Not by accident @ishustava . I put on purpose because I'm sure some people won't dig into the `template` folder or `test/units` folder like I did :-) to see how it works.

Lord-Y: @ishustava anything blocking the merge?

lkysow: Hi, this is documented above (https://github.com/hashicorp/consul-helm/pull/609/files#diff-ba05023d0f1b5d4b64b6cad4cf9a7ecdR1160-R1167). I don't think it makes sense to document this here because we'd need to have a commented out config for all the possible settings which would be a lot of comments, this would also need to be done for `terminatingGateways` and finally we don't use commented-out config as documentation anywhere else in our chart so this also breaks that consistency.

Based on those reasons can you please remove this config.

lkysow (changes requested):

values.yaml
Lord-Y: @lkysow thx for your feedback. I removed the extra comments.

lkysow (approved):
Thanks!!

PR #611, opened 2020-09-22
Fix some acceptance tests bugs

ishustava:
Changes proposed in this PR:
* Fix a bug where if the namespace is set in your kube context, the tests will ignore it
* Update client-go, terratest, and consul versions


test/acceptance/framework/consul_cluster.go
```diff
@@ -103,25 +105,29 @@ func (h *HelmCluster) Destroy(t *testing.T) {
 	helm.Delete(t, h.helmOptions, h.releaseName, false)
 
 	// delete PVCs
-	h.kubernetesClient.CoreV1().PersistentVolumeClaims(h.helmOptions.KubectlOptions.Namespace).DeleteCollection(&metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: "release=" + h.releaseName})
+	h.kubernetesClient.CoreV1().PersistentVolumeClaims(h.helmOptions.KubectlOptions.Namespace).DeleteCollection(context.TODO(), metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: "release=" + h.releaseName})
```
lkysow, line 108: will we ever be able to pass in a ctx here? If not then we could use .background

lkysow (approved):

test/acceptance/framework/environment.go
```diff
@@ -70,25 +71,48 @@ type kubernetesContext struct {
...
+		require.NoError(t, err)
+
+		contextName := helpers.KubernetesContextFromOptions(t, k.options)
+		if rawConfig.Contexts[contextName].Namespace != "" {
+			k.options.Namespace = rawConfig.Contexts[contextName].Namespace
+		} else {
+			k.options.Namespace = metav1.NamespaceDefault
+		}
+	}
+	return k.options
```
ashwin-venkatesh, line 107: 🏅

test/acceptance/go.mod
```diff
@@ -3,12 +3,13 @@ module github.com/hashicorp/consul-helm/test/acceptance
 go 1.14
 
 require (
-	github.com/gruntwork-io/terratest v0.27.3
-	github.com/hashicorp/consul/api v1.5.0
-	github.com/hashicorp/consul/sdk v0.5.0
+	github.com/gruntwork-io/terratest v0.29.0
+	github.com/hashicorp/consul/api v1.6.0
+	github.com/hashicorp/consul/sdk v0.6.0
+	github.com/hashicorp/serf v0.9.3
```
ashwin-venkatesh, line 9: What required a dependency on `serf` out of curiosity?

thisisnotashwin (approved):
Have a question about the addition about the serf dependency but that's about it.

This looks great

test/acceptance/framework/consul_cluster.go
ashwin-venkatesh: +1 to using `context.Background()` in tests. From the context docs:

```
// Background returns a non-nil, empty Context. It is never canceled, has no
// values, and has no deadline. It is typically used by the main function,
// initialization, and tests, and as the top-level Context for incoming
// requests.
```

test/acceptance/go.mod
ishustava: Good catch! *shakes fist at git rebase* I've rebased this branch with master and this retry keeps haunting me!

The answer is we never need a serf dependency 😄

test/acceptance/framework/consul_cluster.go
ishustava: I don't think we need to pass ctx at the moment. Let me switch to background instead, thank you both!

test/acceptance/go.mod
ashwin-venkatesh: lol. yeah "rebase reintroducing things i fixed" has bitten me many a time.

PR #615, opened 2020-09-24
Health checks base

kschoche:
Adding support for health checks controller for connectInject.

currently ACL support is not fully functional.


values.yaml
```diff
@@ -802,6 +802,17 @@ connectInject:
   image: null # image for consul-k8s that contains the injector
   default: false # true will inject by default, otherwise requires annotation
 
+  # Health checks enables propogation of k8s pod health status to consul service instances
```
lkysow, line 805:
```suggestion
  # Health checks enables propagation of k8s pod health status to Consul service instances.
```

values.yaml
```diff
@@ -802,6 +802,17 @@ connectInject:
...
+  # this has the effect of marking a Consul service instance passing/critical in the case
+  # that the connect-injected pod has passed/failed health checks.
+  # Setting this to false will cause the health checks controller to not start.
+  # Toggling this on/off is not recommended as it may leave stale Consul health checks that
+  # will eventually expire and be marked critical which would require manual cleanup or
+  # restarting the pod.
+  # NOTE: it is highly recommended to enable TLS with this feature.
```
lkysow, line 812:
```suggestion
  # This has the effect of marking a Consul service instance passing/critical in the case
  # that the connect-injected pod has passed/failed health checks.
  # Setting this to false will cause the health check status to not be synced.
  # Toggling this on/off is not recommended as it may leave stale Consul health checks that
  # will eventually expire and be marked critical which would require manual cleanup or
  # restarting the pod.
  # NOTE: It is highly recommended to enable TLS with this feature because it requires
  # making calls to Consul clients across the cluster. Without TLS enabled, these calls
  # could leak ACL tokens.
```

templates/connect-inject-deployment.yaml
```diff
@@ -84,6 +84,7 @@ spec:
                 -envoy-image="{{ .Values.global.imageEnvoy }}" \
                 -consul-k8s-image="{{ default .Values.global.imageK8S .Values.connectInject.image }}" \
                 -listen=:8080 \
+                -enable-health-checks="{{ .Values.connectInject.healthChecks.enabled }}" \
```
lkysow, line 87: Should this be something like `-enable-health-check-sync` so it's not thought to be enabling health checks on the injector itself? Or is that unnecessary?
cc @hashicorp/consul-eco-platform

templates/connect-inject-deployment.yaml
```diff
@@ -248,7 +249,8 @@ spec:
           - |
             consul-k8s acl-init \
               -secret-name="{{ template "consul.fullname" . }}-connect-inject-acl-token" \
-              -k8s-namespace={{ .Release.Namespace }}
+              -k8s-namespace={{ .Release.Namespace }} \
+              -enable-health-checks="{{ .Values.connectInject.healthChecks.enabled }}"
```
lkysow, line 253: need test that:
* flag defaults to true
* flag can be set to false

templates/connect-inject-clusterrole.yaml
```diff
@@ -17,6 +17,14 @@ rules:
     - "list"
     - "watch"
     - "patch"
+{{- if .Values.connectInject.healthChecks.enabled }}
```
lkysow, line 20: This needs a test too

templates/connect-inject-deployment.yaml
ishustava: That makes sense to me. Another option is to call it `-enable-health-checks-controller`.

kschoche: I like `-enable-health-checks-controller`, that is a great one

templates/connect-inject-deployment.yaml
```diff
@@ -248,7 +250,8 @@ spec:
           - |
             consul-k8s acl-init \
               -secret-name="{{ template "consul.fullname" . }}-connect-inject-acl-token" \
-              -k8s-namespace={{ .Release.Namespace }}
+              -k8s-namespace={{ .Release.Namespace }} \
+              -enable-health-checks={{ .Values.connectInject.healthChecks.enabled }}
```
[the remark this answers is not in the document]
kschoche: I'm using this to update the acl rules for injectConnect in consul-k8s, although right now it is not functional and this might be part of it.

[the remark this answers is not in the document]
ishustava: This command doesn't and shouldn't update acl rules; it just waits for a kube secret with the provided name to be created. You would need to pass it to the `server-acl-init` in `server-acl-init` job to change the rules.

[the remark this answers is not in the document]
kschoche: Ah! And this is exactly why my changes to server-acl-init are not getting called ;-)
Good catch here, sorry I hadnt started to dig into why ACLs werent working until maybe an hour ago.
I'll move this!

values.yaml
```diff
@@ -802,6 +802,22 @@ connectInject:
...
+  # that the connect-injected pod has passed/failed health checks.
+  # Setting this to false will cause the health check status to not be synced.
+  # Toggling this on/off is not recommended as it may leave stale Consul health checks that
+  # will eventually expire and be marked critical which would require manual cleanup or
+  # restarting the pod.
+  # reconcilePeriod defines how often a full state reconcile is done after the initial reconcile
+  # at startup is completed.
+  # NOTE: It is highly recommended to enable TLS with this feature because it requires
+  # making calls to Consul clients across the cluster. Without TLS enabled, these calls
+  # could leak ACL tokens.
```
ashwin-venkatesh, line 816: This could probably need a little more clarity. I think we should mention that the feature would require making HTTP requests across nodes, and if unencrypted, will expose the ACLs in plaintext. Well something to that effect.

.circleci/config.yml
```diff
@@ -139,7 +139,7 @@ jobs:
               -kubeconfig="$primary_kubeconfig" \
               -secondary-kubeconfig="$secondary_kubeconfig" \
               -debug-directory="$TEST_RESULTS/debug" \
-              -consul-k8s-image=hashicorpdev/consul-k8s:latest
+              -consul-k8s-image=kschoche/consul-k8s-dev
```
lkysow, line 142: If you want to test this in CI faster: `-run TestHealthChecks`

values.yaml
```diff
@@ -802,6 +802,22 @@ connectInject:
   image: null # image for consul-k8s that contains the injector
   default: false # true will inject by default, otherwise requires annotation
 
+  # Health checks enables propagation of k8s pod health status to Consul service instances.
+  # This has the effect of marking a Consul service instance passing/critical in the case
+  # that the connect-injected pod has passed/failed health checks.
+  # Setting this to false will cause the health check status to not be synced.
+  # Toggling this on/off is not recommended as it may leave stale Consul health checks that
+  # will eventually expire and be marked critical which would require manual cleanup or
```
ishustava, line 810: I'm not sure it's worth calling it out here since it implies that you won't restart your consul clients for 11yrs (or so), which I think is pretty unlikely.

values.yaml
```diff
@@ -802,6 +802,22 @@ connectInject:
...
+  # restarting the pod.
+  # reconcilePeriod defines how often a full state reconcile is done after the initial reconcile
+  # at startup is completed.
```
ishustava, line 813: could we move this to be above the reconcilePeriod property?

test/unit/connect-inject-deployment.bats
```diff
@@ -79,6 +79,47 @@ load _helpers
...
+      --set 'connectInject.enabled=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command' | tee /dev/stderr)
+
+  local actual=$(echo "$cmd" |
+    yq 'any(contains("-enable-health-checks-controller=false"))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
+@test "connectInject/Deployment: health checks sync period set by default" {
```
ishustava, line 95:
```suggestion
@test "connectInject/Deployment: health checks reconcile period set by default" {
```

test/unit/connect-inject-deployment.bats
```diff
@@ -79,6 +79,47 @@ load _helpers
...
+  cd `chart_dir`
+  local cmd=$(helm template \
+      -s templates/connect-inject-deployment.yaml \
+      --set 'connectInject.enabled=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command' | tee /dev/stderr)
+
+  local actual=$(echo "$cmd" |
+    yq 'any(contains("-health-checks-reconcile-period"))' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
```
ishustava, line 105: could you add an assertion here that `-enable-health-checks-controller=true`? or maybe in a separate test based on the name of this one?

test/unit/connect-inject-deployment.bats
```diff
@@ -870,6 +911,28 @@ load _helpers
   [ "${actual}" = "true" ]
 }
 
+@test "connectInject/Deployment: CONSUL_HTTP_ADDR env variable not set when healthchecks are not enabled" {
```
ishustava, line 914: removing the double negative 😃 
```suggestion
@test "connectInject/Deployment: CONSUL_HTTP_ADDR env variable not set when health checks are disabled" {
```

ishustava (commented):
Looking pretty good! I've left some edits and comments, but overall there wasn't anything major. Great work!!

test/acceptance/tests/connect/connect_inject_test.go
```diff
@@ -32,6 +32,7 @@ func TestConnectInject(t *testing.T) {
 			ctx := suite.Environment().DefaultContext(t)
 
 			helmValues := map[string]string{
+				"global.imageK8S":              "kschoche/consul-k8s-dev",
```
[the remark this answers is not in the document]
kschoche: good catch, thx!

values.yaml
```diff
@@ -802,6 +802,22 @@ connectInject:
...
+  # NOTE: It is highly recommended to enable TLS with this feature because it requires
+  # making calls to Consul clients across the cluster. Without TLS enabled, these calls
+  # could leak ACL tokens.
+  healthChecks:
+    enabled: false
+    reconcilePeriod: "10m"
```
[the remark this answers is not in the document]
kschoche: good catch, I meant to have that say 1m :D

test/acceptance/tests/fixtures/bases/static-client/deployment.yaml
```diff
@@ -18,4 +18,4 @@ spec:
           image: tutum/curl:latest
           command: [ "/bin/sh", "-c", "--" ]
           args: [ "while true; do sleep 30; done;" ]
-      serviceAccountName: static-client
\ No newline at end of file
+      serviceAccountName: static-client
```
kschoche, line 21:
```suggestion
      serviceAccountName: static-client```

File: test/acceptance/tests/fixtures/cases/static-server-hc/patch.yaml [hunk 17]
```diff
@@ -0,0 +1,26 @@
...
+  name: static-server
+spec:
+  template:
+    metadata:
+      annotations:
+        "consul.hashicorp.com/connect-inject": "true"
+    spec:
+      containers:
+        - name: static-server
+          image: kschoche/http-echo:latest
```
kschoche, line 13: TODO: do not use my private alpine build

test/unit/connect-inject-deployment.bats
kschoche: the default value for `-enable-health-checks-controller` is `false` so I'm not sure that it would be the right test here. Surely when we do set it enabled by default it would make sense to add this as a test!
Thoughts?

test/acceptance/tests/connect/health_checks_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+			releaseName := helpers.RandomName()
+			consulCluster := framework.NewHelmCluster(t, helmValues, ctx, cfg, releaseName)
+			consulCluster.Create(t)
+
+			t.Log("creating static-server and static-client deployments")
+			helpers.DeployKustomize(t, ctx.KubectlOptions(t), cfg.NoCleanupOnFailure, cfg.DebugDirectory, "../fixtures/cases/static-server-hc")
+			helpers.DeployKustomize(t, ctx.KubectlOptions(t), cfg.NoCleanupOnFailure, cfg.DebugDirectory, "../fixtures/cases/static-client-inject")
+			t.Log("checking that connection is successful")
+			helpers.CheckStaticServerConnection(t, ctx.KubectlOptions(t), true, staticClientName, "http://localhost:1234")
```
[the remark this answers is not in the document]
kschoche: I agree this is a great plan forward.
I did a bit of investigation into how this is done and pulled together a POC that works using the existing `cluster.SetupConsulClient` functionality. I got it to work but unfortunately got pretty messy with having to copy/paste functions from consul-k8s so we can generate check-ids and such. It also wouldnt work with TLS support because it appears to expect ACL as well, which health checks implementation wont support. I had to add some hooks as well to be able to change the destination client address because it is expecting remote connections.

I think it'd be a good TODO, which I'll add in the code, to add going forward to sanity check the current code, but I do think the current code is testing the end to end functionality. Thoughts?

If you'd like I can hop on a zoom and show you my findings.

test/unit/connect-inject-deployment.bats
ishustava: My comment I think is more around the fact that there aren't any tests that check that if `connectInject.healthChecks.enabled` is `true`, then there is `-enable-health-checks-controller=true`.

test/acceptance/tests/connect/health_checks_test.go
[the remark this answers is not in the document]
ishustava: Sounds good to leave as a TODO!

test/unit/connect-inject-deployment.bats
kschoche: Ah, great point, let me add that! thank you

test/acceptance/tests/connect/health_checks_test.go
```diff
@@ -0,0 +1,47 @@
...
+// 1. Test that health checks enabled by default works for passing+failing
+// 2. Test that health checks enabled with TLS/ACL works for passing+failing
+
+// Test that health checks work in a default installation
+// Workflow is as follows :
+// Deploy with a failing health check
+// Test that traffic fails
+// kubectl exec to the static-client and `touch /tmp/healthy`
+// Test that traffic passes
+func TestHealthChecksDefault(t *testing.T) {
```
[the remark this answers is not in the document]
kschoche: I made a couple changes here regarding deploying with a passing check so we do not need the new DeployKustomizeWithoutWait functionality. I've also listed as a TODO the consul health check validation.

test/acceptance/tests/connect/health_checks_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/hashicorp/consul-helm/test/acceptance/helpers"
+	"strconv"
+	"testing"
+)
+
+// Test that health checks work in a default installation and a secure installation with TLS/auto-encrypt permutations.
+// Deploy with a passing health check
+// Test that traffic passes
+// update the container with readiness probe so that it fails
+// Test that traffic now fails
```
ashwin-venkatesh, line 15:
```suggestion
// Deploy with a passing health check.
// Test that the service is accessible over the mesh.
// Update the container with readiness probe so that it fails.
// Test that the service is inaccessible over the mesh.
```

values.yaml
```diff
@@ -802,6 +802,19 @@ connectInject:
...
   default: false # true will inject by default, otherwise requires annotation
 
+  # healthChecks enables propagation of k8s pod health status to Consul service instances.
+  # NOTE: It is highly recommended to enable TLS with this feature because it requires
+  # making calls to Consul clients across the cluster. Without TLS enabled, these calls
+  # could leak ACL tokens.
+  healthChecks:
+    # true has the effect of marking a Consul service instance passing/critical in the event
+    # that the connect-injected pod has passed/failed readiness checks.
+    # Setting this to false will cause the controller to not run and health check status to not be synced.
```
ashwin-venkatesh, line 812:
```suggestion
    # enables the Consul Health Check controller that syncs the readiness checks on
    # connect-injected services with the agent, to route traffic only to healthy pods.
```

ashwin-venkatesh (approved):
This looks good outside of the comments left by @ishustava !!

values.yaml
ishustava: This suggestion looks great! nit: could we capitalize `enables` plz. Also, let's remove the last clause: `, to route traffic only to healthy pods` since it makes it sound that the hc controller routes traffic.

ishustava (approved):
🎉

values.yaml
ashwin-venkatesh: 💯

PR #616, opened 2020-09-24
fix bats tests for hc-base

kschoche:
bats test was failing due to indexing on some tests because we've modified the clusterrole for connect-inject, this resolves it


ishustava:
@kschoche looks like these tests are failing in CI. I'll wait for them to be fixed before reviewing.

kschoche:
> @kschoche looks like these tests are failing in CI. I'll wait for them to be fixed before reviewing.
good catch it was on the wrong base branch!

test/unit/connect-inject-clusterrole.bats
```diff
@@ -59,7 +59,7 @@ load _helpers
       --set 'global.enablePodSecurityPolicies=false' \
       . | tee /dev/stderr |
       yq -r '.rules | length' | tee /dev/stderr)
-  [ "${actual}" = "1" ]
+  [ "${actual}" = "2" ]
```
ishustava, line 62: could we use map/select for these tests? I think it'll both make it more resilient against future changes and more readable.

kschoche: sure thing, great idea!

PR #626, opened 2020-09-30
Add datacenter flag to controller deployment

thisisnotashwin:
This is a companion PR to https://github.com/hashicorp/consul-k8s/pull/343 which passes the datacenter flag to the controller.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -140,7 +140,8 @@ jobs:
               -kubeconfig="$primary_kubeconfig" \
               -secondary-kubeconfig="$secondary_kubeconfig" \
               -debug-directory="$TEST_RESULTS/debug" \
-              -consul-k8s-image=hashicorpdev/consul-k8s:crd-controller-base-latest
+              -consul-k8s-image=ashwinvenkatesh/consul-k8s:meta@sha256:531a8865c4c585b53962bef8ff7e7df375741b16ee4f012d53f348deb30ba5f9 \
+              -consul-image=ashwinvenkatesh/consul:meta@sha256:ba215ab440990e19e87f5fa0952f981dcbf2605175b97899dfec259ee1abeca5
 
       - store_test_results:
           path: /tmp/test-results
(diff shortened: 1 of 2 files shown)


templates/controller-deployment.yaml
```diff
@@ -58,6 +58,7 @@ spec:
         - |
           consul-k8s controller \
             --webhook-tls-cert-dir=/tmp/controller-webhook/certs \
+            --datacenter={{ .Values.global.datacenter }}
```
lkysow, line 61: does this take single `-`'s?

lkysow (approved):

templates/controller-deployment.yaml
ashwin-venkatesh: I think they do. Want me to update all flags with a single `-`?

.circleci/config.yml
```diff
@@ -140,7 +140,8 @@ jobs:
               -kubeconfig="$primary_kubeconfig" \
               -secondary-kubeconfig="$secondary_kubeconfig" \
               -debug-directory="$TEST_RESULTS/debug" \
-              -consul-k8s-image=hashicorpdev/consul-k8s:crd-controller-base-latest
+              -consul-k8s-image=ashwinvenkatesh/consul-k8s:meta@sha256:531a8865c4c585b53962bef8ff7e7df375741b16ee4f012d53f348deb30ba5f9 \
+              -consul-image=ashwinvenkatesh/consul:meta@sha256:ba215ab440990e19e87f5fa0952f981dcbf2605175b97899dfec259ee1abeca5
```
ashwin-venkatesh, line 144: These will be reset before the code gets merged

templates/controller-deployment.yaml
kschoche: I think that makes sense to do at some point

ashwin-venkatesh: how about now? 😉

kschoche (approved):
💥 🧨 Looks great!

lkysow:
@ashwin-venkatesh we need this merged I think for the tests to pass now cuz we've made that flag mandatory.

PR #627, opened 2020-10-01
Config for controller

lkysow:


thisisnotashwin (approved):
🦊

kschoche (approved):
great work here! as always curious how we come up with resource values but we know how that will sort itself out :-)

lkysow:
> great work here! as always curious how we come up with resource values but we know how that will sort itself out :-)

Sorry I should have made note that setting resources is a todo

PR #629, opened 2020-10-02
Support -log-level flag

lkysow:
Companion to https://github.com/hashicorp/consul-k8s/pull/348


lkysow:
These are just failing because it's setting `-log-level` and it's using the old image.

thisisnotashwin (approved):
🦊

PR #631, opened 2020-10-06
Update proxy-defaults config type to object

ashwin-venkatesh:
- It was previously updated to 'Any' with an updated to a newer version of codegen. The Any type seems to not treat inputs as expected.
- Updated test fixtures with config to prevent similar bugs in the future.

### DIFF
--- a/templates/crd-proxydefaults.yaml
+++ b/templates/crd-proxydefaults.yaml
@@ -39,7 +39,7 @@ spec:
           properties:
             config:
               description: Config is an arbitrary map of configuration values used by Connect proxies. Any values that your proxy allows can be configured globally here. Supports JSON config values. See https://www.consul.io/docs/connect/proxies/envoy#configuration-formatting
-              type: Any
+              type: object
             expose:
               description: Expose controls the default expose path configuration for Envoy.
               properties:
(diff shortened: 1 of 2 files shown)


lkysow (approved):

.circleci/config.yml
```diff
@@ -140,7 +140,8 @@ jobs:
               -kubeconfig="$primary_kubeconfig" \
               -secondary-kubeconfig="$secondary_kubeconfig" \
               -debug-directory="$TEST_RESULTS/debug" \
-              -consul-k8s-image=hashicorpdev/consul-k8s:crd-controller-base-latest
+              -consul-k8s-image=hashicorpdev/consul-k8s:crd-controller-base-latest \
+              -consul-image=ashwinvenkatesh/consul:main-oct6@sha256:f34ca59e93c564082a18e18850d6dd8c6ac43b97afbe9bd9183ac4aab15c70b1
```
ashwin-venkatesh, line 144: We need to provide our own image here as we rely on features that are currently on the master branch of `consul`

ishustava (commented):
Looks good, pending green acceptance tests. For the tests, I think we need to set the consul image in each test that needs the latest image rather than providing the global flag. That way tests that are using enterprise will keep using the enterprise image.

lkysow:
> Looks good, pending green acceptance tests. For the tests, I think we need to set the consul image in each test that needs the latest image rather than providing the global flag. That way tests that are using enterprise will keep using the enterprise image.

Totally, I switched that back in another PR I'm pretty sure.

ishustava (approved):
🍏 🎉

PR #632, opened 2020-10-06
Install CRDs only if controller.enabled

lkysow:
Since we're cutting beta off of master, we don't want to install CRDs unless the controller is enabled.

This also runs our acceptance tests off of `hashicorpdev/consul` for OSS and `hashicorp/consul-enterprise:1.8.4-ent` for enterprise. This gets us support for metadata on config entries which we need.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -141,7 +141,6 @@ jobs:
               -secondary-kubeconfig="$secondary_kubeconfig" \
               -debug-directory="$TEST_RESULTS/debug" \
               -consul-k8s-image=hashicorpdev/consul-k8s:crd-controller-base-latest \
-              -consul-image=hashicorpdev/consul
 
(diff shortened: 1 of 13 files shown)


thisisnotashwin (approved):
🦊

.circleci/config.yml
```diff
@@ -141,7 +141,6 @@ jobs:
               -secondary-kubeconfig="$secondary_kubeconfig" \
               -debug-directory="$TEST_RESULTS/debug" \
               -consul-k8s-image=hashicorpdev/consul-k8s:crd-controller-base-latest \
-              -consul-image=hashicorpdev/consul
```
lkysow, line 144: This was breaking our tests because it was setting the image for enterprise tests too (which need consul-ent)

kschoche (approved):

PR #634, opened 2020-10-06
Service Intentions L4 support

ashwin-venkatesh:
Add support for service intentions

- Skip enterprise tests as they don't currently have an image with service intentions

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -140,8 +140,7 @@ jobs:
               -kubeconfig="$primary_kubeconfig" \
               -secondary-kubeconfig="$secondary_kubeconfig" \
               -debug-directory="$TEST_RESULTS/debug" \
-              -consul-k8s-image=hashicorpdev/consul-k8s:crd-controller-base-latest \
-              -consul-image=hashicorpdev/consul
+              -consul-k8s-image=hashicorpdev/consul-k8s:crd-controller-base-latest
 
       - store_test_results:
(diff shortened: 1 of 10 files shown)


test/acceptance/tests/fixtures/crds/serviceintentions.yaml
```diff
@@ -0,0 +1,24 @@
...
+  name: svc1
+spec:
+  protocol: http
+---
+apiVersion: consul.hashicorp.com/v1alpha1
+kind: ServiceDefaults
+metadata:
+  name: svc2
+spec:
+  protocol: http
```
lkysow, line 13: This shoudn't require http should it? Since it should also work for L4

lkysow (commented):
👍 pending tests passing after go.mod is fixed

.circleci/config.yml
```diff
@@ -141,6 +141,7 @@ jobs:
               -secondary-kubeconfig="$secondary_kubeconfig" \
               -debug-directory="$TEST_RESULTS/debug" \
               -consul-k8s-image=hashicorpdev/consul-k8s:crd-controller-base-latest \
+              -consul-image=hashicorpdev/consul
```
lkysow, line 144: You need to delete this I think.

lkysow (approved):

ishustava (approved):
Looks great!

PR #638, opened 2020-10-08
Add annotation to prevent controller from getting connect injected

ashwin-venkatesh:
The controller should not get connect injected into the mesh


lkysow (approved):
What about cert gen

ashwin-venkatesh:
> What about cert gen

It was done already

ishustava (approved):
🎉

PR #648, opened 2020-10-15
Run enterprise license job for all tests

ishustava:
This is a followup to #654 

In addition to running tests with a license, update the enterprise license job to not suppress error output. Otherwise, our logs look like:

```
Applying license...
Retrying in 2s...
Applying license...
Retrying in 2s...
Applying license...
Retrying in 2s...
Applying license...
...
```
which makes it hard to debug errors. I've switched the order of `timeout` commands so that we try the newer version first (without the `-t` flag).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -144,6 +144,8 @@ jobs:
             do
               if ! gotestsum --no-summary=all --jsonfile=jsonfile-${pkg////-} -- $pkg -p 1 -timeout 30m -failfast \
                     -enable-enterprise \
+                    -enterprise-license-secret-name=ent-license \
+                    -enterprise-license-secret-key=key \
                     -enable-multi-cluster \
                     -kubeconfig="$primary_kubeconfig" \
                     -secondary-kubeconfig="$secondary_kubeconfig" \
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,7 +133,9 @@ jobs:
             eval "$(echo export primary_kubeconfig=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .kubeconfigs.value[0]))"
             eval "$(echo export secondary_kubeconfig=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .kubeconfigs.value[1]))"
 
-            # create an enterprise license secret in the primary cluster
+            # Create an enterprise license secret in the primary cluster.
+            # This license is set as a CircleCI project env variable.
+            # The license expires 15-Oct-2025.
             KUBECONFIG=$primary_kubeconfig kubectl create secret generic ent-license --from-literal=key="${CONSUL_ENT_LICENSE}"
 
             # We have to run the tests for each package separately so that we can
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,6 +133,9 @@ jobs:
             eval "$(echo export primary_kubeconfig=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .kubeconfigs.value[0]))"
             eval "$(echo export secondary_kubeconfig=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .kubeconfigs.value[1]))"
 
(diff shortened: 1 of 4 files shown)


test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
```diff
@@ -74,6 +74,8 @@ func TestMeshGatewayDefault(t *testing.T) {
 		"server.extraVolumes[0].load":          "true",
 		"server.extraVolumes[0].items[0].key":  "serverConfigJSON",
 		"server.extraVolumes[0].items[0].path": "config.json",
+		"server.enterpriseLicense.secretName":  "",
+		"server.enterpriseLicense.secretKey":   "",
```
ishustava, line 78: We need to do it on the secondary servers since the license should only be applied to the primary.

test/acceptance/tests/controller/controller_test.go
```diff
@@ -41,7 +41,7 @@ func TestController(t *testing.T) {
 				"controller.enabled":    "true",
 				"connectInject.enabled": "true",
 				// todo: remove when 1.9.0 is released.
-				"global.image": "hashicorpdev/consul",
+				"global.image": "hashicorp/consul-enterprise:1.9.0-ent-beta2",
```
ishustava, line 44: This was easier than setting the license values to `""`

.circleci/config.yml
```diff
@@ -133,6 +133,9 @@ jobs:
             eval "$(echo export primary_kubeconfig=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .kubeconfigs.value[0]))"
             eval "$(echo export secondary_kubeconfig=$(terraform output -state ../../terraform/gke/terraform.tfstate -json | jq -r .kubeconfigs.value[1]))"
 
+            # create an enterprise license secret in the primary cluster
+            KUBECONFIG=$primary_kubeconfig kubectl create secret generic ent-license --from-literal=key="${CONSUL_ENT_LICENSE}"
```
lkysow, line 137: Can you add a comment that this was configured through circle ci and when it's supposed to expire?

templates/enterprise-license-job.yaml
```diff
@@ -92,7 +92,7 @@ spec:
                 chmod +x ./apply-license.sh
 
                 # Time out after 20 minutes. Use || to support new timeout versions that don't accept -t
-                timeout -t 1200 ./apply-license.sh 2> /dev/null || timeout 1200 ./apply-license.sh 2> /dev/null
+                timeout 1200 ./apply-license.sh || timeout -t 1200 ./apply-license.sh
```
lkysow, line 95: We did this so you don't see timeout usage errors unnecessarily. Didn't realize it was swallowing the error output from `consul license put` though.

I guess the only fix to ensure the timeout usage errors are silenced would be to capture the output from `consul license put` and redirect it to /stdout?

test/acceptance/tests/controller/controller_test.go
lkysow: Sorry, don't understand this. Not a big deal though because this will all be removed with 1.9.0.

lkysow (approved):

test/acceptance/tests/controller/controller_test.go
ishustava: Sorry, this was confusing! This is because if we set ent license flags globally through test flags and use OSS image, the helm install will fail. One option is to keep OSS image and set ent license flags to `""`, another option is to do what I did. It was simpler to switch to the ent image since, as you said, it doesn't matter and it was less code to write 😄 .

templates/enterprise-license-job.yaml
ishustava: Ah, that's true! Let me try and redirect the output.

ndhanushkodi (approved):
Thanks for adding the comments about zeroing out the licenses in the secondary DC!!

templates/enterprise-license-job.yaml
ishustava: Output redirection is much better! I've switched to that.

PR #654, opened 2020-10-19
support new versions of timeout in ent-license

### LINKED ISSUE #647: enterprise license job fails on consul versions >= 1.8.1
Consul 1.8.1 upgraded alpine (https://github.com/hashicorp/docker-consul/pull/149) which included a newer version of `timeout` that takes the timeout as an argument instead of with the `-t` flag.

Before:
```
timeout
BusyBox v1.29.3 (2019-01-24 07:45:07 UTC) multi-call binary.

Usage: timeout [-t SECS] [-s SIG] PROG ARGS

Runs PROG. Sends SIG to it if it is not gone in SECS seconds.
Defaults: SECS: 10, SIG: TERM.
```

Now:
```
timeout -s
timeout: option requires an argument: s
BusyBox v1.31.1 () multi-call binary.

Usage: timeout [-s SIG] SECS PROG ARGS

Runs PROG. Sends SIG to it if it is not gone in SECS seconds.
Default SIG: TERM.
```

kschoche:
New versions of busybox packaged with consul-1.9 provide a version of `timeout` which does not support the `-t` flag, try it once with and once without if that fails.

I've tested this manually with 1.9.0-beta1 ent and verified it fails without the changes and passes with the changes, and continues to pass in 1.8.x versions.

Resolves #647

### DIFF
--- a/templates/enterprise-license-job.yaml
+++ b/templates/enterprise-license-job.yaml
@@ -92,7 +92,7 @@ spec:
                 chmod +x ./apply-license.sh
 
                 # Time out after 20 minutes. Use || to support new timeout versions that dont accept -t
-                timeout -t 1200 ./apply-license.sh || timeout 1200 ./apply-license.sh
+                timeout -t 1200 ./apply-license.sh 2> /dev/null || timeout 1200 ./apply-license.sh 2> /dev/null
           {{- if .Values.global.tls.enabled }}
           volumeMounts:
             - name: consul-ca-cert
--- a/templates/enterprise-license-job.yaml
+++ b/templates/enterprise-license-job.yaml
@@ -91,8 +91,8 @@ spec:
                 EOF
                 chmod +x ./apply-license.sh
 
-                # Time out after 20 minutes.
-                timeout -t 1200 ./apply-license.sh
+                # Time out after 20 minutes. Use || to support new timeout versions that dont accept -t
+                timeout -t 1200 ./apply-license.sh || timeout 1200 ./apply-license.sh
           {{- if .Values.global.tls.enabled }}
           volumeMounts:
             - name: consul-ca-cert
--- a/templates/enterprise-license-job.yaml
+++ b/templates/enterprise-license-job.yaml
@@ -91,7 +91,7 @@ spec:
                 EOF
                 chmod +x ./apply-license.sh
 
-                # Time out after 20 minutes. Use || to support new timeout versions that dont accept -t
+                # Time out after 20 minutes. Use || to support new timeout versions that don't accept -t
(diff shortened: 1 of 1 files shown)


templates/enterprise-license-job.yaml
```diff
@@ -91,8 +91,8 @@ spec:
                 EOF
                 chmod +x ./apply-license.sh
 
-                # Time out after 20 minutes.
-                timeout -t 1200 ./apply-license.sh
+                # Time out after 20 minutes. Use || to support new timeout versions that dont accept -t
+                timeout -t 1200 ./apply-license.sh || timeout 1200 ./apply-license.sh
```
lkysow, line 95: I think this might cause users confusion if the license apply fails because when they look at the logs, they'll read:
```
timeout: unrecognized option: t
BusyBox v1.31.1 () multi-call binary.

Usage: timeout [-s SIG] SECS PROG ARGS

Runs PROG. Sends SIG to it if it is not gone in SECS seconds.
Default SIG: TERM.
Applying license...
some-error
Retrying in 2s...
some-error
Retrying in 2s...
```

Can we suppress the usage output?

kschoche: How about this?
```suggestion
                timeout -t 1200 ./apply-license.sh &> /dev/null || timeout 1200 ./apply-license.sh
```
Example: 
```
ls -l /tmp/nonexistantfile &> /dev/null || ls -l /tmp/existantfile
-rw-r--r--  1 kyle  wheel  0 Oct 19 18:47 /tmp/existantfile
```

lkysow (approved):
🎉 I tested with the following:
```yaml
server:
  replicas: 1
  bootstrapExpect: 1
  enterpriseLicense:
    secretName: consul-ent-license
    secretKey: license
global:
  name: consul
  image: hashicorp/consul-enterprise:1.7.0-ent
```

```yaml
server:
  replicas: 1
  bootstrapExpect: 1
  enterpriseLicense:
    secretName: consul-ent-license
    secretKey: license
global:
  name: consul
  image: hashicorp/consul-enterprise:1.8.4-ent
```

templates/enterprise-license-job.yaml
```diff
@@ -91,8 +91,8 @@ spec:
                 EOF
                 chmod +x ./apply-license.sh
 
-                # Time out after 20 minutes.
-                timeout -t 1200 ./apply-license.sh
+                # Time out after 20 minutes. Use || to support new timeout versions that dont accept -t
```
ishustava, line 94:
```suggestion
                # Time out after 20 minutes. Use || to support new timeout versions that don't accept -t
```

ishustava (approved):
Looks great!

PR #655, opened 2020-10-19
Un-skip controller tests for enterprise

lkysow:
Don't test that the destination namespace is set by default. Instead,
set it explicitly. This requires moving to kustomize for
service-intentions resources.

We're doing this because the namespace defaulting logic is broken right now. However unskipping this test proves that the enterprise integration is working, except for the namespace defaulting logic.


test/acceptance/tests/fixtures/bases/crds/serviceintentions.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: consul.hashicorp.com/v1alpha1
+kind: ServiceIntentions
+metadata:
+  name: intentions
+spec:
+  destination:
+    name: svc1
+  sources:
+    - name: svc2
+      action: allow
```
lkysow, line 10: note: we don't need to set `sources[].namespace` because if this is empty, consul will set it to the value of `destination.namespace` (which we're setting through kustomize)

PR #658, opened 2020-10-21
Update MutatingWebHook API Version

lawliet89:
`admissionregistration.k8s.io/v1beta1` is [deprecated](https://v1-16.docs.kubernetes.io/docs/setup/release/notes/#deprecations-and-removals) since 1.16 and will be removed in 1.19.

This PR will set the right API Version if it is supported.

**However**, there is a change in behaviour from `v1beta1` to `v1`. Specifically, the default (unspecified) [`failurePolicy`](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy) changes from `Ignore` to `Fail`. 

The Connect Inject webhook does not have a default `failurePolicy` set whereas the Controller Webhook has it set to `Fail`. Should we set a value for the Connect Inject webhook?

### DIFF
--- a/templates/connect-inject-mutatingwebhook.yaml
+++ b/templates/connect-inject-mutatingwebhook.yaml
@@ -1,6 +1,6 @@
 {{- if (or (and (ne (.Values.connectInject.enabled | toString) "-") .Values.connectInject.enabled) (and (eq (.Values.connectInject.enabled | toString) "-") .Values.global.enabled)) }}
 # The MutatingWebhookConfiguration to enable the Connect injector.
-{{- if .Capabilities.APIVersions.Has "admissionregistration.k8s.io/v1" -}}
+{{- if .Capabilities.APIVersions.Has "admissionregistration.k8s.io/v1" }}
 apiVersion: admissionregistration.k8s.io/v1
 {{- else }}
 apiVersion: admissionregistration.k8s.io/v1beta1
@@ -16,6 +16,7 @@ metadata:
     release: {{ .Release.Name }}
 webhooks:
   - name: {{ template "consul.fullname" . }}-connect-injector.consul.hashicorp.com
+    failurePolicy: Fail
     clientConfig:
       service:
         name: {{ template "consul.fullname" . }}-connect-injector-svc
--- a/templates/connect-inject-mutatingwebhook.yaml
+++ b/templates/connect-inject-mutatingwebhook.yaml
@@ -1,6 +1,10 @@
 {{- if (or (and (ne (.Values.connectInject.enabled | toString) "-") .Values.connectInject.enabled) (and (eq (.Values.connectInject.enabled | toString) "-") .Values.global.enabled)) }}
 # The MutatingWebhookConfiguration to enable the Connect injector.
+{{- if .Capabilities.APIVersions.Has "admissionregistration.k8s.io/v1" -}}
+apiVersion: admissionregistration.k8s.io/v1
+{{- else }}
 apiVersion: admissionregistration.k8s.io/v1beta1
+{{- end }}
 kind: MutatingWebhookConfiguration
 metadata:
   name: {{ template "consul.fullname" . }}-connect-injector-cfg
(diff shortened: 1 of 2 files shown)


ashwin-venkatesh:
Hey @lawliet89 ! Thank you so much for the contribution. This looks great.

> The Connect Inject webhook does not have a default failurePolicy set whereas the Controller Webhook has it set to Fail. Should we set a value for the Connect Inject webhook?

Yes, please. That would be excellent!!

Additionally, it would be great if we could have a test around the `if` logic to validate the correct API version does get set, but it definitely seems non-trivial. Let me know what you think.

lawliet89:
Should I set the value to `Ignore` to connect inject to replicate the existing behaviour?

I'm not sure how we can test that if clause. I'll look into it.

ashwin-venkatesh:
> Should I set the value to Ignore for connect inject to replicate the existing behaviour?

I think setting the value to `Fail` so that our webhooks have identical behavior would be ideal in this case.

> I'm not sure how we can test that if clause. I'll look into it.

Yeah. I'm not entirely certain what a test case would look like either. Thanks so much for looking into it. Really appreciate it.

lawliet89:
I have set `failurePolicy: Fail`.

I am not able to test the `apiVersion` reliably.

helm template has the `--api-versions` flag but it seems to be an additive operation.

```
$ helm template -s templates/connect-inject-mutatingwebhook.yaml  --set 'connectInject.enabled=true' --api-versions "admissionregistration.k8s.io/v1beta1"  .
---
# Source: consul/templates/connect-inject-mutatingwebhook.yaml
# The MutatingWebhookConfiguration to enable the Connect injector.
apiVersion: admissionregistration.k8s.io/v1
kind: MutatingWebhookConfiguration
metadata:
  name: RELEASE-NAME-consul-connect-injector-cfg
  namespace: default
  labels:
    app: consul
    chart: consul-helm
    heritage: Helm
    release: RELEASE-NAME
webhooks:
  - name: RELEASE-NAME-consul-connect-injector.consul.hashicorp.com
    failurePolicy: Fail
    clientConfig:
      service:
        name: RELEASE-NAME-consul-connect-injector-svc
        namespace: default
        path: "/mutate"
      caBundle: ""
    rules:
      - operations: [ "CREATE" ]
        apiGroups: [""]
        apiVersions: ["v1"]
        resources: ["pods"]
```

ashwin-venkatesh:
@lawliet89 that's fair. I will test this PR against an older K8s version tomorrow as a quick sanity check, but outside of that, this looks great. Really appreciate the effort.

templates/connect-inject-mutatingwebhook.yaml
```diff
@@ -12,6 +16,7 @@ metadata:
     release: {{ .Release.Name }}
 webhooks:
   - name: {{ template "consul.fullname" . }}-connect-injector.consul.hashicorp.com
+    failurePolicy: Fail
```
ashwin-venkatesh, line 19: To be compliant with the `admissionregistration.k8s.io/v1` spec of webhooks, a few additional changes need to be made to the webhooks. We need to add the `sideEffects: None` to the webhook here along with `failurePolicy`.

Additionally, every webhook in this file and `templates/controller-mutatingwebhookconfiguration.yaml` need to add an `admissionReviewVersions` field to each of their webhooks as well. I was able to add those and get those webhooks to get submitted successfully. 
```
  admissionReviewVersions:
  - "v1beta1"
  - "v1"
```

ashwin-venkatesh (changes requested):
Hey @lawliet89 ,

I ran into a few problem during my testing while trying to deploy this against a cluster with Kubernetes v1.16.13 with regards to validations on the webhook configuration. Could you update the PR with the fixed required here.

I do have the fixed locally, and if you'd like, I could update the PR with them as well.

Let me know if you have any questions regarding my comments.

templates/connect-inject-mutatingwebhook.yaml
lawliet89: Sorry for missing these out, I have added them.

ashwin-venkatesh (approved):
Great work!! Thanks @lawliet89 for making the changes!!

PR #661, opened 2020-10-22
acceptance-tests: expect conn reset by peer errors when health checks are failing

ishustava:
The behavior currently is that on some cloud providers the error flips back and forth between "connection reset by peer" and "empty reply from server" when health checks are failing. On openshift, we never see `"empty reply from server"`  errors, and so the tests never pass.

* When a connection to the upstream is failing due to a failing health check,
  there will not be a healthy proxy host to connect to, and so we cannot assert
  that we will get "empty reply from server" error as we do in other tests.

* Bump GKE version to 1.16 since 1.15 is now deprecated.


kschoche (approved):
excellent work, thanks for fixing this issue! :)

lkysow (approved):

PR #662, opened 2020-10-23
Add age and labels to CRDs

thisisnotashwin:
Add the column `Age` to CRDs. Additionally add labels to the CRD to ensure they can be selected using the `app` like other consul components.


kschoche (approved):
🥇

lkysow (approved):

PR #663, opened 2020-10-25
WIP: Automatically generate gossip encryption key

### LINKED ISSUE #656: add support for auto-generating gossip encryption secret
Currently it is required to BYO the gossip encryption key, and this is documented in the values.yaml file: 
```
 # gossipEncryption configures which Kubernetes secret to retrieve Consul's
 # gossip encryption key from (see https://www.consul.io/docs/agent/options.html#_encrypt).
 # If secretName or secretKey are not set, gossip encryption will not be enabled.
 # The secret must be in the same namespace that Consul is installed into.
 #
 # The secret can be created by running:
 #    kubectl create secret generic consul-gossip-encryption-key \
 #      --from-literal=key=$(consul keygen).
 #
 ```
It would be helpful to have the option to have helm auto-generate the gossip encryption key and install the secret at install time.

### PR DESCRIPTION
**Pull request by nflaig**
Resolves #656 

Still work in progress to get some feedback. 

I tried to avoid breaking changes that's why I did not add a new value but in my opinion it would make sense to have default values for `.Values.global.gossipEncryption.secretName` and `.Values.global.gossipEncryption.secretKey` such as `gossip-encryption-key` and `key` and introduce a new value `.Values.global.gossipEncryption.enabled` to determine if it should be enabled or not but this would be a breaking change.

> values.yaml

```yaml
gossipEncryption:
    enabled: false
    secretName: "gossip-encryption-key"
    secretKey: "key"
```

Also one more thing to point out is that it does not use `consul keygen` but the key generated by the helm chart should be as good.

### Todos
- [ ] add tests


lkysow (commented):
Hi Nico, thanks for the PR! Our security team has reviewed the gossip key generation and determined that there is not enough entropy in `randAscii` since 1 bit of each ascii character is always the same. We've looked at other helm functions and haven't found one that satisfies the security constraints so I think we're going to need to do a Kubernetes job and use `consul keygen` unfortunately.

t-eckert:

PR #672, opened 2020-11-02
acceptance-tests: Add timestamps to logs

### LINKED ISSUE #625: Add timestamps to acceptance test logs
Often I find I need to cross-reference pod logs with the test logs. If the test logs had timestamps that would be a lot easier.

ishustava:
Fixes #625 

Changes proposed in this PR:

- Add timestamps to logs printed out by tests
- Bump terratest library to the latest version to include [this change](https://github.com/gruntwork-io/terratest/pull/674)
- Break up framework package into smaller packages
   - Move suite related code to the `suite` package
   - Move config-related code to the `config` package
   - Move test environment and test context related code to the `environment` package
   - Move consul cluster creation code to the `consul` package
   - Move flags to the `flags` package
   - Move the `helpers` package to `framework/helpers`
   - New logging code is all in the `logger` package

Notes to reviewers:
I thought this refactor made the tests read nicer and also organization a bit simper. I'm happy to split out the refactor from the logging change.


lkysow:
```
logger.go:35: 2020-11-02T19:46:09Z NAME: test-ltmdz1
```
I wonder if you can set `t.Helper()` or something so the logger doesn't include this stack level

ishustava:
> I wonder if you can set t.Helper() or something so the logger doesn't include this stack level

Yeah! good catch!

test/acceptance/framework/logger/logger.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+	"testing"
+	"time"
+
+	terratestTesting "github.com/gruntwork-io/terratest/modules/testing"
+)
+
+// TestLogger implements terratest's TestLogger interface
+// so that we can pass it to terratest objects to have consistent logging
+// across all tests.
```
ndhanushkodi, line 13: This is awesome!! I'm going to try to summarize my understanding of how to use this logger for future acceptance test code, and please do correct me if I'm wrong. In our own test helper methods, we should use `logger.Logf(...)`. In any terratest methods we leverage, we would pass it a logger `terratestLogger.New(logger.TestLogger{})` if that method in terratest already accepts a logger. Or if the method in terratest does not accept a logger, we'd reimplement it to take a logger, as described in the comment on line 12 in `test/acceptance/framework/k8s/kubectl.go`.

test/acceptance/framework/k8s/kubectl.go
```diff
@@ -1,10 +1,11 @@
...
 
 import (
 	"github.com/gruntwork-io/terratest/modules/k8s"
-	"github.com/gruntwork-io/terratest/modules/logger"
+	terratestLogger "github.com/gruntwork-io/terratest/modules/logger"
 	"github.com/gruntwork-io/terratest/modules/shell"
 	"github.com/gruntwork-io/terratest/modules/testing"
+	"github.com/hashicorp/consul-helm/test/acceptance/framework/logger"
 	"github.com/stretchr/testify/require"
 )
```
ndhanushkodi, line 11: This comment is regarding L12-15.

I know you made a contribution to terratest to pass in a custom logger for the tunnel struct, and I was wondering if we'd be able to leverage changes from their larger logging change https://github.com/gruntwork-io/terratest/pull/510 to not have to reimplement these methods anymore?

ndhanushkodi (approved):
This looks awesome! I made some comments attempting to understand how we should use the new logger package you wrote. If you think it's useful to document your answers to those comments/questions in some kind of code comment or in the commit message that could be helpful! But if it's just me misunderstanding some stuff don't worry about it :)

lkysow (approved):
Looks good barring Nitya's comments. We'll need a follow-up PR to swap the newer tests from `t.Log`.

test/acceptance/framework/consul/consul_cluster.go
```diff
@@ -76,6 +76,7 @@ func NewHelmCluster(
 		kubernetesClient:   ctx.KubernetesClient(t),
 		noCleanupOnFailure: cfg.NoCleanupOnFailure,
 		debugDirectory:     cfg.DebugDirectory,
+		logger:             terratestLogger.New(logger.TestLogger{}),
```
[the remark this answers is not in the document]
ishustava: There's no reason! good catch and thank you!

test/acceptance/framework/consul/consul_cluster.go
```diff
@@ -91,7 +92,7 @@ func (h *HelmCluster) Create(t *testing.T) {
 	// Fail if there are any existing installations of the Helm chart.
 	h.checkForPriorInstallations(t)
 
-	err := helm.InstallE(t, h.helmOptions, helmChartPath, h.releaseName)
+	err := helm.InstallE(t, h.helmOptions, config.HelmChartPath, h.releaseName)
 	require.NoError(t, err)
```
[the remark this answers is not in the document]
ishustava: 💯

test/acceptance/framework/k8s/kubectl.go
ishustava: Good question! The reason we're re-implementing some of them is so that we can set our own logger rather than use their [default one](https://github.com/gruntwork-io/terratest/blob/90e471fe491ff724bcb61c89e38bf6cde76e5eca/modules/logger/logger.go#L17-L19). These same methods in the terratest library would use the terratest logger (which is printing logs in verbose format regardless of whether you run test with or without `-v` flag). 

That PR you mentioned enabled us to pass our own logger, whereas previously, they would just use the terratest logger and not let you modify it. Prior to this PR, we were using `logger.TestingT` logger, but it doesn't have timestamps.

Hope this makes sense!

test/acceptance/framework/logger/logger.go
ishustava: Yup exactly!

ishustava:
I'll merge this one after the release on Thursday.

test/acceptance/framework/k8s/kubectl.go
ndhanushkodi: @ishustava Thanks, that makes sense! Would that mean in a future cleanup we could potentially remove our reimplementation and just pass in the new logger you wrote?

ishustava: I think we'd need to modify the implementation in the terratest library to be able to not re-implement these. The reason for that is that the place where you can modify the logger is when you create the Command object in the `RunKubectlAndGetOutputWithLoggerE` [here](https://github.com/hashicorp/consul-helm/blob/c4a699cb8080a574d5db07a69c96cff8a3e40989/test/acceptance/framework/k8s/kubectl.go#L41).

PR #673, opened 2020-11-04
pass health checks controller flag only when enabled

kschoche:
It is unnecessary to pass the health checks controller flag and this does cause some back compatibility issues if someone were to grab the latest consul-helm repo/release and use it against an old consul-k8s release. So make this optional and only pass when health checks are enabled.

### DIFF
--- a/templates/connect-inject-deployment.yaml
+++ b/templates/connect-inject-deployment.yaml
@@ -85,8 +85,10 @@ spec:
                 -envoy-image="{{ .Values.global.imageEnvoy }}" \
                 -consul-k8s-image="{{ default .Values.global.imageK8S .Values.connectInject.image }}" \
                 -listen=:8080 \
+                {{- if .Values.connectInject.healthChecks.enabled }}
                 -enable-health-checks-controller={{ .Values.connectInject.healthChecks.enabled }} \
                 -health-checks-reconcile-period={{ .Values.connectInject.healthChecks.reconcilePeriod }} \
+                {{- end }}
                 {{- if .Values.connectInject.overrideAuthMethodName }}
                 -acl-auth-method="{{ .Values.connectInject.overrideAuthMethodName }}" \
                 {{- else if .Values.global.acls.manageSystemACLs }}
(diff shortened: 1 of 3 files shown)


test/acceptance/tests/connect/health_checks_test.go
```diff
@@ -40,6 +40,7 @@ func TestHealthChecks(t *testing.T) {
 			cfg := suite.Config()
 
 			helmValues := map[string]string{
+				"global.imageK8S":                    "kschoche/consul-k8s-dev",
```
kschoche, line 43:
```suggestion
```

templates/connect-inject-deployment.yaml
```diff
@@ -85,8 +85,10 @@ spec:
                 -envoy-image="{{ .Values.global.imageEnvoy }}" \
                 -consul-k8s-image="{{ default .Values.global.imageK8S .Values.connectInject.image }}" \
                 -listen=:8080 \
+                {{- if .Values.connectInject.healthChecks.enabled }}
                 -enable-health-checks-controller={{ .Values.connectInject.healthChecks.enabled }} \
```
lkysow, line 89:
```suggestion
                -enable-health-checks-controller=true \
```

This would match 
```
                {{- if .Values.connectInject.centralConfig.enabled }}
                -enable-central-config=true \
                {{- end }}
```

and

```
                {{- if .Values.global.enableConsulNamespaces }}
                -enable-namespaces=true \
```

lkysow (approved):

templates/connect-inject-deployment.yaml
kschoche: I have no idea how that got in there.. thx for the catch!

test/unit/connect-inject-deployment.bats
```diff
@@ -116,15 +119,16 @@ load _helpers
       yq '.spec.template.spec.containers[0].command' | tee /dev/stderr)
 
   local actual=$(echo "$cmd" |
-    yq 'any(contains("-enable-health-checks-controller=false"))' | tee /dev/stderr)
-  [ "${actual}" = "true" ]
+    yq 'any(contains("-enable-health-checks-controller"))' | tee /dev/stderr)
+  [ "${actual}" = "false" ]
 }
 
 @test "connectInject/Deployment: health checks reconcile period set by default" {
```
ishustava, line 126:
```suggestion
@test "connectInject/Deployment: health checks reconcile period set by default when health checks are enabled" {
```

ishustava (approved):
🎉

PR #674, opened 2020-11-06
Set check_update_interval for healthchecks

lkysow:
The check_update_interval setting controls how often Consul clients
update Consul servers when the output of their checks have changed but
the status hasn't. This
defaults to 5m because often users would create checks that would run
every minute or so and whose output
contained a timestamp that would change on every iteration of the check.
This would cause a lot of writes to the Consul servers even though
the status of the check didn't change.

In our use-case, we only update the check when the pod's readiness
status changes, so we won't be spamming updates. We run into an issue
though because we always set the check to failing when the pod is first
created and then, because of an issue with Consul where you can't set
the check output on first creation, we issue an update to the check with
the same status (critical) but with output "Pod is pending". This output
will take 5 minutes to show up in the UI without this setting. That
might be okay if the pod immediately becomes healthy but if it doesn't
then it's likely going to be confusing for users why the check is
failing but it doesn't have any output.

Additionally, if the pod's readiness probes fail immediately, then we'll
update the output to be something like "Readiness probes failing" but
again, the UI won't be updated because the status is still critical and
so again the UI would show the check failing with no output.

(also rename the test file to match the file it's testing as per our convention).

### DIFF
--- a/templates/client-config-configmap.yaml
+++ b/templates/client-config-configmap.yaml
@@ -20,4 +20,13 @@ data:
       "enable_central_service_config": true
     }
   {{- end }}
+
+  {{- if (and .Values.connectInject.enabled .Values.connectInject.healthChecks.enabled) }}
+  {{/* We set check_update_interval to 0s so that check output is immediately viewable
+       in the UI. */}}
+  config.json: |-
+    {
+      "check_update_interval": "0s"
+    }
+  {{- end }}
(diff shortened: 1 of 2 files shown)


kschoche (approved):
🧨

ndhanushkodi (approved):
Looks great 🎉

PR #680, opened 2020-11-10
update changelog for health checks

kschoche:
Updating the changelog for changes related to health checks, also did some minor formatting updates to a previous changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,19 +1,25 @@
 ## Unreleased
 
 FEATURES:
-* Support Kubernetes health probe synchronization with Consul for connect injected pods via `connectInject.healthChecks` [[GH-651](https://github.com/hashicorp/consul-helm/pull/651)].
+* Kubernetes health check synchronization with Consul for connect injected pods via `connectInject.healthChecks` [[GH-651](https://github.com/hashicorp/consul-helm/pull/651)].
   The default behavior for this feature is `enabled: true`.
   See [https://www.consul.io/docs/k8s/health](https://www.consul.io/docs/k8s/health) for more information.
-  **It is recommended to enable TLS with this setting enabled.**
+  In order to enable this feature for existing installations it is required to restart all connect injected deployments so that they are re-injected.
+  Until this is done health checks for these deployments will not be synced to Consul.
+
+  **It is recommended to enable TLS with this setting enabled because it requires making calls to Consul clients across the cluster.
+    Without TLS enabled, these calls could leak ACL tokens should the cluster network become compromised.**
+
+BREAKING CHANGES:
+* This helm release only supports consul-k8s versions 0.20+
+* With the addition of the connect-inject health checks controller any connect services which have failing Kubernetes readiness
+  probes will no longer be routable through connect until their Kubernetes health probes are passing.
+  Previously, if any connect services were failing their Kubernetes readiness checks they were still routable through connect.
+  Users should verify that their connect services are passing Kubernetes readiness probes prior to using health checks synchronization.
 
 BUG FIXES:
-* Support new versions of timeout in ent-license job [[GH-654](https://github.com/hashicorp/consul-helm/pull/654)].
-
-DEPRECATIONS:
-* `create-inject-token` in the server-acl-init job has been undeprecated.
-  `create-inject-auth-method` has been deprecated and replaced by `create-inject-token`.
-  `create-inject-namespace-token` in the server-acl-init job has been deprecated. [[GH-665](https://github.com/hashicorp/consul-helm/pull/665)].
-  See: [[GH-368](https://github.com/hashicorp/consul-k8s/pull/368)] for additional documentation regarding these changes.
+* Fix issue where Consul enterprise license job would fail for Consul versions >= 1.8.1" [[GH-654](https://github.com/hashicorp/consul-helm/pull/654)].
+  [[#647](https://github.com/hashicorp/consul-helm/issues/647)]
 
 IMPROVEMENTS:
 * Connect: support passing extra arguments to the injected envoy sidecar. [[GH-675](https://github.com/hashicorp/consul-helm/pull/675)]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,32 @@
 ## Unreleased
 
+FEATURES:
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,16 +1,32 @@
...
+* Support new versions of timeout in ent-license job [[GH-654](https://github.com/hashicorp/consul-helm/pull/654)].
+
+DEPRECATIONS:
+* `create-inject-token` in the server-acl-init job has been undeprecated.
+  `create-inject-auth-method` has been deprecated and replaced by `create-inject-token`.
+  `create-inject-namespace-token` in the server-acl-init job has been deprecated. [[GH-665](https://github.com/hashicorp/consul-helm/pull/665)].
+  See: [[GH-368](https://github.com/hashicorp/consul-k8s/pull/368)] for additional documentation regarding these changes.
+
 IMPROVEMENTS:
-  * Connect: support passing extra arguments to the injected envoy sidecar. [[GH-675](https://github.com/hashicorp/consul-helm/pull/675)]
```
kschoche, line 4: This whole section had 2 extra spaces so I've removed them.

CHANGELOG.md
```diff
@@ -1,16 +1,32 @@
 ## Unreleased
 
+FEATURES:
+* Support Kubernetes health probe synchronization with Consul for connect injected pods via `connectInject.healthChecks` [[GH-651](https://github.com/hashicorp/consul-helm/pull/651)].
+  The default behavior for this feature is `enabled: true`.
+  See [https://www.consul.io/docs/k8s/health](https://www.consul.io/docs/k8s/health) for more information.
+  **It is recommended to enable TLS with this setting enabled.**
```
ndhanushkodi, line 7: Maybe a comment on why it's recommended to enable TLS would be helpful, or a link to where in the doc it would describe that. So that I could know which network connections are insecure if I don't turn on TLS.

CHANGELOG.md
```diff
@@ -1,16 +1,32 @@
 ## Unreleased
 
+FEATURES:
+* Support Kubernetes health probe synchronization with Consul for connect injected pods via `connectInject.healthChecks` [[GH-651](https://github.com/hashicorp/consul-helm/pull/651)].
+  The default behavior for this feature is `enabled: true`.
+  See [https://www.consul.io/docs/k8s/health](https://www.consul.io/docs/k8s/health) for more information.
```
ndhanushkodi, line 6: This link doesn't resolve yet, but I'm assuming that's just coming in the future?

CHANGELOG.md
```diff
@@ -1,16 +1,32 @@
...
+  **It is recommended to enable TLS with this setting enabled.**
+
+BUG FIXES:
+* Support new versions of timeout in ent-license job [[GH-654](https://github.com/hashicorp/consul-helm/pull/654)].
+
+DEPRECATIONS:
+* `create-inject-token` in the server-acl-init job has been undeprecated.
+  `create-inject-auth-method` has been deprecated and replaced by `create-inject-token`.
+  `create-inject-namespace-token` in the server-acl-init job has been deprecated. [[GH-665](https://github.com/hashicorp/consul-helm/pull/665)].
+  See: [[GH-368](https://github.com/hashicorp/consul-k8s/pull/368)] for additional documentation regarding these changes.
```
ndhanushkodi, line 16: The table in PR 368 was so helpful! The descriptions of the flags are clear and understandable. Great idea to link it here, hopefully something like that will make it into https://www.consul.io/docs/k8s/health.

ndhanushkodi (approved):
Awesome work!! I only had a few non-blocking comments.

CHANGELOG.md
kschoche: Good catch, there is an open PR here for it : https://github.com/hashicorp/consul/pull/8819 
I can hold off on merging this until it gets merged.

CHANGELOG.md
kschoche: doc in link ^ describes it

CHANGELOG.md
```diff
@@ -1,16 +1,32 @@
 ## Unreleased
 
+FEATURES:
+* Support Kubernetes health probe synchronization with Consul for connect injected pods via `connectInject.healthChecks` [[GH-651](https://github.com/hashicorp/consul-helm/pull/651)].
```
lkysow, line 4: 1. This needs a breaking change section as well. If any connect services were failing their kubernetes readiness checks previously then they were still routable through connect, now they won't be. Users should verify that their connect services are passing Kubernetes readiness probes prior to enabling health checks.
1. Can we add that in order to enable this feature for existing users they need to restart all the connect deployments so they are re-injected with a new label. Until they restart those deployments, health checks won't be synced.
1. I think we can be a bit more explicit here. Can we add something like "Now Kubernetes pods that fail their readiness checks will no longer receive Connect traffic until the readiness check passes"

CHANGELOG.md
lkysow: I think we should still document it here since it only requires a few more words ("because the health checks controller makes calls to Consul agents across all nodes so ACL tokens could be stolen if an attacker was watching the network")

CHANGELOG.md
```diff
@@ -1,16 +1,32 @@
...
+  The default behavior for this feature is `enabled: true`.
+  See [https://www.consul.io/docs/k8s/health](https://www.consul.io/docs/k8s/health) for more information.
+  **It is recommended to enable TLS with this setting enabled.**
+
+BUG FIXES:
+* Support new versions of timeout in ent-license job [[GH-654](https://github.com/hashicorp/consul-helm/pull/654)].
```
lkysow, line 10: Let's be explicit about the bug that was fixed. Something like "Fix issue where Consul enterprise license job would fail for Consul versions >= 1.8.1".

Also we should link to the issue fixed for bug fixes if there is one (https://github.com/hashicorp/consul-helm/issues/647)

CHANGELOG.md
lkysow, line 16: These are consul-k8s deprecations. I don't think they should be listed for the consul-helm changelog since there's nothing for users to do.

What we do need to list is a breaking change that only consul-k8s v0.20.0 is supported with this helm chart version.

CHANGELOG.md
lkysow: Or your excellent description in values.yaml
```
  # NOTE: It is highly recommended to enable TLS with this feature because it requires
  # making calls to Consul clients across the cluster. Without TLS enabled, these calls
  # could leak ACL tokens.
```

thisisnotashwin (approved):
looks good!

PR #685, opened 2020-11-12
Add script to aid crd migration

lkysow:
This script will add a metadata key to all service-defaults and
proxy-defaults config entries.

Custom resources can then be created for these CRDs. Without the change
to the metadata, the CRD controller will think these resources are
managed by another datacenter and it won't sync any changes for the
custom resources.


lkysow:
Will be closed by https://github.com/hashicorp/consul-k8s/pull/419

PR #691, opened 2020-11-16
Enable envoy sidecar debug logs in acceptance tests

ndhanushkodi:
* Envoy sidecar debug logs are enabled by default anytime you run acceptance tests

### DIFF
--- a/test/acceptance/framework/consul/consul_cluster_test.go
+++ b/test/acceptance/framework/consul/consul_cluster_test.go
@@ -14,15 +14,45 @@ import (
 // to a helm install, it will respect the helmValues over
 // the values from TestConfig.
 func TestNewHelmCluster(t *testing.T) {
-	helmValues := map[string]string{
-		"global.image":           "test-image",
-		"feature.enabled":        "true",
-		"server.bootstrapExpect": "3",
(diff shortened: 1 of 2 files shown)


test/acceptance/framework/consul/consul_cluster_test.go
```diff
@@ -14,15 +14,45 @@ import (
...
-		"server.replicas":        "3",
+	tests := []struct {
+		name       string
+		helmValues map[string]string
+		want       map[string]string
+	}{
+		{
+			name:       "defaults are used when no helmValues are set",
+			helmValues: map[string]string{},
+			want: map[string]string{
```
kschoche, line 25: Looks great!
minor nit which isn't blocking, could you reorder the helm values so between the 3 maps the entries are in a similar order, it makes it a little bit difficult to read!

kschoche (approved):

ishustava (approved):
🎉

test/acceptance/framework/consul/consul_cluster_test.go
ndhanushkodi: yup totally, good catch!

PR #695, opened 2020-11-18
Error mirroringK8S=true but Consul NS's disabled

lkysow:
Based on an issue someone had reported. They had set
```yaml
connectInject:
  consulNamespaces:
    mirroringK8S: true
```

But didn't have

```yaml
global:
  enableConsulNamespaces: true
```

This caused us to ignore that configuration.

NOTE: We can't error on `connectInject.consulNamespaces.destinationNS` because that has a default value so it's always set.

### DIFF
--- a/test/unit/connect-inject-deployment.bats
+++ b/test/unit/connect-inject-deployment.bats
@@ -683,6 +683,17 @@ load _helpers
 #--------------------------------------------------------------------
 # namespaces
 
+@test "connectInject/Deployment: fails if namespaces are disabled and mirroringK8S is true" {
+  cd `chart_dir`
+  run helm template \
+      -s templates/connect-inject-deployment.yaml  \
+      --set 'global.enableConsulNamespaces=false' \
+      --set 'connectInject.consulNamespaces.mirroringK8S=true' \
+      --set 'connectInject.enabled=true' .
+  [ "$status" -eq 1 ]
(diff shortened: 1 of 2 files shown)


test/unit/connect-inject-deployment.bats
```diff
@@ -683,6 +683,16 @@ load _helpers
 #--------------------------------------------------------------------
 # namespaces
 
+@test "connectInject/Deployment: fails if namespaces are disabled and mirroringK8S is true" {
+  cd `chart_dir`
+  run helm template \
+      -s templates/connect-inject-deployment.yaml  \
```
ashwin-venkatesh, line 689: I know namespaces defaults to `false`, but can we set it to false explicitly in this test for clarity? Or rename the testcase to explicitly state namespace aren't enabled?

thisisnotashwin (commented):
Thanks for making this change!! Just have a suggestion with regards to the test but it looks great otherwise.

kschoche (approved):

test/unit/connect-inject-deployment.bats
lkysow: done.

thisisnotashwin (approved):
🦊

test/unit/connect-inject-deployment.bats
ashwin-venkatesh: Thanks!!

PR #696, opened 2020-11-19
acceptance-tests: Fold health checks tests into connect

ishustava:
These tests are quite similar except for the additional assertion in the health checks tests that asserts
that traffic is not routed when the static-server probe fails.

This will save us quite a bit on run-time for both regular and *namespace tests (e.g. on openshift the connect package now takes longer than 40min to run)  because the bulk of the test time is taken up by the installs. This change essentially re-uses existing installs that we already have to do to test connect.

Note: since health checks are enabled by default I did not explicitly set it to true in the helm configs in the tests.


test/acceptance/framework/consul/consul_cluster.go
```diff
@@ -149,7 +149,7 @@ func (h *HelmCluster) Destroy(t *testing.T) {
 	}
 
 	// Delete any secrets that have h.releaseName in their name.
-	secrets, err := h.kubernetesClient.CoreV1().Secrets(h.helmOptions.KubectlOptions.Namespace).List(context.Background(), metav1.ListOptions{LabelSelector: "release=" + h.releaseName})
+	secrets, err := h.kubernetesClient.CoreV1().Secrets(h.helmOptions.KubectlOptions.Namespace).List(context.Background(), metav1.ListOptions{})
```
ishustava, line 152: This is a bug that leaves secrets around after test runs. We don't add release labels to the secrets we create in the helm chart jobs.

kschoche (approved):
This is awesome work and a great cleanup!

ndhanushkodi (approved):
Looks great 💯 🎉

PR #700, opened 2020-11-20
Run Acceptance tests against EKS 1.17

ishustava:
* Add terraform to create EKS clusters
* Update test docker image to use terraform 1.13
  so that we can use "count" with terraform modules.
* Add AWS CLI to the terraform docker image
* Add a circleci job to run acceptance tests against EKS.

Note: Waiting for EKS creds for circleci to be created so that we can run these in CI. I've ran both TF and a subset of tests locally, and it all works.

TODO:
- [] Run tests in CI
- [] Once they pass, move the job to the nightly pipeline

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ jobs:
           command: bats ./test/unit
   unit-helm3:
     docker:
-      - image: docker.mirror.hashicorp.services/hashicorpdev/consul-helm-test:0.6.0
+      - image: docker.mirror.hashicorp.services/hashicorpdev/consul-helm-test:0.7.0
 
     steps:
       - checkout
@@ -96,7 +96,7 @@ jobs:
       - TEST_RESULTS: /tmp/test-results
     docker:
       # This image is build from test/docker/Test.dockerfile
-      - image: docker.mirror.hashicorp.services/hashicorpdev/consul-helm-test:0.5.0
+      - image: docker.mirror.hashicorp.services/hashicorpdev/consul-helm-test:0.7.0
 
     steps:
       - run:
@@ -172,7 +172,7 @@ jobs:
       - TEST_RESULTS: /tmp/test-results
     docker:
       # This image is build from test/docker/Test.dockerfile
-      - image: docker.mirror.hashicorp.services/hashicorpdev/consul-helm-test:0.5.0
+      - image: docker.mirror.hashicorp.services/hashicorpdev/consul-helm-test:0.7.0
 
     steps:
       - checkout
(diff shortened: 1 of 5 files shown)


lkysow (approved):

test/acceptance/framework/k8s/deploy.go
```diff
@@ -83,21 +83,7 @@ func CheckStaticServerConnection(
...
-		output, err := RunKubectlAndGetOutputE(t, options, args...)
-		if expectSuccess {
-			require.NoError(r, err)
-			require.Contains(r, output, "hello world")
-		} else {
-			require.Error(r, err)
-			require.Contains(r, output, failureMessage)
-		}
-	})
+	CheckStaticServerConnectionMultipleFailureMessages(t, options, expectSuccess, deploymentName, []string{failureMessage}, curlArgs...)
```
ndhanushkodi, line 86: nice!

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
```diff
@@ -225,23 +225,17 @@ func TestMeshGatewaySecure(t *testing.T) {
...
-func verifyFederation(t *testing.T, primaryClient, secondaryClient *api.Client, secure bool) {
-	const consulMemberStatusAlive = 1
-
-	retrier := &retry.Timer{Timeout: 1 * time.Minute, Wait: 1 * time.Second}
+func verifyFederation(t *testing.T, primaryClient, secondaryClient *api.Client, releaseName string, secure bool) {
+	retrier := &retry.Timer{Timeout: 2 * time.Minute, Wait: 1 * time.Second}
 
 	retry.RunWith(retrier, t, func(r *retry.R) {
-		members, err := primaryClient.Agent().Members(true)
+		secondaryServerHealth, _, err := primaryClient.Health().Node(fmt.Sprintf("%s-consul-server-0", releaseName), &api.QueryOptions{Datacenter: "dc2"})
```
ndhanushkodi, line 232: Was just curious what this line does, is it the consul client on the primary datacenter is checking the health endpoint of the server in secondary datacenter? Why do we need to make this request from opposite datacenters?

test/terraform/gke/main.tf
```diff
@@ -43,6 +43,6 @@ resource "null_resource" "kubectl" {
   provisioner "local-exec" {
     when       = destroy
     on_failure = continue
-    command    = "rm $HOME/.kube/${google_container_cluster.cluster[count.index].name}"
+    command    = "rm $HOME/.kube/consul-k8s*"
```
ndhanushkodi, line 46: has the location of the kubectl credentials moved? or is this a simplification so we don't have to use the variable?

ndhanushkodi (approved):
this is great!! just had a few clarifying questions.

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
ishustava: Sorry! I will add a comment.

Yup exactly. We're trying to see if one datacenter can "see" the other, and vice versa. For that, we calling the health API to check the health of the server in dc1 from the perspective of dc2 and vice versa. This is how we now the federation is "successful" because that means one datacenter can query and forward requests to the other.

test/terraform/gke/main.tf
ishustava: good question! This is because, in terraform 1.13+, they've removed an ability to reference resources from destroy-time provisioners. This means that you can't reference other resources (e.g. `google_container_cluster`) in these provisioners because terraform may not know what that reference resolves to if that resource has already been destroyed.

This change allows us to work around this new restriction. Since our GKE clusters and their respective kubeconfig files have the `consul-k8s` prefix, we can just delete those files based on that instead of referencing the name of the terraform resource.

lkysow (approved):
⚡

test/terraform/gke/main.tf
ndhanushkodi: Ah thank you for the explanation!

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
ndhanushkodi: Ahh, got it. Thank you!

ishustava: Added a comment! Thanks for the question 🙏

PR #701, opened 2020-11-21
acceptance-tests: use new logger with timestamps everywhere

ishustava:
This is a follow up to #671: change remaining tests and framework functions to use the new logger with timestaamps.


lkysow (approved):

kschoche (approved):
Looks great!

PR #708, opened 2020-11-25
Run acceptance tests against a cluster with Pod security policies

ishustava:
* Enable pod security policies on a GKE cluster. 
* Add `enable-pod-security-policies` flag to acceptance tests framework. When this flag is set, the tests will:
   * set `global.enablePodSecurityPolicies` helm value for all helm installs
   * Create a simple pod security policy, a cluster role that allows access to it, and a role binding, linking psp cluster role to the  default service account in the helm installation namespace. We need these so that test resources we create, like static-client, static-server, and any other test pods we spin up, won't be rejected by the pod security policy controller.
  * Add a role binding to the static-server and static-client base fixtures to bind their service accounts to the psp cluster role.

### DIFF
--- a/test/acceptance/framework/config/config_test.go
+++ b/test/acceptance/framework/config/config_test.go
@@ -76,6 +76,15 @@ func TestConfig_HelmValuesFromConfig(t *testing.T) {
 				"global.openshift.enabled": "true",
 			},
 		},
+		{
+			"sets enablePodSecurity policies when -enable-pod-security-policies is set",
+			TestConfig{
+				EnablePodSecurityPolicies: true,
+			},
+			map[string]string{
+				"global.enablePodSecurityPolicies": "true",
+			},
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
--- a/test/acceptance/framework/config/config_test.go
+++ b/test/acceptance/framework/config/config_test.go
@@ -77,7 +77,7 @@ func TestConfig_HelmValuesFromConfig(t *testing.T) {
 			},
 		},
 		{
-			"sets enablePodSecurity policies when -enable-pod-security-policies is set",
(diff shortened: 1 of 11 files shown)


test/terraform/gke/main.tf
```diff
@@ -1,4 +1,4 @@
-provider "google" {
+provider "google-beta" {
   project = var.project
```
lkysow, line 2: Should we set the specific provider version?

ishustava: Hmm I don't know if we need to, so I'd lean towards keeping it as is until we see problems.

lkysow: If this is a beta though then it might often have breaking changes.

lkysow (approved):

test/acceptance/framework/config/config_test.go
```diff
@@ -76,6 +76,15 @@ func TestConfig_HelmValuesFromConfig(t *testing.T) {
 				"global.openshift.enabled": "true",
 			},
 		},
+		{
+			"sets enablePodSecurity policies when -enable-pod-security-policies is set",
```
ashwin-venkatesh, line 80: not a real comment but it was funny to read `sets enable pod security policies when enable pod security policies is set`. kinda reads like a tongue twister

test/terraform/gke/main.tf
ashwin-venkatesh: Do we have resolution on this?

thisisnotashwin (approved):
Looks great. Once we have clarification on using `google-beta`, feel free to merge this 🥳

test/acceptance/framework/config/config_test.go
ishustava: yeah, it does sound weird! rephrased!

test/terraform/gke/main.tf
ishustava: Yep, added the version to be the latest.

PR #712, opened 2020-11-26
Run acceptance tests against kind

ishustava:
Changes proposed in this PR:
- Add a new flag to the framework, `-use-kind`. The only thing it affects currently are the mesh gateway tests: it uses NodePort service instead of the LoadBalancer service to expose the mesh gateways. This is because load balancer services are not available on kind. For the NodePort services, we don't need to do anything to allow cross-cluster communication because `kind` uses the docker bridge network, and so the nodes are reachable by default.
- Change the main pipeline to use kind instead of gke, and move gke tests to the nightly pipeline.
- Use circleci's parallelism feature to run tests in parallel. Currently, tests are split by package, which is not the best efficient way to parallelize them, but we can address that at a later time. Currently, using 6 parallel nodes reduces the test runtime to about 30min.

Note: I'm not using the mirror for our dev consul-k8s image because for some reason it results in pull image errors on kind due to a `401 (Unauthorized)` error from the docker registry. I have not seen rate-limiting errors from docker though, so I think it's ok.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,12 +91,12 @@ jobs:
       - store_artifacts:
           path: /tmp/test-results
 
-  acceptance-kind:
+  acceptance:
     environment:
       - TEST_RESULTS: /tmp/test-results
     machine:
       image: ubuntu-2004:202010-01
-    parallelism: 4
+    parallelism: 6
     steps:
       - checkout
       - run:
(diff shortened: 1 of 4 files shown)


lkysow (approved):

ndhanushkodi (approved):
this is awesome 🚤

Issue #512, opened 2020-06-23
Print tls-init errors


lkysow:
Currently if any `curl` command fails during `tls-init`, the job will exit without logging anything. Instead we'd like to log the curl error. We need to be careful we're not logging any secrets.

PR #719, opened 2020-12-01
Log tls-init errors

lkysow:
Previously any errors in TLS init due to a curl call failing would not
be logged because we ran curl with the -s flag. Now we run with -S as
well to ensure that those errors are printed.

Also adds some logging in-between steps so we'll know which curl call
failed.

Closes https://github.com/hashicorp/consul-helm/issues/512

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+BUG FIXES:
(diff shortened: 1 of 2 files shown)


ndhanushkodi (approved):
🎉 yay for logging

ishustava:
Going to wait for the outcome of hashicorp/consul-k8s#410 before reviewing this.

PR #721, opened 2020-12-01
Set default for bootstrapExpect to replicas

### LINKED ISSUE #573: Remove bootstrap_expect from values.yml
The [bootstrap_expect](https://www.consul.io/docs/agent/options.html#_bootstrap_expect) configuration of Consul seems to be a moot point in a kubernetes environment. Can we remove it from [values.yml ](https://www.consul.io/docs/k8s/helm.html#v-server-bootstrapexpect)and handle bootstrapping as part of the helm logic?

Reasons why:
- Setting `bootstrap_expect` to a number different than replicas creates log spam about more servers running in "bootstrap" mode than expected.
- Need to confirm - if a large enough replica count is requested, could we potentially split brain?
- Assuming all pods come up in a reasonable time, there's no speed gained before serving traffic by electing a leader before quorum is achieved.

Question
- When new helm deployments are rolled out, how is this "bootstrap" handled under the hood?

lkysow:
In almost all cases, users want to set bootstrapExpect to the number of
server replicas. This change defaults it to null in values.yaml and then
in the template if it's left as null, then we set the -bootstrap-expect
flag to the number of server replicas.

This is backwards compatible, if users have been setting this it will
continue to be set.

Also error out if bootstrapExpect is less than the number of replicas
because this is definitely a misconfiguration as the servers won't wait
until the proper number have started before electing a leader.

Before:
```yaml
server:
  replicas: 1
  bootstrapExpect: 1
```

After:
```yaml
server:
  replicas: 1
```

Closes https://github.com/hashicorp/consul-helm/issues/573

### DIFF
--- a/test/unit/server-statefulset.bats
+++ b/test/unit/server-statefulset.bats
@@ -916,3 +916,34 @@ load _helpers
       yq '.spec.template.spec.containers[0].command | join(" ") | contains("auto_encrypt = {allow_tls = true}")' | tee /dev/stderr)
   [ "${actual}" = "true" ]
 }
+
+#--------------------------------------------------------------------
+# -bootstrap-expect
+
+@test "server/StatefulSet: -bootstrap-expect defaults to replicas" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/server-statefulset.yaml  \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command | join(" ") | contains("-bootstrap-expect=3")' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
+@test "server/StatefulSet: -bootstrap-expect can be set by server.bootstrapExpect" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/server-statefulset.yaml  \
+      --set 'server.bootstrapExpect=5' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command | join(" ") | contains("-bootstrap-expect=5")' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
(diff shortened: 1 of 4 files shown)


test/unit/server-statefulset.bats
```diff
@@ -894,3 +894,34 @@ load _helpers
...
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/server-statefulset.yaml  \
+      --set 'server.bootstrapExpect=5' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[0].command | join(" ") | contains("-bootstrap-expect=5")' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
+@test "server/StatefulSet: errors if bootstrapExpect < replicas" {
```
ashwin-venkatesh, line 920: Is this change backwards compatible? Previously, we would allow `bootstrap expect` to be <= `replicas` but this change would cause such a setting to change.

lkysow: Ooh, you're right! Will add to changelog. I could also remove this but I think this is an invalid configuration.

ashwin-venkatesh: Interesting. I mean, your comment about bootstrapExpect made me realize that my understanding of it was incorrect. I thought it was the min number of servers that needed to be up and running for consensus, which seemed odd but 🤷 . This makes more sense.

lkysow: > To prevent inconsistencies and split-brain (clusters where multiple servers consider themselves leader) situations, you should either specify the same value for -bootstrap-expect or specify no value at all on all the servers. 

From https://www.consul.io/docs/install/bootstrapping

thisisnotashwin (approved):

kschoche (approved):
Excellent change.

PR #723, opened 2020-12-01
Allow configuring of terminationGracePeriodSeconds

lawliet89:
In an autoscaling cluster that scales up and down somewhat rapidly, I notice that when scaling down, there are times when a node has been deleted (and the pod too), but the node remains in the catalog in a critical state.

This PR allows increasing the time given for the Consul pods to terminate gracefully before they are sent SIGKILL.

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -45,7 +45,7 @@ spec:
       tolerations:
         {{ tpl .Values.client.tolerations . | nindent 8 | trim }}
     {{- end }}
-      terminationGracePeriodSeconds: 10
+      terminationGracePeriodSeconds: {{ .Values.client.terminationGracePeriodSeconds }}
       serviceAccountName: {{ template "consul.fullname" . }}-client
 
       {{- if .Values.client.priorityClassName }}
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -57,7 +57,7 @@ spec:
       tolerations:
         {{ tpl .Values.server.tolerations . | nindent 8 | trim }}
     {{- end }}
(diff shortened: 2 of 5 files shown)


lawliet89:
Hey @lkysow @ashwin-venkatesh , possible to have this be looked at?

lkysow:
@lawliet89 so when you're autoscaling and a node is shut down, kube doesn't actually send a SIGTERM to the daemonset pods (see https://github.com/kubernetes/kubernetes/issues/75482). So I'm not sure this would help.

With the release of Consul 1.9.0 we allow the configuration of per-agent reconnect timeouts (the time before Consul considers that node dead):

```yaml
client:
  extraConfig: |
    {"advertise_reconnect_timeout": "15m"}
```

I think that might better solve this.

For a full solution, kube needs to actually send a SIGTERM or we need to implement a controller that watches for nodes stopping and manually deregisters that consul client.

lawliet89:
Thanks @lkysow! I already had `advertise_reconnect_timeout` set to `8h` which I thought was the minimum according to the docs of `reconnect_timeout`. I'll try lowering it and see if it helps.

lawliet89:
Thanks @lkysow! I already have `advertise_reconnect_timeout` set to `8h` which I thought was the minimum according to the docs of `reconnect_timeout`. I'll try lowering it and see if it helps.

> For a full solution, kube needs to actually send a SIGTERM or we need to implement a controller that watches for nodes stopping and manually deregisters that consul client.

I believe Kube does send `SIGTERM` when shutting down. The `terminationGracePeriodSeconds` controls the amount of time it waits before it sends `SIGKILL`. Does Consul not receive `SIGTERM` at all?

Issue #657, opened 2020-10-20
Add support for environment variable HOST_IP


bsc8180:
We want to send dogstatsd metrics to a Datadog agent on the same node as our Consul pod using the telemetry stanza as follows

(configmap extra-from-values.json)

`{
  "telemetry": [
    {
      "disable_hostname": true,
      "dogstatsd_addr": "HOST_IP:8125",
      "dogstatsd_tags": ["env:weweq","service:consul","version:1.8.0"]
    }
  ]
}`

We currently do this for Vault, courtesy of the HOST_IP environment variable set by default in the [helm template](https://github.com/hashicorp/vault-helm/blob/ee4e532159b023dece6d69ec58666cd50d85a47b/templates/server-statefulset.yaml#L72).

We would like to to see the same functionality in the Consul helm chart, or clearer documentation on how to add complex environment variables to extraEnvironmentVars (the documentation shows only simple k/v pairs).

Should just be a matter of adding 

`            - name: HOST_IP
              valueFrom:
                fieldRef:
                  fieldPath: status.hostIP`

to [server-statefulset.yaml](https://github.com/hashicorp/consul-helm/blob/203d9073da4ce5c1067fba03d8ce0abcc0f4e20d/templates/server-statefulset.yaml#L107)

watsonian:
Prior art from the vault helm chart:

https://github.com/hashicorp/vault-helm/blob/fc8ebfdd4e596bf9b22f814265ddf57b4a454d69/templates/server-statefulset.yaml#L72-L75

watsonian:
Also worth pointing out this bit from the vault helm chart:

https://github.com/hashicorp/vault-helm/blob/994797cff4dc12887acc9fc7fccb70d34f3254b9/templates/_helpers.tpl#L119-L135

yannlambret:
Would it be possible to simply replace the current implementation of the partial template `consul.extraEnvironmentVars`, so that we can declare environment variables in the native specification format? I don't really see the added value of the map interface here.

PR #734, opened 2020-12-08
Add support to use HOST_IP as a variable in config file

### LINKED ISSUE #339: Support for Telemetry/DogStatsD
Hi,

Similar to https://github.com/hashicorp/consul-helm/issues/262 it would be great if the chart could support configuring telemetry options. We're using Datadog for monitoring/alerting and it seems that only a subset of metrics are ingested by the datadog-agent via Consul's API (down to `consul.peers` [in this list](https://docs.datadoghq.com/integrations/consul/#metrics)), which I configured by adding autodiscovery annotations to our server and client pods. According to the Consul overview on Datadog's Integration page:

> The Consul Agent can provide further metrics via DogStatsD.

The problem is, the chart doesn't easily allow me to configure DogStatsD. The two ways of configuring DogStatsD seem to be either [mounting a Unix Domain Socket](https://docs.datadoghq.com/agent/kubernetes/dogstatsd/#use-dogstatsd-over-a-unix-domain-socket) or [using a hostPort](https://docs.datadoghq.com/agent/kubernetes/dogstatsd/#alternatively-use-hostport). Respectively:

1. It's not possible to mount a volume using `hostPath` because the chart only support extra volumes that are either secrets or configmaps. Also, I don't think [Consul itself supports a file path](https://www.consul.io/docs/agent/options.html#telemetry-dogstatsd_addr) being passed as the value of `telemetry.dogstatsd_addr`
2. The only way it's possible currently to pass a `host:port` value to `telemetry.dogstatsd_addr` is by adding it to `extraConfig`, but that prevents us using `HOST_IP` since that isn't exposed in the container environment and since Consul can't be configured using environment variables in its config files.

The way we've approached this now is by exposing port 8125 on our datadog-agent pods as a service so we can hard-code the value of `telemetry.dogstatsd_addr` as `dogstatsd.namespace:8125` in `extraConfig`, but I think the ideal way of shipping metrics to DogStatsD is by sending them to the datadog-agent running on the same host as the sending pod ([Datadog say to configure DogStatsD in this way in their documentation](https://docs.datadoghq.com/agent/kubernetes/dogstatsd/#pass-the-node-s-ip-address-to-your-application)). One thing we've noticed is that we can't tie our metrics reliably to the k8s worker they originate from since a Consul pod is sending the metrics to a service, which resolves to one of a number of datadog-agent pods and not always the one running on the same k8s worker, and the `host` tag on the metrics seems to reflect the host the datadog-agent is deployed to, but for us there's still value is being able to see the cluster-wide status by looking at the  metrics in aggregate.

I thought the way that you would implement telemetry support would be by adding options to `values.yaml` such that you can toggle the command line flags for telemetry options, but it looks like the telemetry options can only be passed in configuration files ([see the Consul configuration docs](https://www.consul.io/docs/agent/options.html#configuration-files)). If that's the case, then I'm not sure how you would pass the right host IP to the pod, depending on the worker node it is deployed to?

Maybe it's possible to do something similar to vault-helm and render the config with the host IP? See: https://github.com/hashicorp/vault-helm/blob/master/templates/_helpers.tpl#L112-L124

wd:
Add codes to replace `HOST_IP` to the host ip address in config files.

Fix #339 ,  #657  .

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -165,6 +165,9 @@ spec:
             - |
               CONSUL_FULLNAME="{{template "consul.fullname" . }}"
 
+              mkdir -p /consul/config && cp /consul/config-from-cm/* /consul/config
+              find /consul/config -type f -name "*.json" -exec sed -Ei "s|HOST_IP|${HOST_IP?}|g" {} \;
+
               exec /bin/consul agent \
                 -node="${NODE}" \
                 -advertise="${ADVERTISE_IP}" \
@@ -229,7 +232,7 @@ spec:
             - name: data
               mountPath: /consul/data
             - name: config
-              mountPath: /consul/config
+              mountPath: /consul/config-from-cm
             {{- if .Values.global.tls.enabled }}
             - name: consul-ca-cert
               mountPath: /consul/tls/ca


lkysow:
Hi, I think we'd better follow the pattern that vault uses: https://github.com/hashicorp/vault-helm/blob/69a3dc618d6e5e96fa5a611a5a64f4f2388e03ad/templates/_helpers.tpl#L111

Also:
* needs to be done for the server statefulset
* needs bats tests

david-yu:
Hi there @wd , we've been able to make these changes in a separate [PR](https://github.com/hashicorp/consul-helm/pull/1042) which was recently merged. I'll go ahead and close this PR, let me know if you have other questions or concerns!

PR #739, opened 2020-12-11
Increase wait timeout for creating helm clusters

ndhanushkodi:
**Changes Proposed**
- Bumps the terratest module
- Changes the terratest module to use the fork with helm-extra-args changes, and use that to increase the wait timeout for creating helm clusters

I have them as 2 separate commits so the go.mod changes aren't confusing. 

Also these changes will apply to all tests using cluster.Create() but it will wait until k8s resources are ready OR until 15 min are up, so I think its ok to have this change here. Let me know in the review if you think it's better to do this based on a config flag and to only set it for tests where we think it'll help. We think this will help with AKS tests where mounting volumes is slow.


test/acceptance/framework/consul/consul_cluster.go
```diff
@@ -70,10 +70,18 @@ func NewHelmCluster(
 
 	logger := terratestLogger.New(logger.TestLogger{})
 
+	// Wait up to 15 min for K8s resources to be in a ready state. Increasing this
+	// from the default of 5 min could help with flakiness in environments like
+	// AKS where volumes take a long time to mount.
+	extraArgs := map[string][]string{
+		"install": {"--timeout", "15m", "--wait"},
```
lkysow, line 77: I don't think we should use `--wait` because we already have `WaitForAllPodsToBeReady` that gives us better logging about which pods aren't ready.

lkysow (approved):

kschoche (approved):
Second'ing Lukes comment on the wait but otherwise looks awesome!

test/acceptance/framework/consul/consul_cluster.go
ndhanushkodi: Yeah that makes sense, thanks!

PR #742, opened 2020-12-16
Document values.yaml for docs generation

lkysow:
Change how we document our values.yaml so it is compatible with docs
generation.

Annotations are explained here: https://github.com/hashicorp/consul-helm/blob/helm-docs/CONTRIBUTING.md#valuesyaml-annotations

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -1,76 +1,86 @@
 # Available parameters and their default values for the Consul chart.
 
-# global holds values that affect multiple components of the chart.
+# Holds values that affect multiple components of the chart.
 global:
-  # enabled is the master enabled/disabled setting.
-  # If true, servers, clients, Consul DNS and the Consul UI will be enabled.
-  # Each component can override this default via its component-specific
-  # "enabled" config.
-  # If false, no components will be installed by default and per-component
-  # opt-in is required, such as by setting `server.enabled` to true.
+  # The main enabled/disabled setting. If true, servers,
+  # clients, Consul DNS and the Consul UI will be enabled. Each component can override
+  # this default via its component-specific "enabled" config. If false, no components
+  # will be installed by default and per-component opt-in is required, such as by
+  # setting `server.enabled` to true.
   enabled: true
 
-  # name sets the prefix used for all resources in the helm chart.
-  # If not set, the prefix will be "<helm release name>-consul".
+  # Set the prefix used for all resources in the Helm chart. If not set,
+  # the prefix will be `<helm release name>-consul`.
+  # @type: string
   name: null
 
-  # domain is the domain Consul will answer DNS queries for
-  # (see https://www.consul.io/docs/agent/options.html#_domain) and the domain
-  # services synced from Consul into Kubernetes will have,
-  # e.g. `service-name.service.consul`.
+  # The domain Consul will answer DNS queries for
+  # (see `-domain` (https://consul.io/docs/agent/options#_domain)) and the domain services synced from
+  # Consul into Kubernetes will have, e.g. `service-name.service.consul`.
   domain: consul
 
-  # image is the name (and tag) of the Consul Docker image for clients and
-  # servers. This can be overridden per component.
-  # This should be pinned to a specific version tag, otherwise you may
-  # inadvertently upgrade your Consul version.
+  # The name (and tag) of the Consul Docker image for clients and servers.
+  # This can be overridden per component. This should be pinned to a specific
+  # version tag, otherwise you may inadvertently upgrade your Consul version.
   #
   # Examples:
-  #   # Consul 1.5.0
-  #   image: "hashicorp/consul:1.5.0"
-  #   # Consul Enterprise 1.5.0
-  #   image: "hashicorp/consul-enterprise:1.5.0-ent"
+  #
+  # ```yaml
+  # # Consul 1.5.0
+  # image: "consul:1.5.0"
+  # # Consul Enterprise 1.5.0
+  # image: "hashicorp/consul-enterprise:1.5.0-ent"
+  # ```
   image: "hashicorp/consul:1.9.0"
 
-  # array of objects containing image pull secret names that will be applied to
-  # each service account.
-  # This can be used to reference image pull secrets if using
-  # a custom consul or consul-k8s Docker image.
-  # See https://kubernetes.io/docs/concepts/containers/images/#using-a-private-registry.
+  # Array of objects containing image pull secret names that will be applied to each service account.
+  # This can be used to reference image pull secrets if using a custom consul or consul-k8s Docker image.
+  # See https://kubernetes.io/docs/concepts/containers/images/#using-a-private-registry for reference.
   #
   # Example:
-  #   imagePullSecrets:
+  #
+  # ```yaml
+  # imagePullSecrets:
   #   - name: pull-secret-name
   #   - name: pull-secret-name-2
+  # ```
+  # @type: array<map>
   imagePullSecrets: []
 
-  # imageK8S is the name (and tag) of the consul-k8s Docker image that
-  # is used for functionality such as catalog sync and connect. This can be overridden
-  # per component.
+  # The name (and tag) of the consul-k8s (https://github.com/hashicorp/consul-k8s)
+  # Docker image that is used for functionality such the catalog sync.
+  # This can be overridden per component.
   imageK8S: "hashicorp/consul-k8s:0.21.0"
 
-  # datacenter is the name of the datacenter that the agents should register
-  # as. This can't be changed once the Consul cluster is up and running
-  # since Consul doesn't support an automatic way to change this value
-  # currently: https://github.com/hashicorp/consul/issues/1858.
+  # The name of the datacenter that the agents should
+  # register as. This can't be changed once the Consul cluster is up and running
+  # since Consul doesn't support an automatic way to change this value currently:
+  # https://github.com/hashicorp/consul/issues/1858.
   datacenter: dc1
 
-  # enablePodSecurityPolicies controls whether pod
-  # security policies are created for the Consul components created by this
-  # chart. See https://kubernetes.io/docs/concepts/policy/pod-security-policy/.
+  # Controls whether pod security policies are created for the Consul components
+  # created by this chart. See https://kubernetes.io/docs/concepts/policy/pod-security-policy/.
   enablePodSecurityPolicies: false
 
-  # gossipEncryption configures which Kubernetes secret to retrieve Consul's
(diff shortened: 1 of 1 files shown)


values.yaml
```diff
@@ -79,140 +89,146 @@ global:
...
     secretKey: ""
 
-  # Enables TLS encryption across the cluster to verify authenticity of the
-  # servers and clients that connect. Note: It is HIGHLY recommended that you also
-  # enable Gossip encryption.
-  # See https://learn.hashicorp.com/consul/security-networking/agent-encryption
-  #
-  # Note: this relies on functionality introduced with Consul 1.4.1. Make sure
-  # your global.image value is at least version 1.4.1.
+  # Enables TLS (https://learn.hashicorp.com/tutorials/consul/gossip-encryption-secure)
```
ndhanushkodi, line 92:
```suggestion
  # Enables TLS (https://learn.hashicorp.com/tutorials/consul/tls-encryption-secure)
```
Since this talks about enabling TLS rather than gossip encryption specifically

values.yaml
```diff
@@ -79,140 +89,146 @@ global:
...
+  # Configure federation.
   federation:
     # If enabled, this datacenter will be federation-capable. Only federation
-    # through mesh gateways is supported.
+    # via mesh gateways is supported.
     # Mesh gateways and servers will be configured to allow federation.
-    # Requires global.tls.enabled, meshGateway.enabled and connectInject.enabled
+    # Requires `global.tls.enabled`, `meshGateway.enabled` and `connectInject.enabled`
     # to be true.
-    # Requires Consul 1.8+.
```
ndhanushkodi, line 199: Do we want to remove this? If true, it could be helpful to know.

values.yaml
```diff
@@ -222,25 +238,38 @@ global:
...
+  # Consul server agents.
+  # @type: string
   image: null
+
+  # The number of server agents to run. This determines the fault tolerance of
+  # the cluster. Please see the deployment table (https://consul.io/docs/internals/consensus#deployment-table)
+  # for more information.
   replicas: 3
 
   # bootstrapExpect is the number of servers that are expected to be running.
```
ndhanushkodi, line 275:
```suggestion
  # The number of servers that are expected to be running.
```
This looked like the kind of change that was being made everywhere, feel free to ignore if this is unnecessary here.

values.yaml
```diff
@@ -286,39 +342,98 @@ server:
...
       cpu: "100m"
 
-  # updatePartition is used to control a careful rolling update of Consul
-  # servers. This should be done particularly when changing the version
-  # of Consul. Please refer to the documentation for more information.
+  # This value is used to carefully
+  # control a rolling update of Consul server agents. This value specifies the
+  # partition (https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#partitions)
+  # for performing a rolling update. Please read the linked Kubernetes documentation
+  # for more information.
```
ndhanushkodi, line 349: Could we link https://www.consul.io/docs/k8s/upgrade#upgrading-consul-servers here as well?
```suggestion
  # This value is used to carefully
  # control a rolling update of Consul server agents. This value specifies the
  # partition (https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#partitions)
  # for performing a rolling update. Please read the linked Kubernetes documentation
  # and https://www.consul.io/docs/k8s/upgrade#upgrading-consul-servers for more information.
```

values.yaml
```diff
@@ -286,39 +342,98 @@ server:
...
-    #   load: false # if true, will add to `-config-dir` to load by Consul
-    #   items: # optional items array
-    #   - key: key
-    #     path: path
 
-  # Affinity Settings
-  # Commenting out or setting as empty the affinity variable, will allow
-  # deployment to single node services such as Minikube
+  # This value defines the affinity (https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity)
+  # for server pods. It defaults to allowing only a single pod on each node, which
```
ndhanushkodi, line 419:
```suggestion
  # for server pods. It defaults to allowing only a single server pod on each node, which
```

values.yaml
```diff
@@ -329,125 +444,168 @@ server:
...
+  # @type: map
   extraLabels: null
 
-  # Extra annotations to attach to the server pods.
-  # This should be a multi-line YAML string.
-  # Example:
-  #   annotations: |
-  #     "annotation-key": "annotation-value"
+  # This value defines additional annotations for
+  # server pods. This should be a formatted as a multi-line string.
```
ndhanushkodi, line 484:
```suggestion
  # server pods. This should be formatted as a multi-line string.
```

values.yaml
```diff
@@ -329,125 +444,168 @@ server:
...
+    # ```
+    #
+    # @type: string
     annotations: null
 
-  # extraEnvVars is a list of extra environment variables to set with the stateful set. These could be
-  # used to include proxy settings required for cloud auto-join feature,
-  # in case kubernetes cluster is behind egress http proxies. Additionally, it could be used to configure
-  # custom consul parameters.
+  # extraEnvironmentVars is a list of extra environment variables to set within the stateful set.
```
ndhanushkodi, line 507:
```suggestion
  # A list of extra environment variables to set within the stateful set.
```
Ignore if you feel this is unneccessary, just following the pattern of the other changes I see here.

values.yaml
```diff
@@ -461,71 +619,128 @@ client:
...
+  # ```
+  #
+  # @type: string
   annotations: null
 
-  # extraEnvVars is a list of extra environment variables to set with the pod. These could be
-  # used to include proxy settings required for cloud auto-join feature,
-  # in case kubernetes cluster is behind egress http proxies. Additionally, it could be used to configure
-  # custom consul parameters.
+  # extraEnvironmentVars is a list of extra environment variables to set within the stateful set.
```
ndhanushkodi, line 730:
```suggestion
  # A list of extra environment variables to set within the stateful set.
```
Ignore if unnecessary :)

values.yaml
```diff
@@ -622,18 +857,30 @@ ui:
...
+    # @type: string
     annotations: null
 
     # Additional ServiceSpec values
     # This should be a multi-line string mapping directly to a Kubernetes
     # ServiceSpec object.
+    # @type: string
     additionalSpec: null
 
 # syncCatalog will run the catalog sync process to sync K8S with Consul
```
ndhanushkodi, line 886:
```suggestion
# Configure the catalog sync process to sync K8S with Consul
```

values.yaml
```diff
@@ -800,16 +1076,25 @@ syncCatalog:
   logLevel: info
 
   # Override the default interval to perform syncing operations creating Consul services.
+  # @type: string
   consulWriteInterval: null
 
 # ConnectInject will enable the automatic Connect sidecar injector.
```
ndhanushkodi, line 1082:
```suggestion
# Configures the automatic Connect sidecar injector.
```

ndhanushkodi (approved):
Most of my comments are little nits for style and are non-blocking if you want to address them in a future PR instead. I like how the comments read now without all starting with the name of the key!

values.yaml
lkysow: good catch, adding back

values.yaml
lkysow: No good catch, this was still here because I rebased off of master and this was a new doc block that was written in the old style.

values.yaml
lkysow: 👍

values.yaml
```diff
@@ -587,6 +814,7 @@ client:
 # for this to have any effect:
 # https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/#configure-stub-domain-and-upstream-dns-servers
 dns:
+  # @type: boolean
   enabled: "-"
 
   # type can be used to control the type of service created. For
```
ndhanushkodi, line 820:
```suggestion
  # Used to control the type of service created. For
```

values.yaml
```diff
@@ -648,35 +895,49 @@ syncCatalog:
...
+  # synced by default. If false, the service must be annotated
+  # (https://consul.io/docs/k8s/service-sync#sync-enable-disable) properly to sync.
+  # In either case an annotation can override the default.
+  default: true
 
   # Optional priorityClassName.
   priorityClassName: ""
 
   # toConsul and toK8S control whether syncing is enabled to Consul or K8S
   # as a destination. If both of these are disabled, the sync will do nothing.
```
ndhanushkodi, line 914:
```suggestion
  # If true, will sync Kubernetes services to Consul. This can be disabled to
  # have a one-way sync. If both of `syncCatalog.toConsul` and `syncCatalog.toK8s`
  # are disabled, the sync will do nothing.
```

values.yaml
```diff
@@ -648,35 +895,49 @@ syncCatalog:
...
   toConsul: true
+
+  # If true, will sync Consul services to Kubernetes. This can be disabled to
+  # have a one-way sync.
   toK8S: true
 
   # k8sPrefix is the service prefix to prepend to services before registering
```
ndhanushkodi, line 921: From here on down I see quite a few old style comments starting with the yaml key, just wanted to point out in case you wanted to fix those as well

ndhanushkodi (approved):
I had remembered seeing a few more (especially towards the end of the file) old-style comments and wanted to call out where I saw most of them since I saw you were fixing them from my first review!

values.yaml
```diff
@@ -1,76 +1,86 @@
...
+  # Controls whether pod security policies are created for the Consul components
+  # created by this chart. See https://kubernetes.io/docs/concepts/policy/pod-security-policy/.
   enablePodSecurityPolicies: false
 
-  # gossipEncryption configures which Kubernetes secret to retrieve Consul's
-  # gossip encryption key from (see https://www.consul.io/docs/agent/options.html#_encrypt).
-  # If secretName or secretKey are not set, gossip encryption will not be enabled.
-  # The secret must be in the same namespace that Consul is installed into.
+  # Configures which Kubernetes secret to retrieve Consul's
+  # gossip encryption key from (see [-encrypt](/docs/agent/options#_encrypt)). If secretName or
```
ishustava, line 66: I'm curious why does the link for `-encrypt` look different from the one in `domain` i.e. it has `[]`?

ishustava (approved):
Awesome! Great work on that generation, Luke!

values.yaml
lkysow: yeah this was a mistake. We need to make the urls absolute so folks just reading the yaml can use them.

lkysow:
>I had remembered seeing a few more (especially towards the end of the file) old-style comments and wanted to call out where I saw most of them since I saw you were fixing them from my first review!

Thanks! Fixed them up. I think I got lazy during the hackathon 😆

PR #747, opened 2020-12-18
If @type: map show the type

lkysow:
Fix issue where we weren't showing the type even if it was explicitly
set to map.


ishustava (approved):
🎉

ndhanushkodi (approved):
🚀

PR #311, opened 2019-12-12
allow for the uid and gid to be defined

sbockelman:
... for the server and client components so it does not default to root

PR #748, opened 2020-12-18
Run Consul servers and clients as non-root and make it configurable.

ishustava:
Based on work done in #311 

## Changes proposed in this PR

Add new values `server.securityContext` and `client.securityContext` to allow configuring securityContext setting for clients and servers. The map directly to pod's security context in Kubernetes.

We also default these settings to running both servers and clients as non-root and with the `consul` uid and gid, which are created in the Consul docker image.

When openshift is enabled, we don't want these settings to be there since openshift generates random user and group and sets them for each pod.

## Testing

No need for the reviewers to test since passing acceptance tests is enough of a validation

## TODO

- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ BREAKING CHANGES:
 * Setting `server.bootstrapExpect` to a value less than `server.replicas` will now
   give an error. This was a misconfiguration as the servers wouldn't wait
   until the proper number have started before electing a leader. [[GH-721](https://github.com/hashicorp/consul-helm/pull/721)]
+* Clients and servers now run as non root. Users can also configure `server.securityContext` and `client.securityContext`
+  if they wish to overwrite this behavior. Please see [Helm reference](https://www.consul.io/docs/k8s/helm) for more information.
+  [[GH-748](https://github.com/hashicorp/consul-helm/pull/748)]
 
 FEATURES:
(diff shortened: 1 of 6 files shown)


thisisnotashwin (approved):
This looks great!! 🥳

lkysow (approved):
Nice!!

lawliet89:
Is there any guidance for upgrading Consul servers to the new version of the chart wrt this change? 

When I `ls` the `/consul/data` directory, I get:

```
/ # ls -al /consul/data/
total 44
drwxrwsr-x    6 root     consul        4096 Mar 26  2019 .
drwxr-xr-x    1 consul   consul        4096 Dec  9 22:07 ..
-rw-rw-r--    1 root     consul         394 Mar 26  2019 checkpoint-signature
drwxrws---    2 root     consul       16384 Mar 26  2019 lost+found
-rw-rw----    1 root     consul          36 Mar 26  2019 node-id
drwxrws---    2 root     consul        4096 Oct 17  2019 proxy
drwxrwsr-x    3 root     consul        4096 Mar 26  2019 raft
drwxrwsr-x    2 root     consul        4096 Nov 25 03:58 serf
```

Based on my understanding, GID 1000 is `consul` and UID 100 is also `consul`, so theoretically the server pods can continue to read and write to `/consul/data`.

PR #749, opened 2020-12-18
Use consul-k8s subcommand to perform TLS-init job

thisisnotashwin:
Changes proposed in this PR:
- Use `tls-init` subcommand from `consul-k8s` https://github.com/hashicorp/consul-k8s/pull/410 to perform the creation of TLS server certificates.

How I've tested this PR:
- Perform a helm install/upgrade whilst running the new tls-init job. It creates the desired server certificates and rotates them when run subsequently. Consul servers need to be restarted in order for them to use the new certificates. It does not update the CA certs once they are created for the first time or provided ensuring the server certificates are always signed by the same CA.

- Perform a helm install. Update the values file to add `additionalDNSSans` and perform a helm upgrade.
`openssl s_client -connect localhost:8501` when performed from with a server pod should initially return a certificate that does not have the provided additional DNS sans. When you restart the server, the servers should continue to be able to talk to each other and and the above `s_client` operation should return an updated certificate which has the new DNS SANS.

- The above operation can be retried but with a BYO CA and key.

Checklist:
- [x] Bats tests added
  - Link to PR: https://github.com/hashicorp/consul/pull/9439

### DIFF
--- a/templates/tls-init-job.yaml
+++ b/templates/tls-init-job.yaml
@@ -46,42 +46,25 @@ spec:
       {{- end }}
       containers:
         - name: tls-init
-          image: "{{ .Values.global.image }}"
+          image: "{{ .Values.global.imageK8S }}"
           env:
             - name: NAMESPACE
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
-          # We're using POST requests below to create secrets via Kubernetes API.
-          # Note that in the subsequent runs of the job, POST requests will
-          # return a 409 because these secrets would already exist;
-          # we are ignoring these response codes.
           workingDir: /tmp
           command:
             - "/bin/sh"
             - "-ec"
             - |
-              {{- if (not (and .Values.global.tls.caCert.secretName .Values.global.tls.caKey.secretName)) }}
-              consul tls ca create \
-                -domain={{ .Values.global.domain }}
-              curl -s -X POST --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
-                https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/${NAMESPACE}/secrets \
-                -H "Authorization: Bearer $( cat /var/run/secrets/kubernetes.io/serviceaccount/token )" \
-                -H "Content-Type: application/json" \
-                -H "Accept: application/json" \
-                -d "{ \"kind\": \"Secret\", \"apiVersion\": \"v1\", \"metadata\": { \"name\": \"{{ template "consul.fullname" . }}-ca-cert\", \"namespace\": \"${NAMESPACE}\" }, \"type\": \"Opaque\", \"data\": { \"tls.crt\": \"$( cat {{ .Values.global.domain }}-agent-ca.pem | base64 | tr -d '\n' )\" }}" > /dev/null
-              curl -s -X POST --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
-                https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/${NAMESPACE}/secrets \
-                -H "Authorization: Bearer $( cat /var/run/secrets/kubernetes.io/serviceaccount/token )" \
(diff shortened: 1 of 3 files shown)


kschoche (approved):
🎉 
🥇

templates/tls-init-job.yaml
```diff
@@ -61,27 +61,11 @@ spec:
             - "/bin/sh"
             - "-ec"
             - |
-              {{- if (not (and .Values.global.tls.caCert.secretName .Values.global.tls.caKey.secretName)) }}
```
lkysow, line 64: There's a comment above here we need to remove

templates/tls-init-job.yaml
```diff
@@ -96,16 +80,8 @@ spec:
                 {{- end }}
                 {{- range .Values.global.tls.serverAdditionalDNSSANs }}
                  -additional-dnsname={{ . }} \
```
lkysow, line 82:
```suggestion
                -additional-dnsname={{ . }} \
```

templates/tls-init-job.yaml
```diff
@@ -61,27 +61,11 @@ spec:
...
-              curl -s -X POST --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
-                https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/${NAMESPACE}/secrets \
-                -H "Authorization: Bearer $( cat /var/run/secrets/kubernetes.io/serviceaccount/token )" \
-                -H "Content-Type: application/json" \
-                -H "Accept: application/json" \
-                -d "{ \"kind\": \"Secret\", \"apiVersion\": \"v1\", \"metadata\": { \"name\": \"{{ template "consul.fullname" . }}-ca-key\", \"namespace\": \"${NAMESPACE}\" }, \"type\": \"Opaque\", \"data\": { \"tls.key\": \"$( cat {{ .Values.global.domain }}-agent-ca-key.pem | base64 | tr -d '\n' )\" }}" > /dev/null
-              {{- end }}
-              # Suppress globbing so we can interpolate the $NAMESPACE environment variable
-              # and use * at the start of the dns name when setting -additional-dnsname.
-              set -o noglob
```
lkysow, line 82: Do we still need the `set -o noglob`?

Everything seems to be working even if I use 
```
    serverAdditionalDNSSANs: ["*consul"]
```

But just wondering why that's not causing issues.

ashwin-venkatesh: That was deleted by mistake. Not sure why that isn't failing honestly? I should add it back along with the comment but maybe we did not need it in the first place? 🤷

lkysow: It looks like `sh` doing globbing but maybe the flag parsing is ignoring stuff somehow? I think we may as well keep it to be safe.

```shell-session
$ ls consul*
consul-agent-ca-key.pem  consul-agent-ca.pem

consul:
config  data    tls

$ echo consul*
consul consul-agent-ca-key.pem consul-agent-ca.pem

$ echo consul tls cert create -server -additional-dnsname consul*
consul tls cert create -server -additional-dnsname consul consul-agent-ca-key.pem consul-agent-ca.pem

$ consul tls cert create -server -additional-dnsname consul* && openssl x509 -in dc1-server-consul-0.pem -noout -text
...
            X509v3 Subject Alternative Name:
                DNS:consul, DNS:server.dc1.consul, DNS:localhost, IP Address:127.0.0.1
```

lkysow (approved):
🎉

templates/tls-init-job.yaml
ashwin-venkatesh: yup. ive restored it.

HofmannZ:
@thisisnotashwin

Running v0.28.0 `tls-init` works.
Running master `tls-init` fails with errors about a wrong command bing executed.

How can I test this feature?

lkysow:
Hi @HofmannZ you would need to use a docker image build of `consul-k8s` from `master`. We build `hashicorpdev/consul-k8s:latest` off of `master` so you could do:

```
global:
  imageK8S: "hashicorpdev/consul-k8s:latest"
```

Note that obviously that image isn't stable, nor is `consul-helm` on `master` so this should only be for testing until we do a full release.

PR #753, opened 2020-12-23
Add docs gen annotations to resource settings

ishustava:
Changes proposed in this PR:
- Update `resources` values key throughout with the `@recurse: false`. This is how it is currently and it was probably omitted when we started generating docs
- Add explicit defaults to various image values. Instead of using the actual default, we use something like `envoyproxy/envoy-alpine:<latest supported version>` to avoid having to update docs for version changes.
- Add missing docs for some values.

Checklist:
  - Link to PR: https://github.com/hashicorp/consul/pull/9463

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -31,6 +31,7 @@ global:
   # # Consul Enterprise 1.5.0
   # image: "hashicorp/consul-enterprise:1.5.0-ent"
   # ```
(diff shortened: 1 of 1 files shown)


ndhanushkodi (approved):
Looks great to me!

PR #756, opened 2021-01-05
Increase timeout for checking the static server connection.

ndhanushkodi:
The mesh gateway tests intermittently fail while checking this
connection with connection reset by peer errors. This could potentially
help.

Changes proposed in this PR:
- Increase timeout for checking static server connection, since it fails sometimes (2x in 14 days)
- For background context: EKS [test investigation](https://docs.google.com/document/d/1j88SdxLz4eLW2iBPsQhf-2CGMvVQCITgQxf3M1ia5m0/edit) leading to this PR

How I've tested this PR:
- running eks tests on this branch

- code review

### DIFF
--- a/test/acceptance/framework/k8s/deploy.go
+++ b/test/acceptance/framework/k8s/deploy.go
@@ -104,7 +104,7 @@ func CheckStaticServerConnectionMultipleFailureMessages(
 ) {
 	t.Helper()
 
-	retrier := &retry.Timer{Timeout: 20 * time.Second, Wait: 500 * time.Millisecond}
+	retrier := &retry.Timer{Timeout: 80 * time.Second, Wait: 2 * time.Second}
 
 	args := []string{"exec", "deploy/" + deploymentName, "-c", deploymentName, "--", "curl", "-vvvsSf"}
 	args = append(args, curlArgs...)


lkysow:
What do you think about adding a check for what the primary thinks is running in dc2?

```go
	// Verify federation between servers
	logger.Log(t, "verifying federation was successful")
	verifyFederation(t, primaryClient, secondaryClient, releaseName, false)

	defer func() {
		svcs, _, err := primaryClient.Catalog().Services(&api.QueryOptions{Datacenter: "dc2"})
		if err != nil {
			fmt.Printf("error calling primary on /catalog?dc=dc2: %s\n", err.Error())
			return
		}
		fmt.Printf("primary on /catalog?dc=dc2: %+v\n", svcs)
	}()

```

ndhanushkodi:
> What do you think about adding a check for what the primary thinks is running in dc2?

Great idea!! Giving it a shot [here](https://github.com/hashicorp/consul-helm/pull/757)

kschoche (approved):
Sounds good. Would love for us to come up with a better number, but I understand this is challenging.
A 10x differential for timeouts between one cloud provider and another is kinda nuts!

lkysow (approved):

PR #762, opened 2021-01-08
Expose server serfwan port as hostPort when hostIP is advertised.

ndhanushkodi:
The servers try to talk to each other on the serfwan port, and when the
advertised IP of the server is the host IP, the serfwan port needs to be
exposed so the server doesn't log wan gossip errors.

Changes proposed in this PR:
- When exposeGossipAndRPCPorts is set, the server's serfwan port is also exposed as a hostPort

How I've tested this PR:
- enable exposeGossipAndRPCPorts, and ensure that there are no longer errors being logged.
- EDIT: not sure if this actually worked since I didn't see the error without this change.

- Code review

Checklist:
- [x] Bats tests added

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -255,6 +255,9 @@ spec:
               protocol: "UDP"
               name: serflan-udp
             - containerPort: 8302
+              {{- if .Values.server.exposeGossipAndRPCPorts }}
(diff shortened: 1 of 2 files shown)


ndhanushkodi:
Closing, since we don't see errors with servers talking to eachother when hostIP is advertised.

PR #764, opened 2021-01-13
Remove preStop hook that calls consul leave from Consul servers

ishustava:
## Background
Currently, when Consul servers are restarted, we call the `consul leave` command in the `preStop` hook that is registered with the servers. This approach has a couple of problems:

1. This command does not work when ACLs are enabled on the cluster because this command requires a valid ACL token with `agent:write` permission. Furthermore, we cannot reasonably make it work because even if the server's ACL token had that permission, we could not retrieve it because it's stored in the Consul's data directory. We could work around this by removing the preStop hook and instead setting `leave_on_terminate` to `true` on the servers, which would accomplish the same behavior. However, this solution is not recommended because of point 2.
2. The voting members of the consul cluster (aka the consul servers) should not gracefully leave the cluster. That is so that the quorum size is not affected during upgrades or downtime. See hashicorp/consul#5947.

## Changes proposed in this PR:
- Remove the preStop hook calling `consul leave` from the consul server pods

## How I've tested this PR:

- Deployed Consul from this helm chart, ran `kubectl rollout restart sts/...`, and checked that servers came up healthy.
- Deployed the current stable version of the helm chart, upgraded to this version and checked that the upgrade goes through without problems

## How I expect reviewers to test this PR:
Code review

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,25 +1,9 @@
 ## Unreleased
 
-BREAKING CHANGES:
-* Consul servers start with a provided node ID and no longer call "consul leave" command when restarted or deleted. [[GH-764](https://github.com/hashicorp/consul-helm/pull/764)]
-  If you have an existing Consul cluster with ACLs enabled, please follow the steps below up upgrade.
-  Otherwise, no additional steps are necessary to upgrade.
-  1. Set `leave_on_terminate` on your existing installation, setting `leave_on_terminate` to `true` so that the new node ID can be registered:
-
-     ```bash
-     helm upgrade consul --set server.extraConfig='\"{"leave_on_terminate": true}\"' hashicorp/consul --version <your current version>
-     ```
-
-  2. Upgrade to this version of the Helm chart, removing the `leave_on_terminate` setting:
-
-     ```bash
-     helm repo update
-     helm upgrade consul hashicorp/consul
-     ```
-
 IMPROVEMENTS:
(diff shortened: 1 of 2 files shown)


kschoche (approved):
Great work @ishustava !
I'd personally call it a bug fix but all looks great, thanks for taking this on.

ishustava:
> I'd personally call it a bug fix but all looks great, thanks for taking this on.

good point, @kschoche. Let me fix that.

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ IMPROVEMENTS:
 * Use `consul-k8s` subcommand to perform `tls-init` job. This allows for server certificates to get rotated on subsequent runs.
   Consul servers have to be restarted in order for them to update their server certificates [[GH-749](https://github.com/hashicorp/consul-helm/pull/721)]
 
+BUG FIXES:
+* Consul servers no longer call `consul leave` command when restarted or deleted. [[GH-764](https://github.com/hashicorp/consul-helm/pull/764)]
```
lkysow, line 8: 🚗 could we also say that this is okay because we don't want them to leave anyway since that reduces the quorum size.

ishustava: Ah, good point, will add! Always love the 🚗 comments 😄

ishustava: Updated!

ndhanushkodi (approved):
🎉 looks great!

PR #766, opened 2021-01-14
Re-enable OpenShift tests

lkysow:
Using new mirror

NOTE: AKS tests are still failing, need to debug further.

### DIFF
--- a/test/acceptance/tests/fixtures/bases/static-client/deployment.yaml
+++ b/test/acceptance/tests/fixtures/bases/static-client/deployment.yaml
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
         - name: static-client
-          image: tutum/curl:latest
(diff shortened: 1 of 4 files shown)


test/acceptance/tests/fixtures/bases/static-client/deployment.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
         - name: static-client
-          image: tutum/curl:latest
+          image: docker.mirror.hashicorp.services/curlimages/curl:latest
```
lkysow, line 18: This is a more up-to-date and maintained image. `tutum/curl` was so old that our mirror couldn't mirror it because it had very old metadata

kschoche (approved):
nicely done!

ndhanushkodi (approved):

PR #770, opened 2021-01-19
Use GKE 1.19.x

lkysow:
Test against GKE 1.19


lkysow:
🛑 we're actually going to run kind tests with 1.19 and 1.20 instead since 1.16 is the default gke version

PR #130, opened 2019-03-10
Add ingress option to UI service.

zinref:
Regarding to https://github.com/hashicorp/consul-helm/issues/88. Then, I have introduced the ingress option to be configurable.

PR #774, opened 2021-01-20
Add ingress option

thisisnotashwin:
Changes proposed in this PR:
- Based on https://github.com/hashicorp/consul-helm/pull/130
- Added support to expose http port when `global.tls.enabled` is true and `global.tls.httpsOnly` is `false`.

The ingress resource always performs TLS termination at the point of ingress and can only communicate with the Consul server over the HTTP port. This can be avoided by certain ingress controllers that have the option of [allowing ssl pass-through](https://kubernetes.github.io/ingress-nginx/user-guide/tls/#ssl-passthrough). Added a note to ensure that users have an ingress controller that can passthrough SSL and have it enabled if they want to enable TLS on Consul.

How I've tested this PR:
- Deployed an ingress controller on the kubernetes cluster (nginx).
- Performed a helm deploy with ingress enabled for UI. Verified the fields on the ingress resource created for non-TLS and TLS (with HTTP enabled) use cases and was able to access the UI via the created Ingress. (GKE creates a LoadBalancer for ingress)

- Code review.

Checklist:
- [x] Bats tests added
  - Link to PR: https://github.com/hashicorp/consul/pull/9612

### DIFF
--- a/templates/ui-ingress.yaml
+++ b/templates/ui-ingress.yaml
@@ -1,0 +1,31 @@
+{{- if (and (or (and (ne (.Values.server.enabled | toString) "-") .Values.server.enabled) (and (eq (.Values.server.enabled | toString) "-") .Values.global.enabled)) (or (and (ne (.Values.ui.enabled | toString) "-") .Values.ui.enabled) (and (eq (.Values.ui.enabled | toString) "-") .Values.global.enabled)) (or (and (ne (.Values.ui.ingress.enabled | toString) "-") .Values.ui.ingress.enabled) (and (eq (.Values.ui.ingress.enabled | toString) "-") .Values.global.enabled))) }}
+{{- $serviceName := printf "%s-%s" (include "consul.fullname" .) "ui" -}}
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: {{ template "consul.fullname" . }}-ingress
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "consul.name" . }}
+    chart: {{ template "consul.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  {{- if .Values.ui.ingress.annotations }}
+  annotations:
+    {{ tpl (toYaml .Values.ui.ingress.annotations) . | nindent 4 | trim }}
+  {{- end}}
+spec:
+  rules:
+  {{- range .Values.ui.ingress.hosts }}
+    - host: {{ . }}
+      http:
+        paths:
+          - backend:
+              serviceName: {{ $serviceName }}
+              servicePort: 80
+  {{- end -}}
+  {{- if .Values.ui.ingress.tls }}
+  tls:
+    {{ tpl (toYaml .Values.ui.ingress.tls) . | nindent 4 | trim }}
+  {{- end }}
+{{- end }}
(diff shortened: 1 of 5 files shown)


values.yaml
```diff
@@ -941,6 +941,51 @@ ui:
...
+    #     paths:
+    #       - /example
+    #       - /test
+    # ```
+    #
+    # @type: array<map>
+    hosts: []
+
+    # tls is a list of hosts and secret name in an Ingress
+    # which tells the Ingress controller to secure the channel.
```
lkysow, line 965: Should we tell them what secret to use? Do they need to use passthrough TLS if consul tls is enabled?

lkysow (commented):
I'm wondering about the version of the object we're using:
```
W0121 13:39:33.134204   34932 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
```

thisisnotashwin:
@lkysow re version: I can switch over to version `v1`. This was the version on the initial PR but i think it makes sense to up it given it has been supported since 1.14

test/unit/ui-ingress.bats
```diff
@@ -0,0 +1,166 @@
...
+  local actual=$(helm template \
+      -s templates/ui-ingress.yaml  \
+      --set 'ui.ingress.enabled=true' \
+      --set 'ui.ingress.hosts[0].host=foo.com' \
+      . | tee /dev/stderr |
+      yq -r '.spec.rules[0].host' | tee /dev/stderr)
+  [ "${actual}" = "foo.com" ]
+}
+
+@test "ui/Ingress: port 80 when global.tls.enabled=false enables http port" {
```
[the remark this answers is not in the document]
thisisnotashwin: omg i just read the test names and i admit they are horrible. copy-pasting with proof-reading. this is why we review!

values.yaml
thisisnotashwin: So i realized in my tinkering that i hadn't tested the only TLS (https only enabled) options. ie, the 80 port was always open on the service. the ingress resource always performs TLS termination so we cant point traffic to 443 as the request that reached the service will always be HTTP. I guess it doesnt make sense for this PR to support port 443 in that case, and ill make the requisite changes.

templates/ui-ingress.yaml
```diff
@@ -0,0 +1,48 @@
...
+    {{ tpl .Values.ui.ingress.annotations . | nindent 4 | trim }}
+  {{- end}}
+spec:
+  rules:
+  {{ $global := .Values.global }}
+  {{- range .Values.ui.ingress.hosts }}
+  - host: {{ .host | quote }}
+    http:
+      paths:
+      {{- if (or (not $global.tls.enabled) (not $global.tls.httpsOnly)) }}
```
ryan-dyer-sp, line 26: The only difference I'm seeing between L27-32 and L35-49 is the servicePort.  So why not just make that the conditional instead of the entire block?

thisisnotashwin: I guess the only motivation there would be in a situation when TLS is enabled but both HTTP and HTTPs traffic is allowed, we would need both the block. I can look at DRY-ing this up though.

lkysow (approved):

templates/ui-ingress.yaml
ryan-dyer-sp: I see your point from a what's allowed perspective, but not sure why someone would create an Ingress and allow port 80 traffic, if 443 is available for the UI.  I would _think_ that the primary reason for having 80 and 443 enabled on consul is because they are migrating from 80 to 443 and havent finished shoring up the clients still using 80.

ryan-dyer-sp: And in the case where they are both enabled how does the ingress know which backend to send to if they are defined the exact same way but with different servicePort? This is just a general curiousity question, if the path were different, I could see how it would work.  But I'd almost think this would be an invalid configuration for an Ingress.

kschoche (approved):
🧨

templates/ui-ingress.yaml
thisisnotashwin: In case the Ingress controller does not support TLS passthrough, only supporting port 443 will not work. In the case when both ports are exposed, the controller multiplexes the requests across the ports and responds with whichever response is successful. 
That is a really good questions though. I tested it with both ports open without the passthrough and it worked like a treat.

thisisnotashwin: This allows for maximum flexibility but ensures the ability to expose just the one port (`443`) if that is so desired. Did I understand your question correctly?

Issue #771, opened 2021-01-19
Helm makes HTTP call at the HTTPS port when using external Consul cluster.


jimm-with-a-j:
When filing a bug, please include the following headings if possible. Any example text in this template can be deleted.

### Overview of the Issue

Working with local dev consul agent and a K8s cluster in Minikube I was trying to use the helm chart and noticed it was trying to communicate with the Consul API over HTTP but tries to use the HTTPS port. I didn't see any values to change this to use HTTPS so I had to 'lie' and set the helm value for httpsPort to 8500 (the HTTP port) and then everything went through.

### Reproduction Steps

Install or upgrade helm chart for client only with external Consul server. Make sure the API is accessible to K8s. When setting the externalServers if you leave the defaults it will try to communicate with Consul API at the HTTPS port but sends HTTP requests so it fails. If you then set the httpsPort value to be the HTTP port it works.

1. When running helm install with the following `values.yml`:
```
# config.yaml
global:
  enabled: false
  acls:
    manageSystemACLs: true
    bootstrapToken:
      secretName: bootstrap-token-2
      secretKey: token
  tls:
    enableAutoEncrypt: true

client:
  enabled: true
  # Set this to true to expose the Consul clients using the Kubernetes node
  # IPs. If false, the pod IPs must be routable from the external servers.
  grpc: true
  join:
    - "192.168.86.249"

connectInject:
  enabled: true
  default: true
  
externalServers:
  enabled: true
  hosts:
    - "192.168.86.249"
  k8sAuthMethodHost: 'https://XXX:8443'
```
If changed to the below it works:
```
externalServers:
  enabled: true
  httpsPort: 8500
  hosts:
    - "192.168.86.249"
  k8sAuthMethodHost: 'https://XXX:8443'
```
1. View error

### Logs

<details>
  <summary>Logs</summary>

```
2021-01-19T19:25:05.589Z [INFO]  Bootstrap token is provided so skipping Consul server ACL bootstrapping
2021-01-19T19:25:05.593Z [ERROR] Failure: calling /agent/self to get datacenter: err="Unexpected response code: 400 (Client sent an HTTP request to an HTTPS server.
)"
2021-01-19T19:25:05.593Z [INFO]  Retrying in 1s
2021-01-19T19:25:06.595Z [ERROR] Failure: calling /agent/self to get datacenter: err="Unexpected response code: 400 (Client sent an HTTP request to an HTTPS server.
)"
[...]

PR #775, opened 2021-01-21
Fail if autoencrypt true but tls false

lkysow:
Give an error if a user enables autoencrypt but forgets to set
tls.enabled to true. This helps them notice a misconfiguration.

Checklist:
- [x] Bats tests added
- [ ] PR submitted to update [consul.io Helm docs](https://www.consul.io/docs/k8s/helm) (see [Generating Helm Reference Docs](https://github.com/hashicorp/consul-helm/blob/master/CONTRIBUTING.md#generating-helm-reference-docs))
  - Link to PR:

### DIFF
--- a/templates/tls-init-job.yaml
+++ b/templates/tls-init-job.yaml
@@ -1,4 +1,5 @@
 {{- if (or (and (ne (.Values.server.enabled | toString) "-") .Values.server.enabled) (and (eq (.Values.server.enabled | toString) "-") .Values.global.enabled)) }}
+{{- if (and (not .Values.global.tls.enabled) .Values.global.tls.enableAutoEncrypt) }}{{ fail "if global.tls.enableAutoEncrypt is true, global.tls.enabled must also be true" }}{{ end -}}
 {{- if .Values.global.tls.enabled }}
 # tls-init job generate Consul cluster CA and certificates for the Consul servers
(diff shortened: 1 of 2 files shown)


lkysow:
@ishustava is there a use-case for
```
global:
  tls:
    enabled: false
    enableAutoEncrypt: true
```

?

Related to https://github.com/hashicorp/consul-helm/issues/771

ishustava:
@lkysow 

One use case that comes to mind is when you are trying to just switch from tls enabled to disabled while also having `enableAutoEncrypt` set to true? The error would force you to have to change that too in your config and that could be frustrating for a case when you are just trying to test out different behavior.

This also brings up an interesting question of whether this breaks the pattern for other similar situations with nested settings we have and how they interact with the `enabled` setting within that one section. For example, should we also fail if you set `server.enabled` to `false` but `server.exposeGossipAndRPCPorts` to `true`, or if you set `client.enabled` to `false` but `client.grpc` to `true`? 

What about non-boolean values? For example, is it invalid to set `tls.caCert` but not `tls.enabled` or `acls.replicationToken` but not `acls.manageSystemACLs`?

lkysow:
Decided to close because if the top-level `enabled` is set to `false` then we don't want to validate anything under that key.

PR #786, opened 2021-01-25
fix api version for ingress

### LINKED ISSUE #785: UI ingress doesn't work with kubernetes v1.19
### Overview of the Issue

`ui.ingress` resource doesn't work with kubernetes v1.19
This appears to be due to issues using the wrong api version. Depending on the cluster version, this should probably be `networking.k8s.io/v1beta1` or `networking.k8s.io/v1beta1`, but not `extensions/v1`

### Reproduction Steps

Steps to reproduce this issue, eg:

1. Deploy the chart into a cluster running kubernetes v.1.19

### Logs

<details>
  <summary>Logs</summary>

```
Error: unable to recognize "": no matches for kind "Ingress" in version "extensions/v1"
```

</details>

### Expected behavior

The helm install or upgrade should succeed when `ui.ingress` is enabled

### Additional Context

https://kubernetes.io/docs/concepts/services-networking/ingress/

thisisnotashwin:
Changes proposed in this PR:
- Fix API version of Ingress resource

How I've tested this PR: deployed against cluster with ingress controller

How I expect reviewers to test this PR: 😭 yall know what to do

Fixes #785 

Checklist:
- [ ] Bats tests added

### DIFF
--- a/templates/ui-ingress.yaml
+++ b/templates/ui-ingress.yaml
@@ -1,7 +1,7 @@
 {{- if (and (or (and (ne (.Values.server.enabled | toString) "-") .Values.server.enabled) (and (eq (.Values.server.enabled | toString) "-") .Values.global.enabled)) (or (and (ne (.Values.ui.enabled | toString) "-") .Values.ui.enabled) (and (eq (.Values.ui.enabled | toString) "-") .Values.global.enabled)) (or (and (ne (.Values.ui.service.enabled | toString) "-") .Values.ui.service.enabled) (and (eq (.Values.ui.service.enabled | toString) "-") .Values.global.enabled))) }}
 {{- if (and (ne (.Values.ui.ingress.enabled | toString) "-") .Values.ui.ingress.enabled) }}
 {{- $serviceName := printf "%s-%s" (include "consul.fullname" .) "ui" -}}
-{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
+{{- if or ( gt .Capabilities.KubeVersion.Major "1" ) ( ge .Capabilities.KubeVersion.Minor "19" ) }}
 apiVersion: networking.k8s.io/v1
 {{- else }}
 apiVersion: networking.k8s.io/v1beta1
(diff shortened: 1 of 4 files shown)


ishustava (approved):

lkysow:
```
error: error validating "ing.yaml": error validating data: [ValidationError(Ingress.spec.rules[0].http.paths[0].backend): unknown field "serviceName" in io.k8s.api.networking.v1.IngressBackend, ValidationError(Ingress.spec.rules[0].http.paths[0].backend): unknown field "servicePort" in io.k8s.api.networking.v1.IngressBackend]; if you choose to ignore these errors, turn validation off with --validate=false
```

with
```yaml
ui:
  ingress:
    enabled: true
    hosts:
      - host: foo.bar
        paths:
          - /consul
```

I think the newer ingress object has different fields:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: minimal-ingress
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /
spec:
  rules:
  - http:
      paths:
      - path: /testpath
        pathType: Prefix
        backend:
          service:
            name: test
            port:
              number: 80
```

thisisnotashwin:
@lkysow updated PR based on https://github.com/kubernetes/kubernetes/issues/90077#issuecomment-718778960

templates/ui-ingress.yaml
```diff
@@ -1,7 +1,11 @@
 {{- if (and (or (and (ne (.Values.server.enabled | toString) "-") .Values.server.enabled) (and (eq (.Values.server.enabled | toString) "-") .Values.global.enabled)) (or (and (ne (.Values.ui.enabled | toString) "-") .Values.ui.enabled) (and (eq (.Values.ui.enabled | toString) "-") .Values.global.enabled)) (or (and (ne (.Values.ui.service.enabled | toString) "-") .Values.ui.service.enabled) (and (eq (.Values.ui.service.enabled | toString) "-") .Values.global.enabled))) }}
 {{- if (and (ne (.Values.ui.ingress.enabled | toString) "-") .Values.ui.ingress.enabled) }}
 {{- $serviceName := printf "%s-%s" (include "consul.fullname" .) "ui" -}}
-apiVersion: extensions/v1
+{{- if or ( gt .Capabilities.KubeVersion.Major "1" ) ( ge .Capabilities.KubeVersion.Minor "19" ) }}
```
lkysow, line 4: can you comment why we're not using `Capabilities.APIVersions.Has` using `{{- /* ..docs... */}}`

lkysow (approved):
I haven't tested it myself.

templates/ui-ingress.yaml
thisisnotashwin: Cool. I tested this out on 1.16 and 1.19 and it worked in both cluster. Need an annotation for `nginx` in `1.19` but that was straightforward. Will document the need for the above.

@lkysow Can you run the bats locally? It was failing on my machine but it passes in CI, because of how `helm2`/`helm3` results are different between CI and my local setup.

thisisnotashwin: I ask because im not quite certain how to work around this other than explicitly adding a KubeVersion check in our bats as well which would be a new pattern in the tests.

lkysow (approved):

templates/ui-ingress.yaml
thisisnotashwin: Comment added.

PR #795, opened 2021-01-27
Update pull_request_template.md

ishustava:
No longer require to submit docs PRs on every pull request. It's better if we do it right before a release because we can't merge those PRs until we release anyway.


thisisnotashwin (approved):
masterful!

kschoche (approved):
🍬

PR #799, opened 2021-02-01
Fix nightly kind acceptance tests

ishustava:
* Don't run them in parallel so that we only get one slack notification (we don't care about speed here since all other tests don't run in parallel)
* Add gotestsum summary


ndhanushkodi (approved):

thisisnotashwin (approved):
🥑

PR #800, opened 2021-02-01
acceptance-tests: Increase time we're waiting for pods to be ready to 10 min

ishustava:
On Azure, volumes sometimes could take quite long to provision.
Since we were waiting for 5min before, sometimes it's not enough time
for pods to come up and be healthy.

### DIFF
--- a/test/acceptance/framework/helpers/helpers.go
+++ b/test/acceptance/framework/helpers/helpers.go
@@ -34,8 +34,10 @@ func WaitForAllPodsToBeReady(t *testing.T, client kubernetes.Interface, namespac
 
 	logger.Log(t, "Waiting for pods to be ready.")
 
-	// Wait up to 5m.
-	counter := &retry.Counter{Count: 60, Wait: 5 * time.Second}
+	// Wait up to 15m.
+	// On Azure, volume provisioning can sometimes take close to 5 min,
+	// so we need to give a bit more time for pods to become healthy.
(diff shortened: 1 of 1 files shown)


thisisnotashwin (approved):
🥑

test/acceptance/framework/helpers/helpers.go
```diff
@@ -34,8 +34,10 @@ func WaitForAllPodsToBeReady(t *testing.T, client kubernetes.Interface, namespac
 
 	logger.Log(t, "Waiting for pods to be ready.")
 
-	// Wait up to 5m.
-	counter := &retry.Counter{Count: 60, Wait: 5 * time.Second}
+	// Wait up to 10m.
+	// On Azure, volume provisioning can sometimes take close to 5 min,
+	// so we need to give a bit more time for pods to become healthy.
+	counter := &retry.Counter{Count: 120, Wait: 5 * time.Second}
```
lkysow, line 40: Should we match https://github.com/hashicorp/consul-helm/blob/master/test/acceptance/framework/consul/consul_cluster.go#L85 at 15m? They're both the same timeout, one for when there are acls and so the hook needs to finish before helm install exits, the other (this one) for no acls.

ishustava: Sure that sounds reasonable!

lkysow (approved):

PR #802, opened 2021-02-02
Create templates for grafana and prometheus

thisisnotashwin:
Changes proposed in this PR:
- Add template for prometheus and grafana
- Add script to generate said templates

TO-DO
- Are there other dashboards we would like to support?

How I've tested this PR:
- Followed the guide for [layer 7 observability](https://learn.hashicorp.com/tutorials/consul/kubernetes-layer7-observability?in=consul/day-2-operations) and relied on installed prometheus and grafana for metrics.

- Code review
- Use above guide/BYO apps to test metrics via prometheus and grafana

Checklist:
- [x] Bats tests added

Here is a values file you could use:
```
server:
  replicas: 1
  extraConfig: |
    {
      "ui_config": {
        "enabled": true,
        "metrics_provider": "prometheus",
        "metrics_proxy": {
          "base_url": "http://prometheus-server"
        }
      }
    }
connectInject:
  enabled: true
controller:
  enabled: true
client:
  enabled: true
  grpc: true
ui:
  enabled: true
prometheus:
  enabled: true
grafana:
  enabled: true
```
Also, proxy-default for enabling metrics:
```
kind: ProxyDefaults
metadata:
  name: global
spec:
  config:
    envoy_prometheus_bind_addr: "0.0.0.0:9102"
```

### DIFF
--- a/addons/gen.sh
+++ b/addons/gen.sh
@@ -34,6 +34,14 @@ helm template prometheus prometheus \
   -f "${WD}/values-prometheus.yaml" \
   > "${TEMPLATES}/prometheus.yaml"
 
+sed -i'.orig' 's/replace-me-namespace/{{ .Release.Namespace }}/g' "${TEMPLATES}/prometheus.yaml"
+sed -i'.orig' '1i\
+{{- if .Values.prometheus.enabled }}
+' "${TEMPLATES}/prometheus.yaml"
+sed -i'.orig' -e '$a\
+{{- end }}' "${TEMPLATES}/prometheus.yaml"
+rm "${TEMPLATES}/prometheus.yaml.orig"
+
 function compressDashboard() {
   < "${DASHBOARDS}/$1" jq -c  > "${TMP}/$1"
 }
@@ -54,3 +62,11 @@ function compressDashboard() {
     --from-file=consul-server-monitoring.json="${TMP}/consul-server-monitoring.json"
 
 } > "${TEMPLATES}/grafana.yaml"
+
+sed -i'.orig' 's/replace-me-namespace/{{ .Release.Namespace }}/g' "${TEMPLATES}/grafana.yaml"
+sed -i'.orig' '1i\
+{{- if .Values.grafana.enabled }}
+' "${TEMPLATES}/grafana.yaml"
+sed -i'.orig' -e '$a\
+{{- end }}' "${TEMPLATES}/grafana.yaml"
+rm "${TEMPLATES}/grafana.yaml.orig"
--- a/addons/gen.sh
+++ b/addons/gen.sh
@@ -1,61 +1,48 @@
 #!/usr/bin/env bash
 
-# Copyright consul Authors
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#    http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
 WD=$(dirname "$0")
 WD=$(cd "$WD"; pwd)
 
 set -eux
 
-# This script sets up the plain text rendered deployments for addons
-# See samples/addons/README.md for more information
-
 TEMPLATES="${WD}/../templates"
 DASHBOARDS="${WD}/dashboards"
 TMP=$(mktemp -d)
 
-# Set up prometheus
+# create Prometheus template
 helm template prometheus prometheus \
-  --namespace "replace-me-namespace" \
-  --version 11.16.2 \
   --repo https://prometheus-community.github.io/helm-charts \
-  -f "${WD}/values-prometheus.yaml" \
+  --namespace "replace-me-namespace" \
+  --version 13.2.1 \
+  -f "${WD}/values/prometheus.yaml" \
   > "${TEMPLATES}/prometheus.yaml"
 
+# Find and replace `replace-me-namespace` with `{{ .Release.Namespace }}` in Prometheus template.
 sed -i'.orig' 's/replace-me-namespace/{{ .Release.Namespace }}/g' "${TEMPLATES}/prometheus.yaml"
+# Add a comment to the top of the template file mentioning that the file is auto-generated.
+sed -i'.orig' '1i\
+# This file is auto-generated, see addons/gen.sh
+' "${TEMPLATES}/prometheus.yaml"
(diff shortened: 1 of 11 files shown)


addons/gen.sh
```diff
@@ -0,0 +1,58 @@
...
+  -f "${WD}/values-prometheus.yaml" \
+  > "${TEMPLATES}/prometheus.yaml"
+
+sed -i'.orig' 's/replace-me-namespace/{{ .Release.Namespace }}/g' "${TEMPLATES}/prometheus.yaml"
+sed -i'.orig' '1i\
+{{- if .Values.prometheus.enabled }}
+' "${TEMPLATES}/prometheus.yaml"
+sed -i'.orig' -e '$a\
+{{- end }}' "${TEMPLATES}/prometheus.yaml"
+rm "${TEMPLATES}/prometheus.yaml.orig"
```
thisisnotashwin, line 29: I would love if there was a simpler way to 
• Replace `replace-me-namespace` with `{{ .Release.Namespace }}`
• Top and tail the file with `{{- if .Values.prometheus.enabled }}` and `{{- end }}`

addons/gen.sh
```diff
@@ -0,0 +1,58 @@
...
+    --from-file=consul-server-monitoring.json="${TMP}/consul-server-monitoring.json"
+
+} > "${TEMPLATES}/grafana.yaml"
+
+sed -i'.orig' 's/replace-me-namespace/{{ .Release.Namespace }}/g' "${TEMPLATES}/grafana.yaml"
+sed -i'.orig' '1i\
+{{- if .Values.grafana.enabled }}
+' "${TEMPLATES}/grafana.yaml"
+sed -i'.orig' -e '$a\
+{{- end }}' "${TEMPLATES}/grafana.yaml"
```
thisisnotashwin, line 57: I would love if there was a simpler way to 
• Replace `replace-me-namespace` with `{{ .Release.Namespace }}`
• Top and tail the file with `{{- if .Values.grafana.enabled }}` and `{{- end }}`

addons/gen.sh
```diff
@@ -0,0 +1,58 @@
...
+# Set up grafana
+{
+  helm template grafana grafana \
+    --namespace "replace-me-namespace" \
+    --version 6.2.1 \
+    --repo https://grafana.github.io/helm-charts \
+    -f "${WD}/values-grafana.yaml"
+
+  # Set up grafana dashboards. Compress to single line json to avoid Kubernetes size limits
+  compressDashboard "consul-server-monitoring.json"
```
thisisnotashwin, line 44: this is the only default dashboard I found. are there we would like to ship by default? also, this requires the stats port to be configured on Consul which i don't think is the default, so might need tinkering.

thisisnotashwin: assuming this is for demos, it might be good to keep an eye out, but can be punted to later. @codyde and @DerekStrickland, do y'all have frequently displayed dashboards for grafana?

DerekStrickland: I don't, but @mkcp created [this one](https://grafana.com/grafana/dashboards/13396) that needs these changes in order to work.

thisisnotashwin: Yup! That dashboard is the only one included so far 😄

ndhanushkodi (approved):
It's looking great to me! One thing I found helpful when installing prometheus via it's helm chart was that it would print out helpful lines to paste to port forward to the UI.
```
export POD_NAME=$(kubectl get pods --namespace default -l "app=prometheus,component=server" -o jsonpath="{.items[0].metadata.name}")
kubectl --namespace default port-forward $POD_NAME 9090
```
Maybe our helm chart could do something similar by printing these lines for prometheus and grafana if they are enabled? Definitely not a requirement though, and we can always just add it to documentation instead or expect folks to look at prometheus/grafana docs instead.

addons/values-prometheus.yaml
```diff
@@ -0,0 +1,24 @@
...
+kubeStateMetrics:
+  enabled: false
+nodeExporter:
+  enabled: false
+server:
+  podAnnotations:
+    "consul.hashicorp.com/connect-inject": "false"
+  persistentVolume:
+    enabled: false
+  # Use port 9090 to match Consul documentation (TBD)
```
lkysow: TBD?

addons/gen.sh
lkysow: Yeah can you add comments about what each search/replace is doing

addons/gen.sh
```diff
@@ -0,0 +1,58 @@
...
+
+sed -i'.orig' 's/replace-me-namespace/{{ .Release.Namespace }}/g' "${TEMPLATES}/prometheus.yaml"
+sed -i'.orig' '1i\
+{{- if .Values.prometheus.enabled }}
+' "${TEMPLATES}/prometheus.yaml"
+sed -i'.orig' -e '$a\
+{{- end }}' "${TEMPLATES}/prometheus.yaml"
+rm "${TEMPLATES}/prometheus.yaml.orig"
+
+function compressDashboard() {
```
lkysow: can you add a comment for what this does

addons/gen.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# create Prometheus template
+helm template prometheus prometheus \
+  --repo https://prometheus-community.github.io/helm-charts \
+  --namespace "replace-me-namespace" \
+  --version 13.2.1 \
+  -f "${WD}/values/prometheus.yaml" \
+  > "${TEMPLATES}/prometheus.yaml"
+
+sed -i'.orig' 's/replace-me-namespace/{{ .Release.Namespace }}/g' "${TEMPLATES}/prometheus.yaml"
```
lkysow, line 20: Can we also add a `this file is auto-generated, see addons/gen.sh` to the top of each generated file

addons/values-grafana.yaml
```diff
@@ -0,0 +1,60 @@
...
+ldap:
+  existingSecret: true
+env:
+  GF_SECURITY_ADMIN_USER: "-"
+  GF_SECURITY_ADMIN_PASSWORD: "-"
+  GF_AUTH_BASIC_ENABLED: "false"
+  GF_AUTH_ANONYMOUS_ENABLED: "true"
+  GF_AUTH_ANONYMOUS_ORG_ROLE: Admin
+
+# Expose on port 3000 to match the Consul docs (TBD)
```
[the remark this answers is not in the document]
thisisnotashwin: We are using the default ports everywhere ie 3000 for Grafana and 9090 for Prometheus. Removed that line.

thisisnotashwin:
> Maybe our helm chart could do something similar by printing these lines for prometheus and grafana if they are enabled?

That makes sense. I only erred on the side of not providing this information given that hasnt appeared to be the trend with our existing helm installation (port-forwarding UI eg). i dont have a strong opinion one way or another here.
@lkysow thoughts?

lkysow:
I think that in the end because we're exposing the service metrics through our UI that we don't really expect folks to navigate to grafana in most cases?

lkysow (approved):

PR #805, opened 2021-02-03
Add pod IP to IP SANS of client certs

lkysow:
* So cert is valid if calling client via pod IP rather than host IP.
* Also add the host IP as a meta value to the client agent since this
information is not available through Consul right now.

How I've tested this PR:
* As part of pod deregistration. I got cert errors and then using this branch I did not.

* Can test manually by bringing up consul with/without autoencrypt and using openssl to view the certs or you can trust the config

Checklist:
- [x] Bats tests added


thisisnotashwin (approved):
Trusting the config 🟢

ndhanushkodi (approved):
Looks great! Were folks running into this (was there a GH issue to maybe resolve)?

lkysow:
> Looks great! Were folks running into this (was there a GH issue to maybe resolve)?

No one ran into it but me :D

PR #806, opened 2021-02-04
Cleanup controller

lkysow:
Support cleanup controller that cleans up Consul service instances from
deleted pods.

Part of https://github.com/hashicorp/consul-k8s/pull/433

How I've tested this PR: Acceptance tests

* Force delete pod, see that the service instance is removed
* To test reconcile, can register a service via API directly and ensure it's removed

Checklist:
- [x] Bats tests added

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -155,7 +155,7 @@ jobs:
                     -kubecontext="kind-dc1" \
                     -secondary-kubecontext="kind-dc2" \
                     -debug-directory="$TEST_RESULTS/debug" \
-                    -consul-k8s-image=docker.mirror.hashicorp.services/hashicorpdev/consul-k8s:latest
+                    -consul-k8s-image=ghcr.io/lkysow/consul-k8s-dev:feb09
               then
                 echo "Tests in ${pkg} failed, aborting early"
                 exit_code=1
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -155,7 +155,7 @@ jobs:
                     -kubecontext="kind-dc1" \
                     -secondary-kubecontext="kind-dc2" \
                     -debug-directory="$TEST_RESULTS/debug" \
-                    -consul-k8s-image=ghcr.io/lkysow/consul-k8s-dev:feb09
+                    -consul-k8s-image=docker.mirror.hashicorp.services/hashicorpdev/consul-k8s:latest
               then
                 echo "Tests in ${pkg} failed, aborting early"
                 exit_code=1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ IMPROVEMENTS:
 * CRDs: add value `controller.aclToken` to support manually passing in an ACL token to the CRD controller if independently managing ACLs. [[GH-783](https://github.com/hashicorp/consul-helm/pull/783)]
 * TLS: Consul client certificates now include their pod IPs in the IP SANs. This applies to auto-encrypt enabled and disabled. [[GH-805](https://github.com/hashicorp/consul-helm/pull/805)]
 * Consul client nodes have a new meta key called "host-ip" set to the IP of the Kubernetes node they're running on. [[GH-805](https://github.com/hashicorp/consul-helm/pull/805)]
-* Connect: the latest version of consul-k8s cleans up Consul service instances whose pods are no longer running.
+* Connect: the latest version of consul-k8s cleans up Consul connect service mesh instances whose pods are no longer running.
   This could happen if the pod's `preStop` hook failed to execute for some reason. [[GH-806](https://github.com/hashicorp/consul-helm/pull/806)]
 
 BUG FIXES:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ IMPROVEMENTS:
 * CRDs: add value `controller.aclToken` to support manually passing in an ACL token to the CRD controller if independently managing ACLs. [[GH-783](https://github.com/hashicorp/consul-helm/pull/783)]
 * TLS: Consul client certificates now include their pod IPs in the IP SANs. This applies to auto-encrypt enabled and disabled. [[GH-805](https://github.com/hashicorp/consul-helm/pull/805)]
 * Consul client nodes have a new meta key called "host-ip" set to the IP of the Kubernetes node they're running on. [[GH-805](https://github.com/hashicorp/consul-helm/pull/805)]
+* Connect: the latest version of consul-k8s cleans up Consul service instances whose pods are no longer running.
+  This could happen if the pod's `preStop` hook failed to execute for some reason. [[GH-806](https://github.com/hashicorp/consul-helm/pull/806)]
 
 BUG FIXES:
 * Use `rbac.authorization.k8s.io/v1` instead of `rbac.authorization.k8s.io/v1beta1` API version for the `roles` and `rolebindings` used by the `tls-init`
--- a/templates/connect-inject-clusterrole.yaml
+++ b/templates/connect-inject-clusterrole.yaml
@@ -17,14 +17,12 @@ rules:
     - "list"
     - "watch"
     - "patch"
-{{- if .Values.connectInject.healthChecks.enabled }}
 - apiGroups: [""]
-  resources: ["pods"]
+  resources: ["pods", "nodes"]
   verbs:
     - "get"
     - "list"
     - "watch"
-{{- end}}
 {{- if .Values.global.enablePodSecurityPolicies }}
 - apiGroups: ["policy"]
   resources: ["podsecuritypolicies"]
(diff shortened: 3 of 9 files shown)


templates/connect-inject-clusterrole.yaml
```diff
@@ -17,14 +17,12 @@ rules:
     - "list"
     - "watch"
     - "patch"
-{{- if .Values.connectInject.healthChecks.enabled }}
 - apiGroups: [""]
-  resources: ["pods"]
+  resources: ["pods", "nodes"]
```
lkysow, line 21: Needs nodes because we look up the list of nodes in the controller

templates/connect-inject-clusterrole.yaml
```diff
@@ -33,7 +31,7 @@ rules:
   verbs:
   - use
 {{- end }}
-{{- if and .Values.global.acls.manageSystemACLs (or .Values.global.enableConsulNamespaces .Values.connectInject.healthChecks.enabled) }}
```
lkysow, line 36: Previously we'd only need an acl token if namespaces or healthchecks were enabled but now we always need one because cleanup controller is on by default (and no way right now to turn it off through helm)

test/unit/connect-inject-clusterrole.bats
```diff
@@ -74,110 +74,15 @@ load _helpers
 }
```
lkysow, line 75: Needing to talk to a local consul client used to be conditional on namespaces or health checks being enabled but now it's always required since the cleanup controller needs it. Since we don't write tests against fields that are always existing in the template it lets us delete a lot of these tests.

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ IMPROVEMENTS:
 * CRDs: add value `controller.aclToken` to support manually passing in an ACL token to the CRD controller if independently managing ACLs. [[GH-783](https://github.com/hashicorp/consul-helm/pull/783)]
 * TLS: Consul client certificates now include their pod IPs in the IP SANs. This applies to auto-encrypt enabled and disabled. [[GH-805](https://github.com/hashicorp/consul-helm/pull/805)]
 * Consul client nodes have a new meta key called "host-ip" set to the IP of the Kubernetes node they're running on. [[GH-805](https://github.com/hashicorp/consul-helm/pull/805)]
+* Connect: the latest version of consul-k8s cleans up Consul service instances whose pods are no longer running.
```
kschoche, line 31: meta comment:  I know that we start the line with  `"Connect:"` but do you think this is sufficiently descriptive to a random reader that this is ONLY for connect injected pods, not sync-catalog?

lkysow: No probably not, will update to make it clear. Something like `cleans up Consul connect service instances...`?

test/unit/connect-inject-clusterrole.bats
```diff
@@ -74,110 +74,15 @@ load _helpers
...
 
 #--------------------------------------------------------------------
-# global.acls.manageSystemACLs for namespaces
```
kschoche, line 77: Does it make sense to add a test to cover the new clusterrole rules, e.g. the addition of the `nodes` resource?
 `resources: ["pods", "nodes"]`

kschoche (approved):
This looks great, and I love that we can clean up a bunch of 🦇  tests :)
Ive just left a couple comments.
One additional question: Do we have an idea of how this works with sync-catalog turned on?
Does it make sense to add some test which has connect+sync-catalog enabled to ensure we don't have racing behaviour (if it can exist?)

test/unit/connect-inject-clusterrole.bats
lkysow: Typically we don't write tests for non-conditionals, e.g. we would write a test for:
```
- resources: ["pods"{{ if .something}}, "nodes"{{end}}]
```

to check that nodes isn't added unless `.something`.

But we wouldn't write a test for:
```
- resources: ["pods", "nodes"]
```

Because at that point you're testing that the template is as you wrote it, which would lead to having to test every line in the template which is too much testing.

For example here we have all these resources: https://github.com/hashicorp/consul-helm/blob/master/templates/controller-clusterrole.yaml#L12-L23 but because they're not templated in a conditional, we don't test that they're set in https://github.com/hashicorp/consul-helm/blob/master/test/unit/controller-clusterrole.bats.

kschoche: cool, makes sense! no worries then :)

lkysow:
Good questions.

> One additional question: Do we have an idea of how this works with sync-catalog turned on?

We gate deletion right now on:
1. instance has `pod-name` and `k8s-namespace` metadata
2. instance is registered on a node name that exists in the kube cluster

Catalog synced instances don't have the `pod-name` or `k8s-namespace` metadata and are registered to a node called `k8s-sync` which won't match a kube node name so they will be ignored and the cleanup controller won't deregister them.

> Does it make sense to add some test which has connect+sync-catalog enabled to ensure we don't have racing behaviour (if it can exist?)

On the one hand the best place to test this is in an acceptance test because it's the most real world. On the other, our acceptance tests explicitly only test happy-path scenarios because we didn't want them to get too large and complicated. Would a good middle-path here be to turn on `connectInject` alongside the catalog sync tests (so the cleanup controller will be running) but not write an explicit test?

CHANGELOG.md
lkysow: just unresolving so I don't forget.

kschoche:
> On the one hand the best place to test this is in an acceptance test because it's the most real world. On the other, our acceptance tests explicitly only test happy-path scenarios because we didn't want them to get too large and complicated. Would a good middle-path here be to turn on `connectInject` alongside the catalog sync tests (so the cleanup controller will be running) but not write an explicit test?

I think based on your awesome explanation I'm okay with not writing any type of explicit test and we probably don't need to add sync-catalog to any of the tests to prove it out. So from my standpoint it looks good to go!

lkysow:
> > On the one hand the best place to test this is in an acceptance test because it's the most real world. On the other, our acceptance tests explicitly only test happy-path scenarios because we didn't want them to get too large and complicated. Would a good middle-path here be to turn on `connectInject` alongside the catalog sync tests (so the cleanup controller will be running) but not write an explicit test?
> 
> I think based on your awesome explanation I'm okay with not writing any type of explicit test and we probably don't need to add sync-catalog to any of the tests to prove it out. So from my standpoint it looks good to go!

👍 let's see what @ndhanushkodi thinks too.

test/acceptance/tests/connect/connect_inject_test.go
```diff
@@ -88,3 +92,88 @@ func TestConnectInject(t *testing.T) {
...
+
+			// There should only be one pod because we set replicas to 1 but
+			// in some cases, the previous test's static-client pod is still
+			// terminating so we filter by ready pods.
+			var podName string
+			for _, pod := range pods.Items {
+				if helpers.IsReady(pod) {
+					podName = pod.Name
+					break
+				}
```
lkysow, line 153: This can be removed once https://github.com/hashicorp/consul-helm/pull/818 is merged.

.circleci/config.yml
```diff
@@ -155,7 +155,7 @@ jobs:
                     -kubecontext="kind-dc1" \
                     -secondary-kubecontext="kind-dc2" \
                     -debug-directory="$TEST_RESULTS/debug" \
-                    -consul-k8s-image=docker.mirror.hashicorp.services/hashicorpdev/consul-k8s:latest
+                    -consul-k8s-image=ghcr.io/lkysow/consul-k8s-dev:feb09
```
ndhanushkodi, line 158: I wasn't able to test using your ghcr image or the one created by consul-k8s-ci because I was seeing this error in both cases:
```
k logs consul-consul-connect-injector-webhook-deployment-7f69bc4dftckh
flag provided but not defined: -cleanup-controller-reconcile-period
```
I'll code review the rest for now.

test/unit/connect-inject-clusterrole.bats
ndhanushkodi: Nice!

.circleci/config.yml
ndhanushkodi, line 158: I wasn't able to test using your ghcr image or the one created by consul-k8s-ci because I was seeing this error in both cases:
```
k logs consul-consul-connect-injector-webhook-deployment-7f69bc4dftckh
flag provided but not defined: -cleanup-controller-reconcile-period
```
I'll code review the rest for now.

ndhanushkodi (approved):
Didn't end up having any other comments in the code itself, looks great!

For the acceptance tests, 
>Would a good middle-path here be to turn on connectInject alongside the catalog sync tests (so the cleanup controller will be running) but not write an explicit test?

Would the pods that are being catalog sync'd also end up having a sidecar injected into them, making it so when pods are force deleted, they would be removed from Consul? If so it seems like a weird middle ground that we might want to see in our tests. If it would just continue with the catalog sync behavior with the cleanup controller running in the background, it seems fine to leave that out.

lkysow:
> Would the pods that are being catalog sync'd also end up having a sidecar injected into them, making it so when pods are force deleted, they would be removed from Consul? If so it seems like a weird middle ground that we might want to see in our tests. If it would just continue with the catalog sync behavior with the cleanup controller running in the background, it seems fine to leave that out.

They catalog sync'd pods do not have sidecars.

PR #810, opened 2021-02-08
Rename lifecycle-sidecar to consul-sidecar

ndhanushkodi:
Changes proposed in this PR:
- deprecates helm value lifecycleSidecarContainer and replaces it with
consulSidecarContainer. 
- replaces all uses of lifecycle sidecar
with consul sidecar, including container names, commands, and flags.

This is to enable future functionality within the consul-sidecar
container, since it may no longer be only responsible for managing
lifecycle.

How I've tested this PR:
acceptance tests against consul-k8s rename-sidecar branch

code review

Checklist:
- [x] Bats tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,12 @@ BREAKING CHANGES:
 
   See [Upgrade to CRDs](https://www.consul.io/docs/k8s/crds/upgrade-to-crds)
   for more information on how to upgrade.
+* The lifecycle-sidecar command and container has been renamed to
+  consul-sidecar. The Helm value `global.lifecycleSidecarContainer` has been
+  renamed to `global.consulSidecarContainer`.
+  `global.lifecycleSidecarContainer` is no longer supported and will cause
+  errors on `helm upgrade`. Please use `global.consulSidecarContainer` instead.
+  [[GH-810](https://github.com/hashicorp/consul-helm/pull/810)]
 
 IMPROVEMENTS:
 * Add ability to set extra labels on Consul client pods. [[GH-612](https://github.com/hashicorp/consul-helm/pull/612)]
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -155,7 +155,7 @@ jobs:
                     -kubecontext="kind-dc1" \
                     -secondary-kubecontext="kind-dc2" \
                     -debug-directory="$TEST_RESULTS/debug" \
-                    -consul-k8s-image=docker.mirror.hashicorp.services/hashicorpdev/consul-k8s:latest
+                    -consul-k8s-image=docker.mirror.hashicorp.services/hashicorpdev/consul-k8s@sha256:d158c0b4956421889e7ba80d82c24baa365c5cd3148dac62b555b4c9ba800c0a
               then
```
ndhanushkodi, line 159: Will be removed before merging. This is to use the docker image built for [the corresponding consul-k8s change](https://github.com/hashicorp/consul-k8s/pull/428) when running acceptance tests.

.circleci/config.yml
```diff
@@ -155,7 +155,7 @@ jobs:
                     -kubecontext="kind-dc1" \
```
ndhanushkodi, line 155: Does it make sense to change references to lifecycle-sidecar in hack/helm-reference-gen to also use consul-sidecar? I figure the exact values are getting stale anyway and meant to test that nested nodes work, etc.

thisisnotashwin: Yup!! That sounds right. It would be nice if the fixture looked even lesser like our values file because it definitely is confusing. I don't think it needs to be updated and can be left as is.

thisisnotashwin (approved):
This looks great!! Excellent work on the pair of PRs @ndhanushkodi

ishustava (approved):
Looks great!

PR #818, opened 2021-02-09
Acceptance: wait for previous installation to exit

lkysow:
In some cases the previous installation is listed as uninstalled by Helm
but its pods are still terminating. This can cause issues for the newly
running tests when they query Kubernetes to get information about the
new installation, e.g. getting the server pod IPs might return the IPs
from the old installation as well.

This was the cause of the failure of https://app.circleci.com/pipelines/github/hashicorp/consul-helm/2413/workflows/1fb2bed2-96cd-4779-90c3-fbe5c3f84c7a/jobs/8057/steps

I logged which IPs it was getting and you can see that it was getting two IPs, one of which was from a terminating pod.
```
    consul_dns_test.go:44: 2021-02-09T17:58:19Z Waiting for pods to be ready.
    consul_dns_test.go:61: is ready: true, ip=10.244.0.90
    consul_dns_test.go:61: is ready: false, ip=10.244.0.83

```

### DIFF
--- a/test/acceptance/framework/consul/consul_cluster.go
+++ b/test/acceptance/framework/consul/consul_cluster.go
@@ -287,6 +287,21 @@ func (h *HelmCluster) checkForPriorInstallations(t *testing.T) {
 	for _, r := range installedReleases {
 		require.NotContains(t, r["chart"], "consul", fmt.Sprintf("detected an existing installation of Consul %s, release name: %s", r["chart"], r["name"]))
 	}
+
+	// Wait for all pods in the "default" namespace to exit. A previous
+	// release may not be listed by Helm but its pods may still be terminating.
+	retry.RunWith(&retry.Counter{Wait: 1 * time.Second, Count: 30}, t, func(r *retry.R) {
+		consulPods, err := h.kubernetesClient.CoreV1().Pods(h.helmOptions.KubectlOptions.Namespace).List(context.Background(), metav1.ListOptions{})
+		require.NoError(r, err)
+		if len(consulPods.Items) > 0 {
+			var podNames []string
+			for _, p := range consulPods.Items {
+				podNames = append(podNames, p.Name)
+			}
+			r.Errorf("pods from previous installation still running: %s", strings.Join(podNames, ", "))
+		}
+	})
+
 }
 
 // configurePodSecurityPolicies creates a simple pod security policy, a cluster role to allow access to the PSP,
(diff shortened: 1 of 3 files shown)


test/acceptance/framework/consul/consul_cluster.go
```diff
@@ -287,6 +287,19 @@ func (h *HelmCluster) checkForPriorInstallations(t *testing.T) {
...
+	// Wait for all pods labelled with app=consul to exit. A previous
+	// release may not be listed by Helm but its pods may still be terminating.
+	retry.RunWith(&retry.Counter{Wait: 1 * time.Second, Count: 30}, t, func(r *retry.R) {
+		consulPods, err := h.kubernetesClient.CoreV1().Pods(h.helmOptions.KubectlOptions.Namespace).List(context.Background(), metav1.ListOptions{LabelSelector: "app=consul"})
+		require.NoError(r, err)
+		var podNames []string
+		for _, p := range consulPods.Items {
+			podNames = append(podNames, p.Name)
+		}
+		require.Len(r, podNames, 0)
```
thisisnotashwin, line 300: Any reason we can't do a `require.Len(r, consulPods.Items, 0)`?

thisisnotashwin (commented):
Looks good. Have a small comment.

test/acceptance/framework/consul/consul_cluster.go
lkysow: Oops, I thought `require.Len` was the one that prints out the contents of each slice but that's actually `require.ElementsMatch` I think. Basically I want it to print out the pods that are still there so I think this should be

```go
		consulPods, err := h.kubernetesClient.CoreV1().Pods(h.helmOptions.KubectlOptions.Namespace).List(context.Background(), metav1.ListOptions{LabelSelector: "app=consul"})
		require.NoError(r, err)
		if len(consulPods.Items) > 0 {
			var podNames []string
			for _, p := range consulPods.Items {
				podNames = append(podNames, p.Name)
			}
			r.Errorf("pods from previous installation still running: %s", strings.Join(podNames, ", "))
		}

```

thisisnotashwin: heh. makes sense! you're right about `ElementsMatch` thought i suspect that would be `ElemsMatch(r, podNames, [])` but the rationale makes sense!!

thisisnotashwin (approved):
🦊

ishustava (approved):

test/acceptance/framework/consul/consul_cluster.go
lkysow: Right but in that case the error will print out the go-formatted entire pod object so this way it only prints out the pod names.

thisisnotashwin: you have a beautiful 🧠 @lkysow

PR #822, opened 2021-02-10
Add metrics config for Consul UI

thisisnotashwin:
Changes proposed in this PR:
- Support metrics on Consul UI without the need for extraConfig

How I've tested this PR:
Use this values file:
```
server:
  replicas: 1
connectInject:
  enabled: true
controller:
  enabled: true
client:
  enabled: true
  grpc: true
ui:
  enabled: true
prometheus:
  enabled: true
grafana:
  enabled: true
global:
  imageK8S: ashwinvenkatesh/consul-k8s@sha256:dddeb63f220e9b0db794809f8fab2472434af4fca224246dab50617795af5f75
  metrics:
    enabled: true
    enableAgentMetrics: true
    enableGatewayMetrics: true
 ```
As `global.metrics.enabled` is true and `ui.enabled` is `true`, the server is configured with the following extra config:
```
      "ui_config": {
        "enabled": true,
        "metrics_provider": "{{ .Values.ui.metrics.provider }}",
        "metrics_proxy": {
          "base_url": "{{ .Values.ui.metrics.baseURL }}"
        }
      }
```

Deployed apps into the mesh and adding traffic shows app metrics on the consul UI.

Same as above

Checklist:
- [x] Bats tests added

### CODE REVIEW

File: test/unit/server-config-configmap.bats [hunk 1]
```diff
@@ -61,6 +61,66 @@ load _helpers
...
+  local actual=$(helm template \
+      -s templates/server-config-configmap.yaml  \
+      --set 'ui.metrics.enabled=true' \
+      --set 'ui.enabled=true' \
+      . | tee /dev/stderr |
+      yq '.data["ui-config.json"] | length > 0' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
+@test "server/ConfigMap: creates ui config with .ui.metrics.enabled disabled & ui enabled" {
```
kschoche, line 89:
```suggestion
@test "server/ConfigMap: does not create ui config with .ui.metrics.enabled disabled & ui enabled" {
```
I think I'm reading this correctly?

thisisnotashwin: haha..indeed you are!! good catch bud!

kschoche: Also a super super tiny nit you're using `ui.metrics.enabled enabled` and also `ui enabled`  can we just swap this all to `ui.metrics enabled and ui enabled` ?  the double enabled is not my fav

test/unit/server-config-configmap.bats
```diff
@@ -61,6 +61,66 @@ load _helpers
...
+  local actual=$(helm template \
+      -s templates/server-config-configmap.yaml  \
+      --set 'ui.metrics.enabled=true' \
+      --set 'ui.enabled=true' \
+      . | tee /dev/stderr |
+      yq '.data["ui-config.json"] | length > 0' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
+@test "server/ConfigMap: does not create ui config with .ui.metrics.enabled disabled & ui enabled" {
```
kschoche, line 89: one other super-super-super-minor nit, you're mixing `ui enabled` and `ui.metrics.enabled enabled` and the double-enabled is a bit funky to read.
Can we standardize on something like `ui.metrics enabled and ui enabled` ?

values.yaml
```diff
@@ -1022,6 +1022,21 @@ ui:
     # @type: string
     annotations: null
 
+  # Configurations for displaying metrics in the UI.
+  metrics:
+    # Enable displaying metrics in the UI. The default value of "-"
+    # will inherit from global.metrics.enabled value. This value is only
+    # used if ui.enabled is set to true.
```
kschoche, line 1029:
```suggestion
    # will inherit from global.metrics.enabled value.
```
While this may be true is it redundant?

kschoche, line 1029:
```suggestion
    # will inherit from global.metrics.enabled value.
```

Since this is inside the `ui:` stanza already I think this is implied for all of these.
If we feel its necessary to mention, maybe at the `ui.metrics` level?

values.yaml
```diff
@@ -233,21 +233,25 @@ global:
...
     # servers. This must be greater than 0 for Consul clients and servers
     # to expose any metrics at all.
     # Only applicable if global.metrics.enabled is true.
+    # @type: string
     agentMetricsRetentionTime: 1m
 
     # If true, mesh, terminating, and ingress gateways will expose their
     # Envoy metrics on port 20200 at the /metrics path and all gateway pods
     # will have Prometheus scrape annotations. Only applicable if global.metrics.enabled is true.
+    # @type: boolean
```
kschoche, line 254: 🏆

kschoche (approved):
Looks fantsatic!

values.yaml
```diff
@@ -1022,6 +1026,24 @@ ui:
...
+    # Provider for metrics. See
+    # https://www.consul.io/docs/agent/options#ui_config_metrics_provider
+    # This value is only used if ui.enabled is set to true.
+    # @type: string
+    provider: "prometheus"
+
+    # baseURL is the URL of the prometheus server, usually the service URL.
+    # This value is only used if ui.enabled is set to true.
+    # @type: string
+    baseURL: http://prometheus-server
```
ndhanushkodi, line 1045: Should this be http://prometheus? The service name looks like it's just prometheus.

ndhanushkodi (approved):
I wasn't able to see the UI metrics when running all these steps. My app only had one metric (version) though, so perhaps it only shows specific types of metrics? What services were you able to deploy and see metrics for?

Outside of that looks great!! I was just hoping to run through the steps for extra validation :)

values.yaml
ishustava: Lurking on your PRs here: I think it'd be nice to keep it as prometheus-server so that it just works™️  with the prometheus community helm chart default values (https://github.com/prometheus-community/helm-charts/blob/c79e46a25fc4cff86deb9958be5b23c7a516ab2d/charts/prometheus/values.yaml#L602). If it doesn't make sense to make our templates do the same, then don't worry about it, but it'd be nice to have.

thisisnotashwin: DONE 😄

ishustava (approved):
I haven't looked in detail, but thank you so much for changing the URL to prometheus-server!

PR #834, opened 2021-02-19
Connect Service Metrics

ndhanushkodi:
Changes proposed in this PR:
- Adds connect service metrics flags to configure metrics merging and Prometheus annotations on Connect-injected Pods.

How I've tested this PR:
1. Deploy with the following values: 
```
global:
  domain: consul
  datacenter: dc1
  imageK8S: gcr.io/nitya-293720/consul-k8s-dev:metrics30
  image: gcr.io/nitya-293720/consul-dev:metrics10

server:
  replicas: 1
  bootstrapExpect: 1

client:
  enabled: true
  grpc: true

ui:
  enabled: true
prometheus:
  enabled: true

connectInject:
  enabled: true
  metrics:
    defaultEnabled: true
    defaultEnableMerging: true
    defaultMergedMetricsPort: 12345
    defaultPrometheusScrapePort: 22345
    defaultPrometheusScrapePath: "/metrics"

controller:
  enabled: true
```
2. Apply this app
```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/name: prometheus-example-app
  name: prometheus-example-app
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: prometheus-example-app
  template:
    metadata:
      annotations:
        'consul.hashicorp.com/connect-inject': 'true'
        'consul.hashicorp.com/connect-service-upstreams': 'static-server:1234'
      labels:
        app.kubernetes.io/name: prometheus-example-app
    spec:
      containers:
      - name: prometheus-example-app
        image: quay.io/brancz/prometheus-example-app:v0.3.0
        ports:
        - name: web
          containerPort: 8080
```
3. Port forward to prometheus server
4. Navigate to prometheus dashboard "graph" tab and try to graph the "version" metric

Follow steps above, code review

Checklist:
- [x] Bats tests added

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -144,7 +144,7 @@ jobs:
                     -kubecontext="kind-dc1" \
                     -secondary-kubecontext="kind-dc2" \
                     -debug-directory="$TEST_RESULTS/debug" \
-                    -consul-k8s-image=docker.mirror.hashicorp.services/hashicorpdev/consul-k8s:latest
+                    -consul-k8s-image=docker.mirror.hashicorp.services/hashicorpdev/consul-k8s:4adf933
               then
                 echo "Tests in ${pkg} failed, aborting early"
                 exit_code=1
--- a/templates/connect-inject-deployment.yaml
+++ b/templates/connect-inject-deployment.yaml
@@ -94,6 +94,15 @@ spec:
                 -health-checks-reconcile-period={{ .Values.connectInject.healthChecks.reconcilePeriod }} \
                 {{- end }}
                 -cleanup-controller-reconcile-period={{ .Values.connectInject.cleanupController.reconcilePeriod }} \
+                {{- if (or (and (ne (.Values.connectInject.metrics.defaultEnabled | toString) "-") .Values.connectInject.metrics.defaultEnabled) (and (eq (.Values.connectInject.metrics.defaultEnabled | toString) "-") .Values.global.metrics.enabled)) }}
+                -default-enable-metrics=true \
+                {{- else }}
+                -default-enable-metrics=false \
+                {{- end }}
+                -default-enable-metrics-merging={{ .Values.connectInject.metrics.defaultEnableMerging }}  \
+                -default-merged-metrics-port={{ .Values.connectInject.metrics.defaultMergedMetricsPort }} \
+                -default-prometheus-scrape-port={{ .Values.connectInject.metrics.defaultPrometheusScrapePort }} \
+                -default-prometheus-scrape-path="{{ .Values.connectInject.metrics.defaultPrometheusScrapePath }}" \
                 {{- if .Values.connectInject.envoyExtraArgs }}
                 -envoy-extra-args="{{ .Values.connectInject.envoyExtraArgs }}" \
                 {{- end }}
--- a/values.yaml
+++ b/values.yaml
@@ -1271,6 +1271,43 @@ connectInject:
     # reconcile is done after the initial reconcile at startup is completed.
     reconcilePeriod: "1m"
 
+  # Configures metrics for Consul Connect services. All values are overridable
+  # via annotations on a per-pod basis.
+  metrics:
+    # If true, the connect-injector will automatically
+    # add prometheus annotations to connect-injected pods. It will also
+    # add a listener on the Envoy sidecar to expose metrics. The exposed
+    # metrics will depend on whether metrics merging is enabled:
+    #   - If metrics merging is enabled:
+    #     the Consul sidecar will run a merged metrics server
+    #     combining Envoy sidecar and Connect service metrics,
+    #     i.e. if your service exposes its own Prometheus metrics.
+    #   - If metrics merging is disabled:
+    #     the listener will just expose Envoy sidecar metrics.
+    # This will inherit from `global.metrics.enabled`.
+    defaultEnabled: "-"
+    # Configures the Consul sidecar to run a merged metrics server
+    # to combine and serve both Envoy and Connect service metrics.
+    defaultEnableMerging: false
+    # Configures the port at which the Consul sidecar will listen on to return
+    # combined metrics. This port only needs to be changed if it conflicts with
+    # the application's ports.
+    defaultMergedMetricsPort: 20100
+    # Configures the port Prometheus will scrape metrics from, by configuring
+    # the Pod annotation `prometheus.io/port` and the corresponding listener in
+    # the Envoy sidecar.
+    # NOTE: This is *not* the port that your application exposes metrics on.
+    # That can be configured with the
+    # `consul.hashicorp.com/service-metrics-port` annotation.
+    defaultPrometheusScrapePort: 20200
+    # Configures the path Prometheus will scrape metrics from, by configuring the pod
+    # annotation prometheus.io/path and the corresponding handler in the Envoy
+    # sidecar.
+    # NOTE: This is *not* the path that your application exposes metrics on.
+    # That can be configured with the
+    # `consul.hashicorp.com/service-metrics-path` annotation.
+    defaultPrometheusScrapePath: "/metrics"
+
   # Cleanup controller cleans up Consul service instances that remain registered
   # despite their pods no longer running. This could happen if the pod's `preStop`
   # hook failed to execute for some reason.
--- a/test/unit/connect-inject-deployment.bats
+++ b/test/unit/connect-inject-deployment.bats
@@ -244,6 +244,148 @@ EOF
 }
 
 #--------------------------------------------------------------------
+# metrics
+
+@test "connectInject/Deployment: default connect-inject metrics flags" {
+  cd `chart_dir`
+  local cmd=$(helm template \
+      -s templates/connect-inject-deployment.yaml \
+      --set 'connectInject.enabled=true' \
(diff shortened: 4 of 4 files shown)


thisisnotashwin (approved):
EXCELLENT!!

values.yaml
```diff
@@ -1271,6 +1271,41 @@ connectInject:
...
+    defaultEnableMerging: false
+    # Configures the port at which the Consul sidecar will listen on to return
+    # combined metrics. Only applicable if
+    # global.connectInject.metrics.defaultEnabled is true.
+    defaultMergedMetricsPort: 20100
+    # Configures the port Prometheus will scrape metrics from, by configuring the Pod
+    # annotation prometheus.io/port and the corresponding listener in the Envoy
+    # sidecar. Only applicable if global.connectInject.metrics.defaultEnabled
+    # is true.
+    defaultPrometheusScrapePort: 20200
```
lkysow, line 1302: Should we use 9102? I see that in a couple prometheus articles and defaults.

templates/connect-inject-deployment.yaml
```diff
@@ -94,6 +94,13 @@ spec:
                 -health-checks-reconcile-period={{ .Values.connectInject.healthChecks.reconcilePeriod }} \
                 {{- end }}
                 -cleanup-controller-reconcile-period={{ .Values.connectInject.cleanupController.reconcilePeriod }} \
+                {{- if (or .Values.connectInject.metrics.defaultEnabled (and .Values.global.metrics.enabled (eq (.Values.connectInject.metrics.defaultEnabled | toString) "-"))) }}
```
lkysow, line 97: I think there is still value in having defaults for port, etc, even if by default metrics are **disabled**.

e.g. what if I want metrics to be disabled by default so I can enable them for specific pods via annotation but for all of those pods, I want the default scraping port to be 9102.

In that case I'd like to be able to set `defaultScrapePort: 9102`, `defaultEnabled: false`

lkysow (commented):
I tested this out and it works great. Question about some logic.

values.yaml
ndhanushkodi: Thanks for bringing this up! I also see that in some articles, and it sounds like a reasonable change.

templates/connect-inject-deployment.yaml
ndhanushkodi: That does sound like a reasonable use case, and great catch! 

The rest of the metrics values (`default-enable-merging`, etc) say "Only applicable if metrics.defaultEnabled is true". I think this use case makes sense, so I can remove that sentence from the other metrics values, and that means those flags with those default values will always be passed to the connect-inject command, independent of the `defaultEnabled` value, which seems fine.

values.yaml
```diff
@@ -1271,6 +1271,41 @@ connectInject:
...
+    # If true, this value will enable the connect-injector to automatically
+    # add prometheus annotations to connect-injected pods. It will also 
+    # add a listener on the Envoy sidecar to expose metrics. The exposed
+    # metrics will depend on whether metrics merging is enabled:
+    #   - If metrics merging is enabled:
+    #     the Consul sidecar will run a merged metrics server
+    #     combining Envoy sidecar and Connect service metrics. 
+    #   - If metrics merging is disabled:
+    #     the listener will just expose Envoy sidecar metrics.
+    # This will inherit from global.metrics.enabled.
```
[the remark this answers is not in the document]
ndhanushkodi: Yup, the bullet points look okay here: https://consul-6hgxgzcp0-hashicorp.vercel.app/docs/k8s/helm

values.yaml
lkysow: Oops sorry, to follow up on this I was incorrectly thinking this was the app's metrics port, not the one we tell prometheus to scrape at. So actually keeping this as 20200 is probably a better call because then it's less likely to conflict with the app's metrics port.

templates/connect-inject-deployment.yaml
lkysow: Sounds good!

values.yaml
```diff
@@ -1271,6 +1271,41 @@ connectInject:
...
+    defaultEnabled: “-”
+    # Configures the Consul sidecar to run a merged metrics server
+    # to combine and serve both Envoy and Connect service metrics.
+    # Only applicable if `global.connectInject.metrics.defaultEnabled` is
+    # true.
+    defaultEnableMerging: false
+    # Configures the port at which the Consul sidecar will listen on to return
+    # combined metrics. Only applicable if
+    # global.connectInject.metrics.defaultEnabled is true.
+    defaultMergedMetricsPort: 20100
```
lkysow, line 1297:
```suggestion
    # Configures the port at which the Consul sidecar will listen on to return
    # combined metrics. This port only needs to be changed if it conflicts
    # with the application's ports.
    defaultMergedMetricsPort: 20100
```

I was thinking we could give some indication that 99% of the time users won't need to change this?

values.yaml
```diff
@@ -1271,6 +1271,41 @@ connectInject:
...
+    # Configures the port Prometheus will scrape metrics from, by configuring the Pod
+    # annotation `prometheus.io/port` and the corresponding listener in the Envoy
+    # sidecar. Only applicable if global.connectInject.metrics.defaultEnabled
+    # is true.
+    defaultPrometheusScrapePort: 9102
```
lkysow, line 1302:
```suggestion
    # Configures the port Prometheus will scrape metrics from, by configuring the Pod
    # annotation `prometheus.io/port` and the corresponding listener in the Envoy
    # sidecar.
    # NOTE: This is *not* the port that your application exposes metrics on. That can
    # be configured with the `consul.hashicorp.com/service-metrics-port` annotation.
    defaultPrometheusScrapePort: 9102
```

Again maybe something here to help with confusion.

values.yaml
```diff
@@ -1271,6 +1271,41 @@ connectInject:
...
+    # Configures the path Prometheus will scrape metrics from, by configuring the pod
+    # annotation prometheus.io/path and the corresponding handler in the Envoy
+    # sidecar. Only applicable if global.connectInject.metrics.defaultEnabled
+    # is true.
+    defaultPrometheusScrapePath: "/metrics"
```
lkysow, line 1307:
```suggestion
    # Configures the path Prometheus will scrape metrics from, by configuring the pod
    # annotation prometheus.io/path and the corresponding handler in the Envoy
    # sidecar.
    # NOTE: This is *not* the path that your application exposes metrics on. That can be
    # configured with the `consul.hashicorp.com/service-metrics-path` annotation.
    defaultPrometheusScrapePath: "/metrics"
```

values.yaml
ndhanushkodi: Good idea

values.yaml
ndhanushkodi: omg yeah I missed that too, since in most of these articles you just direct prometheus to directly scrape the app metrics port!

.circleci/config.yml
```diff
@@ -144,7 +144,7 @@ jobs:
                     -kubecontext="kind-dc1" \
                     -secondary-kubecontext="kind-dc2" \
                     -debug-directory="$TEST_RESULTS/debug" \
-                    -consul-k8s-image=docker.mirror.hashicorp.services/hashicorpdev/consul-k8s:latest
+                    -consul-k8s-image=gcr.io/nitya-293720/consul-k8s-dev:metrics30
```
lkysow, line 147: reminder to remove

values.yaml
```diff
@@ -1271,6 +1271,45 @@ connectInject:
...
+    # add prometheus annotations to connect-injected pods. It will also
+    # add a listener on the Envoy sidecar to expose metrics. The exposed
+    # metrics will depend on whether metrics merging is enabled:
+    #   - If metrics merging is enabled:
+    #     the Consul sidecar will run a merged metrics server
+    #     combining Envoy sidecar and Connect service metrics,
+    #     i.e. if your service exposes its own Prometheus metrics.
+    #   - If metrics merging is disabled:
+    #     the listener will just expose Envoy sidecar metrics.
+    # This will inherit from `global.metrics.enabled.`
```
lkysow, line 1287:
```suggestion
    # This will inherit from `global.metrics.enabled`.
```

values.yaml
```diff
@@ -1271,6 +1271,45 @@ connectInject:
...
+    defaultEnabled: "-"
+    # Configures the Consul sidecar to run a merged metrics server
+    # to combine and serve both Envoy and Connect service metrics.
+    # Only applicable if `global.connectInject.metrics.defaultEnabled` is
+    # true.
```
lkysow, line 1292: Is this still true? I thought defaultEnabled could be false but this could be true and then you could enable metrics via annotation `consul.hashicorp.com/enable-metrics` and in that case merging would be default enabled?

lkysow (approved):

values.yaml
ndhanushkodi: You're absolutely right

.circleci/config.yml
ndhanushkodi: Could we merge it with this into feature-metrics so acceptance tests on the branch can stay green, and then when we merge feature-metrics into master we could remove it?

lkysow: oh yeah of course!

PR #841, opened 2021-02-24
Skip cleanup ctrl namespace test if -enable-enterprise is not set

ishustava:


kschoche (approved):
Good catch!

lkysow (approved):
Do the tests need to be fixed too? Wondering how they passed?

Do newer versions of Consul not error out if you pass `Namespace` as a query param maybe?

ishustava:
> Do the tests need to be fixed too? Wondering how they passed?

That's because all tests in ci have -enable-enterprise set, which is global setting. When you that, acceptance tests framework will use an enterprise image for all tests.

PR #854, opened 2021-03-09
Fix: coalesce.go warning while overriding the client affinity with a string value

washtubs:
Create a local-values.yaml like the example shown in the values.yaml in the docs above client affinity:

    client:
      affinity: |
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: node-role.kubernetes.io/master
                operator: DoesNotExist

If you run `helm template -f ./local-values.yaml consul-test . | head`, it will print a warning like so:

    coalesce.go:196: warning: cannot overwrite table with non table for affinity (map[])

The end result is fine, but the warning is annoying. My understanding is when it's "coalescing" your local-values.yaml with the values.yaml it doesn't like it when a string is replacing an object.

Changes proposed in this PR:
- Default client affinity to null instead of an empty map prevents the coalesce.go warning from being printed

How I've tested this PR:

Checklist:
- [ ] Bats tests added

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -768,7 +768,7 @@ client:
   #           operator: DoesNotExist
   # ```
   # @type: string
-  affinity: {}
+  affinity: null
 
   # This value references an existing
   # Kubernetes `priorityClassName` (https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#pod-priority)


ndhanushkodi (approved):
It looks good to me! On L754 we do the same with nodeSelector, where it's type is string and we default it to `null`, and seems reasonable to do the same here since it's not a map.

thisisnotashwin (approved):

PR #855, opened 2021-03-10
Add backticks in comments where required

thisisnotashwin:
Changes proposed in this PR:
- Add backticks to helm comments.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -237,7 +237,7 @@ global:
     enabled: false
 
     # Configures consul agent metrics. Only applicable if
-    # global.metrics.enabled is true.
+    # `global.metrics.enabled` is true.
     # @type: boolean
     enableAgentMetrics: false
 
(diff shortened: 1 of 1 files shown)


values.yaml
```diff
@@ -1040,7 +1040,7 @@ ui:
     provider: "prometheus"
 
     # baseURL is the URL of the prometheus server, usually the service URL.
```
ndhanushkodi, line 1042: L1038 also needs ui.enabled in backticks

values.yaml
```diff
@@ -1040,7 +1040,7 @@ ui:
...
-    # This value is only used if ui.enabled is set to true.
+    # This value is only used if `ui.enabled` is set to true.
     # @type: string
     baseURL: http://prometheus-server
```
ndhanushkodi, line 1045: Also L1304 needs prometheus.io/path in backticks

ndhanushkodi (approved):
🎉

values.yaml
thisisnotashwin: good catch!!

values.yaml
thisisnotashwin: good catch!!

PR #858, opened 2021-03-11
Acceptance tests fixes

ishustava:
Sometimes resources created by terraform for tests don't cleaned up properly during terraform destroy. As a result, over time we accumulate resources in AWS and Azure that we then have to delete manually. This PR proposes to add a tag/label containing either this build URL or build number so we can better track why those resources were not removed.

Changes proposed in this PR:
- Pass either build URL or build number to terraform resources for GCP/AWS/Azure. On GKE, we can't pass build URL because `/` is not a valid character in labels
- Remove scripting logic that makes tests fail early (we don't need it in nightly acceptance tests)
- Fix metrics merging test on GKE. We are running tests there with enterprise license enabled, but since we're setting the consul image to OSS one, we should also unset enterprise license values so that we don't try to run the enterprise license commands against OSS binary,  where they don't exist.

How I've tested this PR:
Enabled all tests in this PR (https://app.circleci.com/pipelines/github/hashicorp/consul-helm/2590/workflows/8925e10f-ddb8-4560-ba4b-c99f08a830a7) and checked that created resources have tags or labels.

Code review

### DIFF
--- a/test/terraform/aks/main.tf
+++ b/test/terraform/aks/main.tf
@@ -14,6 +14,8 @@ resource "azurerm_resource_group" "default" {
   count    = var.cluster_count
   name     = "consul-k8s-${random_id.suffix[count.index].dec}"
   location = var.location
+
+  tags = var.tags
 }
 
 resource "azurerm_kubernetes_cluster" "default" {
@@ -39,6 +41,8 @@ resource "azurerm_kubernetes_cluster" "default" {
   role_based_access_control {
     enabled = true
   }
+
+  tags = var.tags
 }
 
 resource "local_file" "kubeconfigs" {
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -345,7 +345,7 @@ jobs:
           command: |
             terraform init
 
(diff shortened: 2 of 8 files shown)


test/terraform/aks/main.tf
```diff
@@ -14,6 +14,8 @@ resource "azurerm_resource_group" "default" {
   count    = var.cluster_count
   name     = "consul-k8s-${random_id.suffix[count.index].dec}"
   location = var.location
+
+  tags = var.tags
```
kschoche, line 18:
```suggestion
  tags = var.tags
```

kschoche: This was just a super minor suggestion but looks like you're doing this same thing elsewhere, no need to change it.

kschoche (approved):
Looks great!

test/terraform/aks/main.tf
ishustava: yeah, this was intentional.

thisisnotashwin (approved):
Looks great!! I like the idea of not exiting early on nightly tests!

PR #863, opened 2021-03-16
Add `topologySpreadConstraints` support for server pods

lawliet89:
Changes proposed in this PR:

- Add `topologySpreadConstraints` support for server pods

How I've tested this PR:

- Deploy a new Consul cluster with `server.topologySpreadConstraints` set.

- Deploy a new Consul cluster with `server.topologySpreadConstraints` set.

Checklist:
- [x] Bats tests added

Note:

This helps to spread server pods evenly into all AZs.

- This is best used with `podAntiAffinity` set to the topology key.
- Cluster Autoscaler does not understand this at the moment. See https://github.com/kubernetes/autoscaler/issues/3879

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -64,7 +64,7 @@ spec:
       tolerations:
         {{ tpl .Values.server.tolerations . | nindent 8 | trim }}
     {{- end }}
-    {{- if .Values.server.topologySpreadConstraints }}
+    {{- if and (.Values.server.topologySpreadConstraints) (and (ge .Capabilities.KubeVersion.Major "1") (ge .Capabilities.KubeVersion.Minor "18")) }}
       topologySpreadConstraints:
         {{ tpl .Values.server.topologySpreadConstraints . | nindent 8 | trim }}
     {{- end }}
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -64,6 +64,10 @@ spec:
       tolerations:
         {{ tpl .Values.server.tolerations . | nindent 8 | trim }}
     {{- end }}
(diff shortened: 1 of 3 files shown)


templates/server-statefulset.yaml
```diff
@@ -63,6 +63,10 @@ spec:
     {{- if .Values.server.tolerations }}
       tolerations:
         {{ tpl .Values.server.tolerations . | nindent 8 | trim }}
+    {{- end }}
+    {{- if .Values.server.topologySpreadConstraints }}
```
thisisnotashwin, line 67: Could you add a check here similar to:
`{{- if or ( gt .Capabilities.KubeVersion.Major "1" ) ( ge .Capabilities.KubeVersion.Minor "18" ) }}` ?

There is a warning in the values file, but this might be helpful in ensuring it doesn't accidentally get templated when say re-using a values file. We have tests that do something similar in `ui-ingress.bats` for reference!

thisisnotashwin (commented):
Thanks so much for the contribution @lawliet89 !! This looks great. Just added a comment about ensuring we template the value for only valid kube versions.

templates/server-statefulset.yaml
lawliet89: I have added the checks.

A side note: do you think it's better to just have this fail instead of being silently ignored when a user tries to use this on a cluster < 1.18 though?

thisisnotashwin: YES! I like this idea a lot. Would you mind updating the PR to fail if the user attempts to set this value?

lawliet89: I have added that in.

thisisnotashwin (approved):
🙏 thank you @lawliet89

PR #877, opened 2021-03-19
Add back kubeVersion with wildcard

david-yu:
Since Helm uses `Masterminds/semver` previous `kubeVersion` fields with any additional alphanumeric characters after the version would cause installs to fail on GKE and EKS. Since https://github.com/Masterminds/semver/pull/70, we should now be able pass the pre-install check since ` 1.13.0-gke-version` < `1.13.0-0 `, `1.13.0-eks-version` < `1.13.0-0`, and `1.13.0` < `1.13.0-0`.

Changes proposed in this PR:
- Add back `kubeVersion` with wildcard value `1.13.0-0`

How I've tested this PR:

Checklist:
- [ ] Bats tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 BREAKING CHANGES 
 
-* Specify `kubeVersion` in `Chart.yaml` to denote that this chart is tested with Kubernetes 1.13+ [[GH-877](https://github.com/hashicorp/consul-helm/pull/877]
+* Specify `kubeVersion` in `Chart.yaml` to denote that this chart is compatible with Kubernetes 1.13+. [[GH-877](https://github.com/hashicorp/consul-helm/pull/877]
 
 ## 0.31.1 (Mar 19, 2021)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Unreleased
 
+BREAKING CHANGES 
+
+* Specify `kubeVersion` in `Chart.yaml` to denote that this chart is tested with Kubernetes 1.13+ [[GH-877](https://github.com/hashicorp/consul-helm/pull/877]
```
ishustava, line 5: We don't test against kube 1.13 so I think it's best to not say it. 
```suggestion
* Specify `kubeVersion` in `Chart.yaml` to denote that this chart is compatible with Kubernetes 1.13+. [[GH-877](https://github.com/hashicorp/consul-helm/pull/877]
```

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Unreleased
 
+BREAKING CHANGES
```
ishustava, line 3: Should it be an improvement? Hopefully, it shouldn't break anyone 🤞 
```suggestion
IMPROVEMENTS: 
```

ishustava (approved):
✅ 

The acceptance tests failed due to reasons unrelated to this change, so it's good to go 👍

PR #878, opened 2021-03-21
feat: support customized nodePort in ui service

jhandguy:
Changes proposed in this PR:
- add optional nodePort for http or/and https ports in ui service

Checklist:
- [x] Bats tests added

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -984,6 +984,19 @@ ui:
     # @type: string
     type: null
 
+    # Optionally hardcode the nodePort of the ui service if using a NodePort service.
+    # If not set and using a NodePort service, Kubernetes will automatically assign
+    # a port.
+    nodePort:
+
+      # HTTP node port
+      # @type: integer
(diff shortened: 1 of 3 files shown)


values.yaml
```diff
@@ -984,6 +984,19 @@ ui:
     # @type: string
     type: null
 
+    # Optionally hardcode the nodePort of the ui service if using a NodePort service.
```
thisisnotashwin, line 987:
```suggestion
    # Optionally set the nodePort value of the ui service if using a NodePort service.
```

thisisnotashwin (approved):
This is excellent!! Thanks so much for the contribution @jhandguy! Have suggested a minor update to the comment but it looks great otherwise!!

ishustava (approved):
Looks great!

thisisnotashwin:
Thanks again for the contribution @jhandguy !! Have a great weekend!

PR #880, opened 2021-03-22
use non-root folder to create apply-license script

osdemah:
Changes proposed in this PR:
- Use `/tmp/scripts` folder instead of the WORKDIR which is `/` to avoid getting permission denied errors when security context defaults to the non-root users

How I've tested this PR:
- I installed the helm chart with enterprise license on a k8s cluster that doesn't allow root user in security context by default and made sure the job runs Ok.

- Inatall the helm chart in an openshift cluster with enterprise license and compare the pre-PR and post-PR behavior. The chart gets failed with permission denied error as openshift assigns a non-root user to the pods by default.

Checklist:
- [ ] Bats tests added

### DIFF
--- a/templates/enterprise-license-job.yaml
+++ b/templates/enterprise-license-job.yaml
@@ -77,7 +77,8 @@ spec:
             - "-c"
             - |
                 # Create a script that we can execute with the timeout command.
-                cat > apply-license.sh << 'EOF'
+                mkdir -p /tmp/scripts/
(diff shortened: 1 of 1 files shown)


david-yu:
Hi @osdemah thank you for being a customer and we appreciate the PR. We might need to add some Bats tests to make sure that this is caught in our CI and a reference to our Changelog.

lkysow (approved):
Great work! Thanks for the PR.

I tested it out and the license applied as expected.

PR #884, opened 2021-03-24
AKS nightly tests need more time.

ndhanushkodi:
They panicked twice in the last 3 days after the 40m timeout.


.circleci/config.yml
```diff
@@ -278,7 +278,7 @@ jobs:
             export primary_kubeconfig=$(terraform output -state ../../terraform/aks/terraform.tfstate -json | jq -r .kubeconfigs.value[0])
             export secondary_kubeconfig=$(terraform output -state ../../terraform/aks/terraform.tfstate -json | jq -r .kubeconfigs.value[1])
 
-            gotestsum --junitfile "$TEST_RESULTS/gotestsum-report.xml" -- ./... -p 1 -timeout 40m -failfast \
+            gotestsum --junitfile "$TEST_RESULTS/gotestsum-report.xml" -- ./... -p 1 -timeout 45m -failfast \
```
lkysow, line 281: should we bump it way up? I also see there's a no_output_timeout that we may need to configure.

ndhanushkodi: yeah good call, I did change it to 2 hours, and bumped the no_output_timeout. If the tests are succeeding, they do seem to provide output, when I run it locally. I'm going to experiment with no_output_timeout of 10 min and see if it's ok to keep that lower.

lkysow: > If the tests are succeeding, they do seem to provide output, when I run it locally

Oh really? I would have thought that if they don't have `-v` they provide no output, at least when run via gotestsum (e.g. https://app.circleci.com/pipelines/github/hashicorp/consul-helm/2671/workflows/741a306e-5227-4e73-8f92-ed2bae5425d0/jobs/9752)

ndhanushkodi: You're right! They did fail with the lower no_output_time, for the reasons you said so I increased that to be slightly longer than the test timeout. They can probably be equal, if we don't expect the tests to take right up till the time limit

.circleci/config.yml
```diff
@@ -616,7 +616,7 @@ workflows:
             - go-fmt-and-vet-helm-gen
       - test-helm-gen
       - unit-helm
-      - acceptance:
+      - acceptance-aks-1-18:
```
lkysow, line 619: reminder to change this back

lkysow (approved):

thisisnotashwin (approved):

Issue #869, opened 2021-03-18
Consul Helm Sync Catalog Crashloopback because ReadinessProbe


petrukngantuk:
When filing a bug, please include the following headings if possible. Any example text in this template can be deleted.

### Overview of the Issue

`consul-sync-catalog` crashloopback because kubernetes detect readiness probe have 500 response 

### Reproduction Steps

Steps to reproduce this issue, eg:

1. When running helm install with the following `values.yml`:
```
client:
  enabled: false
global:
  acls:
    bootstrapToken:
      secretKey: null
      secretName: null
    createReplicationToken: false
    manageSystemACLs: false
    replicationToken:
      secretKey: null
      secretName: null
  datacenter: dc1
  domain: consul
  enableConsulNamespaces: false
  enablePodSecurityPolicies: false
  enabled: true
  federation:
    createFederationSecret: false
    enabled: false
  gossipEncryption:
    secretKey: key
    secretName: consul-gossip-encryption-key
  image: hashicorp/consul:1.9.3
  imageEnvoy: envoyproxy/envoy-alpine:v1.16.0
  imageK8S: hashicorp/consul-k8s:0.24.0
  imagePullSecrets: []
  name: consul
  openshift:
    enabled: false
  tls:
    caCert:
      secretKey: null
      secretName: null
    caKey:
      secretKey: null
      secretName: null
    enableAutoEncrypt: true
    enabled: true
    httpsOnly: true
    serverAdditionalDNSSANs: []
    serverAdditionalIPSANs: []
    verify: true
server:
  affinity: |
    podAntiAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        - labelSelector:
            matchLabels:
              app: {{ template "consul.name" . }}
              release: "{{ .Release.Name }}"
              component: server
          topologyKey: kubernetes.io/hostname
  annotations: null
  bootstrapExpect: null
  connect: true
  disruptionBudget:
    enabled: true
    maxUnavailable: null
  enabled: '-'
  exposeGossipAndRPCPorts: true
  extraConfig: |
    {}
  extraEnvironmentVars: {}
  extraLabels: null
  extraVolumes: []
  image: null
  nodeSelector: null
  ports:
    serflan:
      port: 8301
  priorityClassName: ""
  replicas: 3
  resources:
    limits:
      cpu: 500m
      memory: 500Mi
    requests:
      cpu: 100m
      memory: 100Mi
  securityContext:
    fsGroup: 1000
[...]

PR #887, opened 2021-03-29
sync-catalog to use consul-ca-cert with client disabled and auto encrypt enabled

kschoche:
Changes proposed in this PR:
- With `client.enabled=false` and `tls.autoEncrypt.enabled=true` sync-catalog needs to use`consul-ca-cert` instead of the autoencrypt CA cert, otherwise it will not be able to use it's readinessProbe which issues consul client API calls.
- Add a check for each place where we test autoEncrypt for if the client is enabled and use the correct CA cert accordingly.

How I've tested this PR:
deployed with : 
```
client:
  enabled: false
server:
  replicas: 1
global:
  enabled: true
  image: hashicorp/consul:1.9.3
  imageEnvoy: envoyproxy/envoy-alpine:v1.16.0
  imageK8S: hashicorp/consul-k8s:0.25.0
  acls:
    manageSystemACLs: true
  tls:
    enableAutoEncrypt: true
    enabled: true
syncCatalog:
  enabled: true
```
And saw that the failure in GH-869 reproduces : `x509: certificate signed by unknown authority` from the readinessProbe.
With proposed changes this goes away and sync-catalog becomes Ready.

Deploy with a build of this branch and yaml above with correct images, sync-catalog should become Ready.
Code review.

Checklist:
- [x] Bats tests added

### DIFF
--- a/templates/sync-catalog-deployment.yaml
+++ b/templates/sync-catalog-deployment.yaml
@@ -1,3 +1,4 @@
+{{ $clientEnabled := ternary true false (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }} 
 {{- if (or (and (ne (.Values.syncCatalog.enabled | toString) "-") .Values.syncCatalog.enabled) (and (eq (.Values.syncCatalog.enabled | toString) "-") .Values.global.enabled)) }}
 # The deployment for running the sync-catalog pod
 apiVersion: apps/v1
@@ -43,7 +44,7 @@ spec:
           - key: {{ default "tls.crt" .Values.global.tls.caCert.secretKey }}
             path: tls.crt
       {{- end }}
-      {{- if .Values.global.tls.enableAutoEncrypt }}
+      {{- if (and .Values.global.tls.enableAutoEncrypt $clientEnabled) }}
       - name: consul-auto-encrypt-ca-cert
         emptyDir:
           medium: "Memory"
@@ -96,7 +97,7 @@ spec:
             {{- end }}
           {{- if .Values.global.tls.enabled }}
           volumeMounts:
-            {{- if .Values.global.tls.enableAutoEncrypt }}
+            {{- if (and .Values.global.tls.enableAutoEncrypt $clientEnabled) }}
             - name: consul-auto-encrypt-ca-cert
             {{- else }}
             - name: consul-ca-cert
@@ -193,7 +194,7 @@ spec:
           resources:
             {{- toYaml . | nindent 12 }}
           {{- end }}
-      {{- if or .Values.global.acls.manageSystemACLs (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt) }}
+      {{- if or .Values.global.acls.manageSystemACLs (and (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt) $clientEnabled) }}
       initContainers:
       {{- if .Values.global.acls.manageSystemACLs }}
       - name: sync-acl-init
@@ -213,7 +214,7 @@ spec:
             memory: "25Mi"
             cpu: "50m"
       {{- end }}
-      {{- if (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt) }}
+      {{- if (and (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt) $clientEnabled) }}
       {{- include "consul.getAutoEncryptClientCA" . | nindent 6 }}
       {{- end }}
       {{- end }}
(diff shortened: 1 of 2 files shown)


templates/sync-catalog-deployment.yaml
```diff
@@ -1,3 +1,4 @@
+{{ $clientEnabled := ternary true false (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}
```
lkysow, line 1:
```suggestion
{{- $clientEnabled := (or (and (ne (.Values.client.enabled | toString) "-") .Values.client.enabled) (and (eq (.Values.client.enabled | toString) "-") .Values.global.enabled)) }}
```

1. I think we can assign directly to the variable
2. Should have the whitespace chomping

templates/sync-catalog-deployment.yaml
```diff
@@ -193,7 +194,7 @@ spec:
           resources:
             {{- toYaml . | nindent 12 }}
           {{- end }}
-      {{- if or .Values.global.acls.manageSystemACLs (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt) }}
+      {{- if or .Values.global.acls.manageSystemACLs (and (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt) $clientEnabled) }}
```
lkysow, line 197:
```suggestion
      {{- if or .Values.global.acls.manageSystemACLs (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt $clientEnabled) }}
```

We don't need nested `and`'s since `and` will and all arguments: 

```
and
	Returns the boolean AND of its arguments by returning the
	first empty argument or the last argument, that is,
	"and x y" behaves as "if x then y else x". All the
	arguments are evaluated.
```

templates/sync-catalog-deployment.yaml
```diff
@@ -213,7 +214,7 @@ spec:
             memory: "25Mi"
             cpu: "50m"
       {{- end }}
-      {{- if (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt) }}
+      {{- if (and (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt) $clientEnabled) }}
```
lkysow, line 217:
```suggestion
      {{- if (and .Values.global.tls.enabled .Values.global.tls.enableAutoEncrypt $clientEnabled) }}
```

test/unit/sync-catalog-deployment.bats
```diff
@@ -517,6 +517,32 @@ load _helpers
   [ "${actual}" = "key" ]
 }
 
+@test "syncCatalog/Deployment: consul-ca-cert volume is used when TLS with auto-encrypt is enabled and client.enabled=false" {
```
lkysow, line 520: Typically we try and test every conditional. In this case, the `consul-ca-cert` volume conditional does not involve `client.enabled` so I don't think this test is required:


```yaml
{- if not (and .Values.externalServers.enabled .Values.externalServers.useSystemRoots) }}
      - name: consul-ca-cert
        secret:
          {{- if .Values.global.tls.caCert.secretName }}
          secretName: {{ .Values.global.tls.caCert.secretName }}
          {{- else }}
          secretName: {{ template "consul.fullname" . }}-ca-cert
          {{- end }}
          items:
          - key: {{ default "tls.crt" .Values.global.tls.caCert.secretKey }}
            path: tls.crt
      {{- end }}
```

test/unit/sync-catalog-deployment.bats
```diff
@@ -517,6 +517,32 @@ load _helpers
...
+      --set 'syncCatalog.enabled=true' \
+      --set 'global.tls.enabled=true' \
+      --set 'global.tls.enableAutoEncrypt=true' \
+      --set 'client.enabled=false' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.volumes[] | select(.name == "consul-ca-cert") | length > 0' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
+@test "syncCatalog/Deployment: consul-auto-encrypt-ca-cert volume is not added with auto-encrypt and client.enabled=false" {
```
lkysow, line 533: Whereas this test makes a lot of sense because this changes based on the value of `client.enabled`. Nice work.

test/unit/sync-catalog-deployment.bats
```diff
@@ -848,6 +887,25 @@ load _helpers
   [ "${actual}" = 'http://RELEASE-NAME-consul-server:8500' ]
 }
 
+@test "syncCatalog/Deployment: consul service is used when client.enabled=false and global.tls.enabled=true and autoencrypt on" {
```
lkysow, line 890: The `CONSUL_HTTP_ADDR` and `CONSUL_CACERT` vars don't change based on auto encrypt so I don't think we need this test either.

The four permutations: clients enabled true/false x tls enabled true/false, are already tested by the four tests in this section.

lkysow (approved):
Great work. I tested this out locally.

Couple comments on the tests and templates. Also we need a changelog entry.

PR #891, opened 2021-04-01
use consul-ca-cert when client disabled for sync catalog and auto encrypt

kschoche:
Changes proposed in this PR:

With client.enabled=false and tls.autoEncrypt.enabled=true sync-catalog needs to useconsul-ca-cert instead of the autoencrypt CA cert, otherwise it will not be able to use it's readinessProbe which issues consul client API calls.
Add a check for each place where we test autoEncrypt for if the client is enabled and use the correct CA cert accordingly.
How I've tested this PR:
deployed with :
```
client:
  enabled: false
server:
  replicas: 1
global:
  enabled: true
  image: hashicorp/consul:1.9.3
  imageEnvoy: envoyproxy/envoy-alpine:v1.16.0
  imageK8S: hashicorp/consul-k8s:0.25.0
  acls:
    manageSystemACLs: true
  tls:
    enableAutoEncrypt: true
    enabled: true
syncCatalog:
  enabled: true
```
And saw that the failure in GH-869 reproduces : x509: certificate signed by unknown authority from the readinessProbe.
With proposed changes this goes away and sync-catalog becomes Ready.

Deploy with a build of this branch and yaml above with correct images, sync-catalog should become Ready.
Code review.

Checklist:
- [x] Bats tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 IMPROVEMENTS: 
 
-* Use consul-ca-cert for sync-catalog when using autoencrypt and client is disabled. [[GH-891](https://github.com/hashicorp/consul-helm/pull/891)]
 * Specify `kubeVersion` in `Chart.yaml` to denote that this chart is compatible with Kubernetes 1.16+. [[GH-883](https://github.com/hashicorp/consul-helm/pull/883)]
 * CRDs: update the CRD versions from v1beta1 to v1. [[GH-883](https://github.com/hashicorp/consul-helm/pull/883)]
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 IMPROVEMENTS: 
 
+* Use consul-ca-cert for sync-catalog when using autoencrypt and client is disabled. [[GH-891](https://github.com/hashicorp/consul-helm/pull/891)]
```
lkysow, line 5: I think this is a bugfix since we did allow sync catalog without clients before, it just would break in a certain configuration. I'd say something like "Sync Catalog: fix issue running with clients disabled and auto encrypt enabled"

test/unit/sync-catalog-deployment.bats
```diff
@@ -834,18 +860,23 @@ load _helpers
...
       --set 'client.enabled=false' \
       . | tee /dev/stderr |
       yq -r '.spec.template.spec.containers[0].env[]' | tee /dev/stderr)
 
   local actual
   actual=$(echo $env | jq -r '. | select(.name == "CONSUL_HTTP_ADDR") | .value' | tee /dev/stderr)
-  [ "${actual}" = 'http://RELEASE-NAME-consul-server:8500' ]
+  [ "${actual}" = 'https://RELEASE-NAME-consul-server:8501' ]
+
+  actual=$(echo $env | jq -r '. | select(.name == "CONSUL_CACERT") | .value' | tee /dev/stderr)
```
lkysow, line 878: nice catch!

lkysow (approved):

thisisnotashwin (approved):
This is excellent!!

PR #903, opened 2021-04-08
Support leader election for Connect-Inject controller

thisisnotashwin:
Changes proposed in this PR:
- Support leader election for the connect-inject controller.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -143,7 +143,7 @@ jobs:
                     -kubecontext="kind-dc1" \
                     -secondary-kubecontext="kind-dc2" \
                     -debug-directory="$TEST_RESULTS/debug" \
-                    -consul-k8s-image=kschoche/consul-k8s-dev2:2021-4-9 # TODO: change once feature-tproxy consul-k8s branch is merged
+                    -consul-k8s-image=ashwinvenkatesh/consul-k8s@sha256:f66106414f9e6b05271bd2944bf0a01308a6b9405221d73e89f2d00159189d6e # TODO: change once feature-tproxy consul-k8s branch is merged
               then
                 echo "Tests in ${pkg} failed, aborting early"
                 exit_code=1
--- a/templates/connect-inject-deployment.yaml
+++ b/templates/connect-inject-deployment.yaml
@@ -19,7 +19,7 @@ metadata:
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
 spec:
-  replicas: 1
+  replicas: 2
   selector:
     matchLabels:
       app: {{ template "consul.name" . }}
--- a/templates/connect-inject-clusterrole.yaml
+++ b/templates/connect-inject-clusterrole.yaml
@@ -16,6 +16,15 @@ rules:
   - "get"
   - "list"
   - "watch"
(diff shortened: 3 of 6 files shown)


templates/connect-inject-deployment.yaml
```diff
@@ -19,7 +19,7 @@ metadata:
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
 spec:
-  replicas: 1
+  replicas: 3
```
kschoche, line 22: curious what our thoughts are on this? will 2 cause election issues?
wondering if we bound it to # of server replicas or something, in case this is a demo environment?

thisisnotashwin: 2 should also be fine. I put up 3 as a shot in the dark. the number of instances here are irrelevant. If you have a single instance running, it will assume it is the leader as it does not use a "consensus" protocol to establish leadership.

kschoche: Ah, interesting!
Okay thanks for the clarification. I'm just trying to think of how this might look if we have a demo environment where we've got a single kind node or something and then three of these try to spin up? 2 even seems too high in that case, but would be ideal in any other case. thoughts?

lkysow: I think 2 makes more sense to me. Demo env is tough, it definitely doesn't need 2 but folks won't change this for prod either and we default to 3 servers for the same reason. We could handle demo env another way maybe via profiles.

ishustava: Yeah, we're definitely pretty inconsistent. Although, if I were to guess it seems that we're leaning towards making it prod-ready by default rather than demo by default because servers and all gateways start with multiple instances. The only components that don't are sync-catalog, connect-injector, and controller. Sync catalog and injector could not run more than one instance, and for the controller we could probably switch too especially if we're thinking to combine it with the endpoints controller.

I think it'd make sense to default to a more HA setting (with 2 instances).

kschoche: Agree! 2 sounds good then!

.circleci/config.yml
```diff
@@ -143,7 +143,7 @@ jobs:
                     -kubecontext="kind-dc1" \
                     -secondary-kubecontext="kind-dc2" \
                     -debug-directory="$TEST_RESULTS/debug" \
-                    -consul-k8s-image=kschoche/consul-k8s-dev2:2021-4-9 # TODO: change once feature-tproxy consul-k8s branch is merged
+                    -consul-k8s-image=ashwinvenkatesh/consul-k8s@sha256:f66106414f9e6b05271bd2944bf0a01308a6b9405221d73e89f2d00159189d6e # TODO: change once feature-tproxy consul-k8s branch is merged
```
kschoche, line 146: Just want to confirm this image has rebased off feature-tproxy from yesterrday?

thisisnotashwin: yup!! this is based of the latest changes on `feature-tproxy`

kschoche (approved):
Excellent work!

ndhanushkodi (approved):
🎉 Looks great! Nice work Ashwin!!

PR #905, opened 2021-04-09
Support transparent proxy in the Consul Helm chart

ishustava:
Changes proposed in this PR:
* Add new connectInject.transparentProxy.defaultEnabled value (default to true)
  that will allow users to enable or disable tproxy for each helm installation.
* Add acceptance tests for connect-inject to test with tproxy
* Acceptance tests default to tproxy not enabled since we don't fully support it for all features yet.

How I've tested this PR:
- Enabled tproxy for connect oss and enterprise tests

- code review

Note that for beta we're only enabling connect tests to run with tproxy. Before GA, ideally, we can default to tproxy all the time and test non-tproxy cases explicitly instead.

Checklist:
- [x] Bats tests added


thisisnotashwin (commented):
This looks great. Looking at the failures in the acceptance, I think those fixtures need to be updated in order to ensure they start to use the t-proxy clients/addresses to the remote server.

Great job with this test though. It was super easy to understand!

templates/connect-inject-deployment.yaml
```diff
@@ -90,6 +90,9 @@ spec:
                 -release-name="{{ .Release.Name }}" \
                 -release-namespace="{{ .Release.Namespace }}" \
                 -listen=:8080 \
+                {{- if .Values.connectInject.transparentProxy.defaultEnabled }}
+                -enable-transparent-proxy \
```
ndhanushkodi, line 94: I'm just noticing this now and it's not a blocker for the release, but should this flag be renamed -default-enable-transparent-proxy in consul-k8s after the release? That's how the metrics flag defaults are named.

test/acceptance/framework/k8s/deploy.go
```diff
@@ -133,15 +133,26 @@ func CheckStaticServerConnectionMultipleFailureMessages(
...
-	CheckStaticServerConnection(t, options, false, deploymentName, "curl: (52) Empty reply from server", curlArgs...)
+	t.Helper()
+	CheckStaticServerConnection(t,
+		options,
+		false,
+		deploymentName,
+		[]string{
+			"curl: (52) Empty reply from server",
+			"curl: (7) Failed to connect to static-server port 80: Connection refused",
+			"curl: (7) Failed to connect to static-server.ns1 port 80: Connection refused",
```
ndhanushkodi, line 153: Just curious, did you see these additional errors when testing out tproxy?

ndhanushkodi (approved):
LGTM!! Just had a few non-blocking questions.

templates/connect-inject-deployment.yaml
ishustava: That's a good catch! I'll a task to rename.

thisisnotashwin (approved):
exquisite!!

test/acceptance/framework/k8s/deploy.go
ishustava: Yeah, when with iptables we now get a connection refused. This is most likely because envoy itself will now reject the connection, whereas previously this error was coming from the upstream proxy that verifies intentions.

PR #912, opened 2021-04-14
Update changelog for the tproxy feature

ishustava:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,47 +13,79 @@ BREAKING CHANGES:
 * Connect: The Helm values for health checks and cleanup controller have been removed: `connectInject.healthChecks` and `connectInject.cleanupController`, as these controllers have been replaced by the endpoints controller. [[GH-899](https://github.com/hashicorp/consul-helm/pull/899)]
 * Connect: connect webhook deployment now uses `webhook-cert-manager` to bootstrap the webhook certificates instead of generating them inside of the webhook. [[GH-861](https://github.com/hashicorp/consul-helm/pull/861)]
 * Connect: Kubernetes Services are now required for all connect injected applications.
-  The Kubernetes service name will be used as the service name to register with Consul unless the annotation `consul.hashicorp.com/connect-service` is provided to the deployment/pod to override this. If using ACLs the ServiceAccountName must match the service name used with Consul.
+  The Kubernetes service name will be used as the service name to register with Consul unless the annotation `consul.hashicorp.com/connect-service` is provided to the pod to override this.
+  If using ACLs the ServiceAccountName must match the service name used with Consul.
 
   Example Service:
-```yaml
----
-apiVersion: v1
-kind: Service
-metadata:
-  name: sample-app
-spec:
-  selector:
-    app: sample-app
-  ports:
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,35 @@ BREAKING CHANGES:
 * Minimum Kubernetes versions supported is 1.16+. [[GH-883](https://github.com/hashicorp/consul-helm/pull/883)]
 * Connect: `-enable-health-checks-controller`, `-health-checks-reconcile-period`, `-cleanup-controller-reconcile-period` have been removed
   and are no longer supported as the controllers have been replaced by the endpoints controller. [[GH-892](https://github.com/hashicorp/consul-helm/pull/899)]
+* Connect: Support transparent proxy. [[GH-905](https://github.com/hashicorp/consul-helm/pull/905)]
+  This feature allows users to reach other services on the Consul Service Mesh by using KubeDNS instead of using `localhost`
+  and enforces all inbound and outbound traffic within a pod to go through the Envoy proxy.
+  Please see Transparent Proxy docs for more information.
```
ndhanushkodi, line 17: So we just talked about a page under service mesh and kubernetes that starts off with the same content, and eventually the kubernetes page would have k8s specific stuff and the service mesh page would be more general. So linking to https://www.consul.io/docs/connect/transparent-proxy would probably work here.

CHANGELOG.md
```diff
@@ -11,6 +11,35 @@ BREAKING CHANGES:
...
+  the `consul.hashicorp.com/transparent-proxy` pod annotation:
+
+  ```yaml
+  ...
+  metadata:
+    name: example
+    labels:
+      app: example
+    annotations:
+      "consul.hashicorp.com/transparent-proxy": "true"
```
thisisnotashwin, line 40: Will t-proxy work if `defaultEnabled` is `false` but the annotation is true?
If not, we should make this value `false` so that is describes how to disable it for a particular service?

ndhanushkodi (approved):
Looks good 🎉 just commented on where we're gonna put the docs.

CHANGELOG.md
thisisnotashwin, line 17: Can we link to some docs here or is that for later?

thisisnotashwin (approved):
This is excellent!! had a minor comment but the rest of the changelog looks good.

CHANGELOG.md
ishustava: yep, it will! In a way the feature itself is always enabled, the helm value controls only whether it should be enabled by default.

CHANGELOG.md
ishustava: Will there be a separate page under the `Kubernetes` section? I think that one would be ideal since it's kubernetes related.

CHANGELOG.md
thisisnotashwin: brilliance!

CHANGELOG.md
ishustava: Alright, I've updated to that link preemptively but we could change it later.

ishustava: done!

PR #914, opened 2021-04-15
Add TransparentProxy to Service and ProxyDefaults

thisisnotashwin:
Changes proposed in this PR:
- Add TransparentProxy to ServiceDefaults and ProxyDefaults
- Only support v1 version of Mutating Webhook

How I've tested this PR: Acceptance tests passed.

How I expect reviewers to test this PR: Code Review

Checklist:
- [ ] Bats tests added

### DIFF
--- a/templates/crd-proxydefaults.yaml
+++ b/templates/crd-proxydefaults.yaml
@@ -87,6 +87,13 @@ spec:
                     description: Mode is the mode that should be used for the upstream connection. One of none, local, or remote.
                     type: string
                 type: object
+              transparentProxy:
+                description: TransparentProxy controls configuration specific to proxies in transparent mode.
+                properties:
+                  outboundListenerPort:
+                    description: The port of the listener where outbound application traffic is being redirected to.
+                    type: integer
+                type: object
             type: object
           status:
             properties:
--- a/templates/crd-servicedefaults.yaml
+++ b/templates/crd-servicedefaults.yaml
@@ -89,6 +89,13 @@ spec:
               protocol:
                 description: Protocol sets the protocol of the service. This is used by Connect proxies for things like observability features and to unlock usage of the service-splitter and service-router config entries for a service.
                 type: string
+              transparentProxy:
+                description: TransparentProxy controls configuration specific to proxies in transparent mode.
+                properties:
+                  outboundListenerPort:
+                    description: The port of the listener where outbound application traffic is being redirected to.
+                    type: integer
+                type: object
             type: object
           status:
             properties:
--- a/test/acceptance/go.mod
+++ b/test/acceptance/go.mod
@@ -4,8 +4,8 @@ go 1.14
 
 require (
 	github.com/gruntwork-io/terratest v0.31.2
-	github.com/hashicorp/consul/api v1.4.1-0.20201015173526-812fe06d6c64
-	github.com/hashicorp/consul/sdk v0.6.0
+	github.com/hashicorp/consul/api v1.4.1-0.20210415000851-62fcf1ff17cd
+	github.com/hashicorp/consul/sdk v0.7.0
 	github.com/stretchr/testify v1.5.1
 	gopkg.in/yaml.v2 v2.2.8
 	k8s.io/api v0.19.3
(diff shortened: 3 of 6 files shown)


test/acceptance/tests/metrics/metrics_test.go
```diff
@@ -42,6 +42,13 @@ func TestComponentMetrics(t *testing.T) {
...
 		"terminatingGateways.gateways[0].name":     "terminating-gateway",
 		"terminatingGateways.gateways[0].replicas": "1",
+
+		// Reduce CPU resource requests because tests were running into CPU scheduling
+		// limits and because we're not really testing performance.
+		"controller.resources.requests.cpu":                   "50m",
+		"ingressGateways.defaults.resources.requests.cpu":     "50m",
+		"terminatingGateways.defaults.resources.requests.cpu": "50m",
+		"meshGateway.resources.requests.cpu":                  "50m",
 	}
```
thisisnotashwin, line 52: These changes are a rebase error. These should go away once `feature-tproxy` is merged into master

templates/crd-servicedefaults.yaml
```diff
@@ -89,6 +89,13 @@ spec:
               protocol:
                 description: Protocol sets the protocol of the service. This is used by Connect proxies for things like observability features and to unlock usage of the service-splitter and service-router config entries for a service.
                 type: string
+              transparentProxy:
+                description: TransparentProxy controls configuration specific to proxies in transparent mode.
+                properties:
+                  outboundListenerPort:
+                    description: The port of the listener where outbound application traffic is being redirected to.
+                    type: integer
+                type: object
```
ishustava, line 98: wondering if our docs should say that we don't support these?

test/acceptance/go.mod
```diff
@@ -4,8 +4,8 @@ go 1.14
 
 require (
 	github.com/gruntwork-io/terratest v0.31.2
-	github.com/hashicorp/consul/api v1.4.1-0.20201015173526-812fe06d6c64
-	github.com/hashicorp/consul/sdk v0.6.0
+	github.com/hashicorp/consul/api v1.4.1-0.20210415000851-62fcf1ff17cd
+	github.com/hashicorp/consul/sdk v0.7.0
```
ishustava, line 8: not sure why this change is needed.

ishustava (approved):
Looks great, Ashwin! One comment about our spec description, but won't block the approval.

templates/crd-proxydefaults.yaml
```diff
@@ -86,6 +86,13 @@ spec:
                     description: Mode is the mode that should be used for the upstream connection. One of none, local, or remote.
                     type: string
                 type: object
+              transparentProxy:
+                description: TransparentProxy controls configuration specific to proxies in transparent mode.
```
kschoche, line 90:
```suggestion
                description: TransparentProxy controls configuration specific to proxies in transparent proxy mode.
```
Thoughts? Do we use "transparent mode" anywhere?

kschoche: Oh I guess it's proxies in transparent mode, 🤷🏽 ?

kschoche (approved):
Looks great! Left one comment but I think it's probably okay!

templates/crd-servicedefaults.yaml
thisisnotashwin: It doesnt say it in the docs but we do have a webhook that rejects any CRD that has this set with an error message that says: Please you annotation <annotation-name>`
I went back and forth on this a little bit and decided to only use the webhook because I figured it would be confusing for the docs to say that it wasn't supported. Do you think it makes sense to also have it in the docs?

test/acceptance/go.mod
thisisnotashwin: We don't strictly need it for the acceptance tests here, but this ensure the acceptance tests run against a version of Consul that does explicitly support the `TransparentProxy` field, even though we do not set it or should allow it to be set. It was done more from a correctness perspective.

templates/crd-proxydefaults.yaml
thisisnotashwin: Exactly. It is about the proxies being in transparent mode 😅

templates/crd-servicedefaults.yaml
ishustava: Hmm yeah, I can see that. Let's keep it as is then.

test/acceptance/go.mod
ishustava: got it, thanks for clarifying!

templates/crd-servicedefaults.yaml
thisisnotashwin: I added this as a part of the commit that adds Mode and UpstreamConfigs to ServiceDefaults. Will update these CRDs as a part of their equivalent helm commit if that is ok?

PR #915, opened 2021-04-15
Fix a bug where we weren't setting -enable-transparent-proxy flag to false

ishustava:
Changes proposed in this PR:
- `enable-transparent-proxy` flag for the `inject-connect` command in consul-k8s defaults to true, but the helm chart was not passing `enable-transparent-proxy` when `connectInject.transparentProxy.defaultEnabled` is set to false.
- A few fixes to acceptance tests:
   - Increase timeout to wait for deployments to come up
   - Add 30sec sleep to all helm installs to compensate for the removed readiness check in the connect-inject webhook.

How I've tested this PR:
- acceptance tests

- code review
- 👁️ acceptance tests

Checklist:
- [x] Bats tests added
- Don't think this needs a separate changelog entry since it'll be part of `feature-tproxy`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,8 +133,8 @@ jobs:
             # exit early if any test fails (-failfast only works within a single
             # package).
             exit_code=0
-            pkgs=$(go list ./... | circleci tests split)
-            echo "Running $pkgs"
+            pkgs=$(go list ./... | circleci tests split --split-by=timings --timings-type=classname)
+            echo "Running $(echo $pkgs | wc -w) packages"
             for pkg in $pkgs
             do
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -133,19 +133,20 @@ jobs:
             # exit early if any test fails (-failfast only works within a single
             # package).
             exit_code=0
-            pkgs=$(go list ./... | circleci tests split)
-            echo "Running $pkgs"
+            pkgs=$(go list ./... | circleci tests split --split-by=timings --timings-type=classname)
+            echo "Running $(echo $pkgs | wc -w) packages:"
+            echo $pkg
```
ishustava, line 138: this is a small optimization. Mostly it doesn't improve the total test runtime by that much

thisisnotashwin (approved):
That is a good catch!! Nice work!!

ndhanushkodi (approved):
Great work catching this!!

PR #921, opened 2021-04-20
Allow unstructed schema for config in ProxyDefaults

ishustava:
Please see hashicorp/consul-k8s#495 for details about why this change is needed

Changes proposed in this PR:
- Update CRD schema for ProxyDefaults based on new generated code from the above consul-k8s PR
- Update acceptance test for proxy defaults to set all possible fields. For the `config` field, we're not setting all acceptable values but I think it's OK since we're trying to check that at least one is accepted by Consul.

How I've tested this PR:
- ran acceptance tests

- code review

Checklist:
- [ ] Bats tests added


thisisnotashwin (approved):
It looks like such a harmless change 😂 Nicely done!!

test/acceptance/tests/controller/controller_test.go
```diff
@@ -92,6 +92,10 @@ func TestController(t *testing.T) {
 					proxyDefaultEntry, ok := entry.(*api.ProxyConfigEntry)
 					require.True(r, ok, "could not cast to ProxyConfigEntry")
 					require.Equal(r, api.MeshGatewayModeLocal, proxyDefaultEntry.MeshGateway.Mode)
+					require.Equal(r, "tcp", proxyDefaultEntry.Config["protocol"])
```
lkysow, line 95: let's also test `foo` and `member` for completeness. Or we can remove those from the fixture. I like testing multiple types, so maybe bool, array, map, string?

ishustava: Hmm, good call. I think I've made an assumption that these are going to be ignored by consul, but they're not!

ishustava: done!

lkysow (approved):

ndhanushkodi (approved):
Nice investigation and catch! Looks great!!

PR #928, opened 2021-04-26
Update proxy defaults

thisisnotashwin:
Changes proposed in this PR:
- Add field `mode` to ProxyDefaults CRD. This field is not expected to be set by actual users of the CRD but the change has been made to ensure consistency between the fields supported by Consul and those supported by the config entry CRDs

How I've tested this PR: Acceptance tests

How I expect reviewers to test this PR: Code review/review description

Checklist:
- [ ] Bats tests added

### DIFF
--- a/templates/crd-proxydefaults.yaml
+++ b/templates/crd-proxydefaults.yaml
@@ -87,6 +87,9 @@ spec:
                     description: Mode is the mode that should be used for the upstream connection. One of none, local, or remote.
                     type: string
                 type: object
+              mode:
+                description: 'Mode can be one of "direct" or "transparent". "transparent" represents that inbound and outbound application traffic is being captured and redirected through the proxy. This mode does not enable the traffic redirection itself. Instead it signals Consul to configure Envoy as if traffic is already being redirected. "direct" represents that the proxy''s listeners must be dialed directly by the local application and other proxies. Note: This cannot be set using the CRD and should be set using annotations on the services that are part of the mesh.'
+                type: string
               transparentProxy:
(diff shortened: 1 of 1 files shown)


templates/crd-proxydefaults.yaml
```diff
@@ -87,6 +87,9 @@ spec:
                     description: Mode is the mode that should be used for the upstream connection. One of none, local, or remote.
                     type: string
                 type: object
+              mode:
+                description: 'Mode can be one of direct or transparent. transparent represents that inbound and outbound and outbound application traffic is being captured and redirected through the proxy. This mode does not enable the traffic redirection itself. Instead it signals Consul to configure Envoy as if traffic is already being redirected. direct represents that the proxy''s listeners must be dialed directly by the local application and other proxies. Note: This cannot be set using the CRD and should be set using annotations on the services that are part of the mesh.'
```
kschoche, line 91:
```suggestion
                description: 'Mode can be one of direct or transparent. transparent represents that inbound and outbound application traffic is being captured and redirected through the proxy. This mode does not enable the traffic redirection itself. Instead it signals Consul to configure Envoy as if traffic is already being redirected. direct represents that the proxy''s listeners must be dialed directly by the local application and other proxies. Note: This cannot be set using the CRD and should be set using annotations on the services that are part of the mesh.'
```
Also I see two `'`'s  on the word `proxy` is that expected?

kschoche (approved):
Looks great, spotted a small typo!

templates/crd-proxydefaults.yaml
thisisnotashwin: Yeah, that gets added because it tries to envelope everything in a single quote but then can differentiate it from an apostrophe

ishustava (approved):

PR #901, opened 2021-04-07
Use disableHostNodeID param in values.yaml to configure Consul client

ilhaan:
Changes proposed in this PR:
- Provide a param in values.yaml to configure `-disable-host-node-id` for consul client. More info on this config [here](https://www.consul.io/docs/agent/options#_disable_host_node_id)

How I've tested this PR:
* Helm `--dry-run` install to ensure generated manifest for client DaemonSet has correct config 
* Installation in a development k8s cluster 
* Ran `bats ./test/unit`. `client/DaemonSet` did not report any issues. 

How I expect reviewers to test this PR:
Either a `--dry-run` install to verify DaemonSet manifest or install in a development cluster. 

Checklist:
- [x] Bats tests added
- [ ] CHANGELOG entry added (*HashiCorp engineers only, community PRs should not add a changelog entry*)

Issue #848, opened 2021-03-01
Does not work with single-node Kubernetes Cluster using kind


weitzjdevk:
### Overview of the Issue

The consul-helm chart fails to be used on single-node Kubernetes Clusters, which are bootstrapped with `kind` (Kubernetes in Docker), when you want to install the consul-agent (in clientmode) and consul-agent (in servermode) on the same node-instance.

Reason being is:

`kind`  uses the same host-node to install multiple pods, and the consul-helm chart uses the host-node-ip for a stable agent IP using the flag: ` -disable-host-node-id=false `

=> therefore using a client and server on the same node, results in conflicting host-node-ids.

The error one is experiencing is similar to the related issue #791 

```bash
    2021-01-27T00:12:51.732Z [WARN]  agent.fsm: EnsureRegistration failed: error="failed inserting node: Error while renaming Node ID: "afa5d55b-ea34-cd37-176f-b24447c850b5": Node name gke-shustava-default-pool-bad0f39a-f890 is reserved by node 7df16f21-b493-a066-f0cb-261a81cba524 with name gke-shustava-default-pool-bad0f39a-f890 (10.24.2.32)"
```

which fixed this behaviour by setting `-disable-host-node-id=false ` (explicitly to `false`)

### Reproduction Steps

Install a client-server Consul-Helm setup using `kind` on a single-node Kubernetes cluster and try to join the client with the server.

### Logs

similar to #791 

```bash
    2021-01-27T00:12:51.732Z [WARN]  agent.fsm: EnsureRegistration failed: error="failed inserting node: Error while renaming Node ID: "afa5d55b-ea34-cd37-176f-b24447c850b5": Node name gke-shustava-default-pool-bad0f39a-f890 is reserved by node 7df16f21-b493-a066-f0cb-261a81cba524 with name gke-shustava-default-pool-bad0f39a-f890 (10.24.2.32)"
```

### Expected behavior

I should be able to run client and server on the same Kubernetes single-node cluster

### Environment details

If not already included, please provide the following:
- `consul` version: 1.9.3
- `consul-k8s` version: 0.24
- `consul-helm` version: 0.30.0
- `values.yaml` used to deploy the helm chart:

Any other information you can provide about the environment/deployment.

single kind Kubernetes cluster.

## Sketched solution to the problem

Make the flag: 

```
-disable-host-node-id=false 
```

configurable in the Helm chart, with a default-value of `false` so it stays backwards compatible with the current setup.
[...]

PR #936, opened 2021-04-29
Remove -disable-host-node-id=false flag

lkysow:
Don't set `-disable-host-node-id=false` on Consul Clients because it causes Clients to
not be able to join the cluster on certain hypervisor or Kind setups. This flag was
added in 0.30.0 to stop Consul logging an error when a Client pod was force deleted
but this error does not affect the pod restarting so removing the flag will have
no real effect.

Fixes https://github.com/hashicorp/consul-helm/pull/901

Checklist:
- [x] Bats tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ FEATURES:
 BUG FIXES:
 * CRDs: Fix a bug where the `config` field in `ProxyDefaults` CR was not synced to Consul because
   `apiextensions.k8s.io/v1` requires CRD spec to have structured schema. [[GH-921](https://github.com/hashicorp/consul-helm/pull/921)]
+* Don't set `-disable-host-node-id=false` on Consul Clients because it causes Clients to
+  not be able to join the cluster on certain hypervisor or Kind setups. This flag was
+  added in 0.30.0 to stop Consul logging an error when a Client pod was force deleted,
+  but this error does not affect the pod restarting so removing the flag will have
+  no real effect. [[GH-936](https://github.com/hashicorp/consul-helm/pull/936)]
 
 BREAKING CHANGES:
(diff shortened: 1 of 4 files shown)


ishustava (approved):
🙏

ndhanushkodi (approved):
🎉

ilhaan:
Kind and other hypervisor types that you mentioned need to set `-disable-host-node-id` to `true` in order to get Consul Clients working correctly (refer to issues: [9467](https://github.com/hashicorp/consul/issues/9467) and [848](https://github.com/hashicorp/consul-helm/issues/848)). 

As per Consul docs [here](https://www.consul.io/docs/agent/options#_disable_host_node_id), `-disable-host-node-id` is `false` by default, which is what #901 had set in `values.yaml` and `templates/client-daemonset.yaml`. 

The changelog for this PR states: 
> This flag was added in 0.30.0 to stop Consul logging an error when a Client pod was force deleted but this error does not affect the pod restarting so removing the flag will have no real effect.

That is incorrect. This was introduced in #901 to allow the override of `disableHostNodeID` to `true` for Kind and other hypervisor types (such as OpenNebula 5.12 launched LXD containers as mentioned [here](https://github.com/hashicorp/consul-helm/issues/848#issuecomment-817046297))

lkysow:
Hi, please see my comment here https://github.com/hashicorp/consul-helm/issues/848#issuecomment-845484423.

> As per Consul docs here, -disable-host-node-id is false by default

The docs are a bit confusing, the relevant section is:

> This defaults to false in Consul prior to version 0.8.5 and in 0.8.5 and later defaults to true

I'm assuming you're running Consul version > 0.8.5 so the default is `true`.

ilhaan:
I am running Consul version > 0.8.5. Thanks for the clarification!

PR #957, opened 2021-05-19
Separate tproxy tests

ishustava:
Changes proposed in this PR:
- Add `-enable-transparent-proxy` flag to the acceptance tests flags which allows you to globally enable or disable tproxy for all tests
- Update all (except a few) connect-related tests to now reach upstreams via kube DNS when transparent proxy is enabled
- Refactor circleci to add commands that we can re-use when running acceptance tests
- Add acceptance tests job with tproxy enabled in parallel to the regular job to be run on every PR
- Enable tproxy on all nightly acceptance tests
- Unfortunately, I had to disable some tests when tproxy is enabled due to either bugs or flakiness (to be fixed later)
   - mesh gateway test with explicit upstreams doesn't work right now because of a bug in consul that is not yet fixed
   - `RestartConsulClients` connect test is flakey and I've disabled it for now to not block this PR and investigate it separately.

How I've tested this PR:
[Acceptance tests](https://app.circleci.com/pipelines/github/hashicorp/consul-helm/3017/workflows/8010f6cd-dec7-4982-aa97-25bfe0a17e52)

- code review
- Do you think nightly should run with tproxy enabled or disabled? I've enabled it because this *should* be the default behavior, but given a few bugs/flakey tests I'm not sure it's the right thing to do. Curious what you think.

Checklist:
- [ ] Bats tests added

### DIFF
--- a/test/acceptance/framework/k8s/deploy.go
+++ b/test/acceptance/framework/k8s/deploy.go
@@ -16,6 +16,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/yaml"
 )
 
+const staticClientName = "static-client"
+
 // Deploy creates a Kubernetes deployment by applying configuration stored at filepath,
 // sets up a cleanup function and waits for the deployment to become available.
 func Deploy(t *testing.T, options *k8s.KubectlOptions, noCleanupOnFailure bool, debugDirectory string, filepath string) {
@@ -75,17 +77,10 @@ func DeployKustomize(t *testing.T, options *k8s.KubectlOptions, noCleanupOnFailu
 // to be "hello world" in a case of success.
 // If expectSuccess is true, it will expect connection to succeed,
 // otherwise it will expect failure due to intentions.
-func CheckStaticServerConnection(
-	t *testing.T,
-	options *k8s.KubectlOptions,
-	expectSuccess bool,
-	deploymentName string,
-	failureMessages []string,
-	curlArgs ...string,
-) {
+func CheckStaticServerConnection(t *testing.T, options *k8s.KubectlOptions, expectSuccess bool, failureMessages []string, curlArgs ...string) {
 	t.Helper()
 
-	CheckStaticServerConnectionMultipleFailureMessages(t, options, expectSuccess, deploymentName, failureMessages, curlArgs...)
+	CheckStaticServerConnectionMultipleFailureMessages(t, options, expectSuccess, failureMessages, curlArgs...)
 }
 
 // CheckStaticServerConnectionMultipleFailureMessages execs into a pod of the deployment given by deploymentName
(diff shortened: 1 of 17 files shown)


test/acceptance/framework/k8s/deploy.go
```diff
@@ -132,26 +120,21 @@ func CheckStaticServerConnectionMultipleFailureMessages(
 
 // CheckStaticServerConnectionSuccessful is just like CheckStaticServerConnection
 // but it always expects a successful connection.
-func CheckStaticServerConnectionSuccessful(t *testing.T, options *k8s.KubectlOptions, deploymentName string, curlArgs ...string) {
+func CheckStaticServerConnectionSuccessful(t *testing.T, options *k8s.KubectlOptions, curlArgs ...string) {
```
ishustava, line 123: Removing deploymentName arg here because it's always the same. Since the function is specific to static-server, I think it's safe to assume deployment name is static-client.

test/acceptance/tests/metrics/metrics_test.go
```diff
@@ -66,7 +66,7 @@ func TestComponentMetrics(t *testing.T) {
 	// This simulates queries that would be made by a prometheus server that runs externally to the consul
 	// components in the cluster.
 	logger.Log(t, "creating static-client")
-	k8s.DeployKustomize(t, ctx.KubectlOptions(t), cfg.NoCleanupOnFailure, cfg.DebugDirectory, "../fixtures/cases/static-client-inject")
+	k8s.DeployKustomize(t, ctx.KubectlOptions(t), cfg.NoCleanupOnFailure, cfg.DebugDirectory, "../fixtures/bases/static-client")
```
ishustava, line 69: We don't need to connect-inject static-client here because it's only used to connect to other pods. With tproxy, it doesn't work because all traffic goes through the proxy and we can't connect to those pods.

kschoche (approved):
Great work!

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
```diff
@@ -36,7 +36,10 @@ func TestMeshGatewayDefault(t *testing.T) {
 		"global.federation.createFederationSecret": "true",
```
ndhanushkodi, line 36: I think this test is supposed to be disabled? I think I am missing where it's disabled and wasn't able to catch that.

ndhanushkodi (approved):
Amazing work on this! I think it makes sense to enable by default the tproxy tests since we'll want to put in the effort to make those tests more robust with pipeline engineer role between now and the release.

test/acceptance/tests/mesh-gateway/mesh_gateway_test.go
ishustava: We're disabling tproxy for these tests below explicitly. Sorry for the inconsistency, I just thought it would still be valuable to run these in the nightly instead of skipping them like we do with the other test.

PR #958, opened 2021-05-19
Re-enable OpenShift Tests

ishustava:
Changes proposed in this PR:
* Create openshift clusters on Azure with terraform
* Enable running tests with tproxy. For that, we need to
  allow all connect-injected apps to use anyuid built-in SCC
  so that Envoy can run as the 5995 user and the privileged built-in
  SCC so that connect-inject-init container can run traffic redirection
  commands
* Run openshift tests in parallel

How I've tested this PR:
[acceptance tests](https://app.circleci.com/pipelines/github/hashicorp/consul-helm/3078/workflows/960bd47a-041c-4cf0-849f-19ca911e02da/jobs/12713)

- code review

Checklist:
- [ ] Bats tests added


templates/mesh-gateway-deployment.yaml
```diff
@@ -225,10 +225,10 @@ spec:
             {{- end }}
           resources:
             requests:
-              memory: "25Mi"
+              memory: "50Mi"
               cpu: "50m"
             limits:
-              memory: "25Mi"
+              memory: "50Mi"
```
ishustava, line 231: These limits were already increased for other gateways, and I've seen it fail due to OOM issues on openshift

kschoche (approved):
sweet!

test/acceptance/framework/consul/consul_cluster.go
```diff
@@ -459,6 +463,52 @@ func createOrUpdateLicenseSecret(t *testing.T, client kubernetes.Interface, cfg
 	})
 }
 
+// configureSCCs creates RoleBindings that bind the default service account to cluster roles
+// allowing access tot he anyuid and privileged Security Context Constraints on OpenShift.
```
ndhanushkodi, line 467:
```suggestion
// allowing access to the anyuid and privileged Security Context Constraints on OpenShift.
```

test/acceptance/framework/consul/consul_cluster.go
```diff
@@ -459,6 +463,52 @@ func createOrUpdateLicenseSecret(t *testing.T, client kubernetes.Interface, cfg
...
+func configureSCCs(t *testing.T, client kubernetes.Interface, cfg *config.TestConfig, namespace string) {
+	anyuidClusterRole := "system:openshift:scc:anyuid"
+	privilegedClusterRole := "system:openshift:scc:privileged"
+	anyuidRoleBinding := "anyuid-test"
+	privilegedRoleBinding := "privileged-test"
```
ndhanushkodi, line 472: Does it make sense for these to be `const`?

test/acceptance/framework/consul/consul_cluster.go
```diff
@@ -459,6 +463,52 @@ func createOrUpdateLicenseSecret(t *testing.T, client kubernetes.Interface, cfg
 	})
 }
 
+// configureSCCs creates RoleBindings that bind the default service account to cluster roles
+// allowing access tot he anyuid and privileged Security Context Constraints on OpenShift.
+func configureSCCs(t *testing.T, client kubernetes.Interface, cfg *config.TestConfig, namespace string) {
+	anyuidClusterRole := "system:openshift:scc:anyuid"
+	privilegedClusterRole := "system:openshift:scc:privileged"
+	anyuidRoleBinding := "anyuid-test"
```
ndhanushkodi, line 471: This rolebinding is for the consul system components that use the default service account, right? could it make sense to name this `consul-system-anyuid` or `system-anyuid`? I was mixing up this role binding and the ones for static-server/client for a moment while looking at it.

test/terraform/openshift/oc-login.sh
```diff
@@ -0,0 +1,19 @@
...
+
+echo "Logging in"
+for i in {1..20}; do KUBECONFIG="$HOME/.kube/$cluster_name" oc login "$apiServer" -u "$kubeUser" -p "$kubePassword" && break; sleep 5; done
+
+echo "Creating the 'consul' project"
+# Idempotently, create and use the 'consul' project
+KUBECONFIG="$HOME/.kube/$cluster_name" oc new-project consul
+KUBECONFIG="$HOME/.kube/$cluster_name" oc project consul
+
+echo "Disabling auto-update of the worker nodes based on config changes"
```
ndhanushkodi, line 18: was this what was causing the flakes earlier?

ndhanushkodi (approved):
Looks amazing!

test/acceptance/framework/consul/consul_cluster.go
ishustava: This would be for any test pods that aren't setting the service account (currently I think it's just the test-dns pod we use in the consul-dns test). All consul components have their own dedicated service accounts.

test/terraform/openshift/oc-login.sh
ishustava: Yep exactly! Openshift has an operator that could reboot worker nodes if the config needs to be updated. As a result, some nodes were updating during test runs causing failures.

test/acceptance/framework/consul/consul_cluster.go
ishustava: I made the first two consts since they are the specific names that openshift creates. The latter two can be anything we want, so I'll keep them as vars.

PR #960, opened 2021-05-20
acceptance-tests: check pod status and conditions in WaitForAllPodsToBeReady instead of container statutes

ishustava:
Sometimes, it seems like `WaitForAllPodsToBeReady` exits before the pods are actually ready. For example, in [this test's artifacts](https://app.circleci.com/pipelines/github/hashicorp/consul-helm/3006/workflows/9ae2156b-f41f-40d2-af61-5b3e9b31e35a/jobs/12114/artifacts), we can see that ingress and mesh gateway pods are not ready yet but in the output of the test it shows that `WaitForAllPodsToBeReady` exited after 1sec. I think it's because previously, we were only checking container statutes, and there are transition states where the status of the container may be Ready, but the pod itself is not yet ready.

Changes proposed in this PR:
- Check the pod phase and status instead of individual containers
- Also, use clusterIP service in the component metrics test for the mesh gateway instead of the load balancer (default) since we don't need mesh gw to be reachable in this test and some clouds take a long time to provision a load balancer.

How I've tested this PR:
- acceptance tests

code review

Checklist:
- [ ] Bats tests added

### DIFF
--- a/test/acceptance/framework/helpers/helpers.go
+++ b/test/acceptance/framework/helpers/helpers.go
@@ -128,19 +128,19 @@ func KubernetesContextFromOptions(t *testing.T, options *terratestk8s.KubectlOpt
 
 // IsReady returns true if pod is ready.
 func IsReady(pod corev1.Pod) bool {
-	if len(pod.Status.ContainerStatuses) == 0 {
+	if pod.Status.Phase == corev1.PodPending {
 		return false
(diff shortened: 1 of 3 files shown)


kschoche (approved):
Nice find here!

lkysow (approved):

PR #961, opened 2021-05-20
create an acl replication token whenever federation is enabled

kschoche:
Changes proposed in this PR:
- add `-create-acl-replication-token=true` to server-acl-init whenever federation is enabled so that we can use it in consul-k8s to know when federation is enabled.

How I've tested this PR:
unit tests + acceptance tests

code review + unit test

Checklist:
- [x] Bats tests added

### DIFF
--- a/templates/server-acl-init-job.yaml
+++ b/templates/server-acl-init-job.yaml
@@ -195,7 +195,7 @@ spec:
                 -create-client-token=false \
                 {{- end }}
 
-                {{- if (or .Values.global.federation.enabled .Values.global.acls.createReplicationToken) }}
+                {{- if (or .Values.global.acls.createReplicationToken (and .Values.global.federation.enabled (and .Values.connectInject.enabled (and .Values.global.tls.enabled .Values.meshGateway.enabled)))) }}
                 -create-acl-replication-token=true \
                 {{- end }}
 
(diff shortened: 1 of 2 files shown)


templates/server-acl-init-job.yaml
```diff
@@ -195,7 +195,7 @@ spec:
                 -create-client-token=false \
                 {{- end }}
 
-                {{- if .Values.global.acls.createReplicationToken }}
+                {{- if (or .Values.global.acls.createReplicationToken (and .Values.global.federation.enabled (and .Values.connectInject.enabled (and .Values.global.tls.enabled .Values.meshGateway.enabled)))) }}
```
thisisnotashwin, line 198: I learnt recently that these `and` statements can be flattened.
```suggestion
                {{- if (or .Values.global.acls.createReplicationToken (and .Values.global.federation.enabled .Values.connectInject.enabled .Values.global.tls.enabled .Values.meshGateway.enabled)) }}
```

thisisnotashwin: For my understanding, is it not enough to only check `.Values.global.federation.enabled` is enabled?

thisisnotashwin (commented):
This looks really good. Have a clarifying question but that's about it.

kschoche:
Closing this because it's addressed via https://github.com/hashicorp/consul-helm/pull/983

PR #236, opened 2019-09-26
Allows to add annotations on the client service account ressource

kri5:
Regarding the issue #235 , This PR allows users to provide annotations for the client service account (It seems to be the only relevant place to support this, at least for now).

If you have any comments, feel free!

PR #964, opened 2021-05-21
Allow setting annotations on service accounts

### LINKED ISSUE #235: Support for annotations on ServiceAccounts
In order to be able to use feature such as [GKE Workload Identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity), it would be great to allow user to specify annotations for the various services accounts supported by this chart.

One use case would be, you want to deploy the clients on k8s, to out-of-cluster servers, using cloud auto-join discovery (gce here), and you need to have permission to do so. Annotations would allow to "inject" the google credentials to the kubernetes services account and allow the clients to discover the servers.

cooperbenson-qz:
Fixes #235.

Changes proposed in this PR:
This is a redux of #236

- Added values for service account annotations
- These values expect a multi-line YAML string of the annotations map

How I've tested this PR:
- Unit tests and some test renders

- Same

Checklist:
- [x] Bats tests added

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -1787,6 +1787,19 @@ meshGateway:
   # @type: integer
   hostPort: null
 
+  serviceAccount:
+    # This value defines additional annotations for the mesh gateways' service account. This should be formatted as a
+    # multi-line string.
+    #
+    # ```yaml
+    # annotations: |
+    #   "sample/annotation1": "foo"
+    #   "sample/annotation2": "bar"
+    # ```
+    #
+    # @type: string
+    annotations: null
+
   # Resource settings for mesh gateway pods.
   # NOTE: The use of a YAML string is deprecated. Instead, set directly as a
   # YAML map.
--- a/values.yaml
+++ b/values.yaml
@@ -1416,6 +1416,19 @@ connectInject:
   # Log verbosity level. One of "debug", "info", "warn", or "error".
   logLevel: info
 
+  serviceAccount:
+    # This value defines additional annotations for the client service account. This should be formatted as a multi-line
+    # string.
+    #
+    # ```yaml
+    # annotations: |
+    #   "sample/annotation1": "foo"
+    #   "sample/annotation2": "bar"
+    # ```
+    #
+    # @type: string
+    annotations: null
+
   # Resource settings for connect inject pods.
   # @recurse: false
   # @type: map
--- a/values.yaml
+++ b/values.yaml
@@ -371,8 +371,9 @@ server:
   # by setting the `server.extraConfig` value.
   connect: true
 
-  # A map of annotations to add to the server service account
-  serviceAccountAnnotations: {}
+  serviceAccount:
+    # A map of annotations to add to the server service account 
+    annotations: {}
 
   # The resource requests (CPU, memory, etc.)
   # for each of the server agents. This should be a YAML map corresponding to a Kubernetes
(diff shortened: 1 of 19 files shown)


cooperbenson-qz:
Here's the full context on why I decided to limit the service accounts I added this feature for. I started adding annotations to all of the service accounts but I ran into issues with two which lead me to rethink the scope:
- [templates/connect-inject-authmethod-serviceaccount.yaml](https://github.com/hashicorp/consul-helm/blob/master/templates/connect-inject-authmethod-serviceaccount.yaml) - I couldn't work out what this was being used for and so was having a hard time working out where in the values structure it should go
- [templates/create-federation-secret-serviceaccount.yaml](https://github.com/hashicorp/consul-helm/blob/master/templates/create-federation-secret-serviceaccount.yaml) - The job is currently configured with a boolean `global.federation.createFederationSecret`. I didn't want to break the values API by changing that to be `global.federation.createFederationSecret.enabled` and adding `global.federation.createFederationSecret.serviceAccount.annotations`. The alternative would be to add `global.federation.createFederationSecretServiceAccountAnnotations`, but that felt like a bad design choice. This will also likely be a problem for all of the other bootsrapping jobs (ACL and TLS in particular).

values.yaml
```diff
@@ -371,6 +371,10 @@ server:
   # by setting the `server.extraConfig` value.
   connect: true
 
+  serviceAccount:
+    # A map of annotations to add to the server service account 
+    annotations: {}
```
ndhanushkodi, line 376: @cooperbenson-qz Are you able to format them as multiline strings for consistency with https://github.com/hashicorp/consul-helm/blob/master/values.yaml#L584?

ndhanushkodi (changes requested):
Thanks so much for following up on this @cooperbenson-qz! Can you go ahead and sign the CLA, and could we use the same format as other annotations in the helm chart for consistency?

values.yaml
cooperbenson-qz: Oh I missed that was the pattern already in use. I'm used to having a map that uses `toYaml`, but I can use the established pattern

cooperbenson-qz:
I'm still waiting on our legal team to review the CLA so I'll work on adding more annotations

cooperbenson-qz:
Added annotations for all service accounts except for the following. All of them lack a suitable area in the values file to add the option to. I can add them if I can get some help determining the correct place to add values for them. 

- templates/connect-inject-authmethod-serviceaccount.yaml
- templates/webhook-cert-manager-serviceaccount.yaml
- Jobs:
  - templates/create-federation-secret-serviceaccount.yaml
  - templates/enterprise-license-serviceaccount.yaml
  - templates/server-acl-init-cleanup-serviceaccount.yaml
  - templates/server-acl-init-serviceaccount.yaml
  - templates/tls-init-cleanup-serviceaccount.yaml
  - templates/tls-init-serviceaccount.yaml

cooperbenson-qz:
@ndhanushkodi mind taking another look now that the CLA is signed? Really eager to get this merged 😃

david-yu:
Hi @cooperbenson-qz thank you for signing the CLA and the updating the PR[ We are going to do a release shortly for our 1.10 beta3 so we will wait to review this until after the release is done this week. Stay tuned.

cooperbenson-qz:
Just wanted to check in since it's been a bit. Still would love to see this merged soon 🙂

ndhanushkodi (approved):
Thank you for making the changes! It looks great!! Sorry for the late review

kschoche (approved):
Thanks for adding this, it looks great!

cooperbenson-qz:
Thanks for the reviews 😄  When you merge, it would be great if you could also close #236

PR #974, opened 2021-05-25
Support adding enterprise license as an ENV VAR and not a secret

thisisnotashwin:
Changes proposed in this PR:
- Use the environment variable for the Consul License to create a license secret which the helm chart will autoload.
- This creates a kubernetes secret in each cluster where a helm install is performed. It also deprecates the flags for license-secret-name and license-secret-key for the acceptance tests.
- The tests now rely on the existence of the the env var CONSUL_ENT_LICENSE and if present, creates a secret with the value of the env var, and then updates the helm config with the secret name and secret key.

How I've tested this PR:
Ran acceptance tests against CI

Code review

Checklist:
- [ ] Bats tests added

### DIFF
--- a/test/acceptance/framework/config/config.go
+++ b/test/acceptance/framework/config/config.go
@@ -13,7 +13,11 @@ import (
 
 // HelmChartPath is the path to the helm chart.
 // Note: this will need to be changed if this file is moved.
-const HelmChartPath = "../../../.."
+const (
+	HelmChartPath     = "../../../.."
+	LicenseSecretName = "license"
+	LicenseSecretKey  = "key"
+)
 
 // TestConfig holds configuration for the test suite.
 type TestConfig struct {
@@ -26,9 +30,8 @@ type TestConfig struct {
 	SecondaryKubeContext   string
 	SecondaryKubeNamespace string
 
-	EnableEnterprise            bool
-	EnterpriseLicenseSecretName string
-	EnterpriseLicenseSecretKey  string
+	EnableEnterprise  bool
+	EnterpriseLicense string
 
 	EnableOpenshift bool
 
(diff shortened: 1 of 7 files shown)


test/acceptance/framework/flags/flags.go
```diff
@@ -93,6 +89,8 @@ func (t *TestFlags) init() {
 
 	flag.BoolVar(&t.flagUseKind, "use-kind", false,
 		"If true, the tests will assume they are running against a local kind cluster(s).")
+
+	t.flagEnterpriseLicense = os.Getenv("CONSUL_ENT_LICENSE")
```
ishustava, line 93: should we also make it possible to set via a flag?

ishustava (approved):
Looks good! Just one minor comment, but non-blocking. Also, would you mind updating the Contributing docs since our flags have changed and I think we have a list of all flags there?

test/acceptance/framework/flags/flags.go
thisisnotashwin: Done ✅

test/acceptance/framework/config/config.go
```diff
@@ -62,9 +65,9 @@ func (t *TestConfig) HelmValuesFromConfig() (map[string]string, error) {
 		setIfNotEmpty(helmValues, "global.image", entImage)
 	}
 
-	if t.EnterpriseLicenseSecretName != "" && t.EnterpriseLicenseSecretKey != "" {
-		setIfNotEmpty(helmValues, "server.enterpriseLicense.secretName", t.EnterpriseLicenseSecretName)
-		setIfNotEmpty(helmValues, "server.enterpriseLicense.secretKey", t.EnterpriseLicenseSecretKey)
+	if t.EnterpriseLicense != "" {
```
kschoche, line 68: Meta comment, no changes needed: I feel like since `EnterpriseLicense` && `EnableEnterprise` are dependent on eachother at this point it'd be great to consolidate them?

kschoche: Oh actually I see that you can set `EnableEnterprise` w/o setting `EnterpriseLicense` and then you pick from the ENV.. nevermind!

kschoche (approved):
Awesome work!

PR #975, opened 2021-05-26
Allow user to explicitely set advertise address for Consul server

jacksgt:
*Problem description*
By default, the Helm chart only allows choosing between the PodIP
and HostIP for the advertise address that is passed to Consul.
In certain configuration it is necessary to specify a different IP
(NAT, VMs etc.). Since the environment key "ADVERTISE_IP" is already
set by the Helm chart, the user can no longer simply overwrite this.
(Kubernetes will complain that `valueRef` cannot be used when `value` is
specified)

*Changes proposed in this PR:*
Thus, this patch introduces an explicit option for this configuration
setting.

*How I've tested this PR:*
I locally deployed the Helm chart with these changes. I also added a testcase for it.

Checklist:
- [x] Bats tests added

### DIFF
--- a/templates/server-statefulset.yaml
+++ b/templates/server-statefulset.yaml
@@ -121,6 +121,9 @@ spec:
           image: "{{ default .Values.global.image .Values.server.image }}"
           env:
             - name: ADVERTISE_IP
+              {{- if .Values.server.advertiseAddr }}
+              value: {{ .Values.server.advertiseAddr }}
+              {{- else }}
               valueFrom:
                 fieldRef:
                   {{- if .Values.server.exposeGossipAndRPCPorts }}
@@ -132,6 +135,7 @@ spec:
                   {{- else }}
                   fieldPath: status.podIP
                   {{- end }}
+              {{- end }}
             - name: POD_IP
               valueFrom:
                 fieldRef:
--- a/test/unit/server-statefulset.bats
+++ b/test/unit/server-statefulset.bats
@@ -232,6 +232,21 @@ load _helpers
   [ "${actual}" = '[{"name":"ADVERTISE_IP","valueFrom":{"fieldRef":{"fieldPath":"status.hostIP"}}}]' ]
 
 }
+#--------------------------------------------------------------------
+# advertiseAddr
+
+@test "server/StatefulSet: advertiseAddr can be explicitely set" {
+  cd `chart_dir`
+  local object=$(helm template \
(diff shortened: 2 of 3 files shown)


jacksgt:
I also tried the following workarounds:

```yaml
server:
 extraConfig: |
   {
     "advertise_addr": "1.2.3.4",
   }
```
=> doesn't work because the command-line parameter takes precedence over configuration file values.
https://github.com/hashicorp/consul-helm/blob/44fbdea61340f533f8bcd7c31571975436cb23d9/templates/server-statefulset.yaml#L170-L172

----------------

```yaml
server:
  extraEnvironmentVars:
   ADVERTISE_IP: "1.2.3.4"
```
=> doesn't work because Kubernetes will complain ADVERTISE_IP has both a `value` and `valueFrom`:
https://github.com/hashicorp/consul-helm/blob/44fbdea61340f533f8bcd7c31571975436cb23d9/templates/server-statefulset.yaml#L123-L125

jacksgt:
Could anyone tell me why the unit test failed? It seems like I don't have access to CircleCI.

jacksgt:
Ping @lkysow @kschoche @ndhanushkodi

lkysow:
```
not ok 912 server/StatefulSet: advertiseAddr can be explicitely set
# (in test file test/unit/server-statefulset.bats, line 248)
#   `[ "${actual}" = "1.2.3.4" ]' failed
# [{"name":"ADVERTISE_IP","value":"1.2.3.4"}]
```

lkysow:
Hi, I'm curious what you would be setting this to? This would apply to all servers so if you used an IP like 1.2.3.4 then all servers would advertise that which probably wouldn't make sense.

jacksgt:
> Hi, I'm curious what you would be setting this to? This would apply to all servers so if you used an IP like 1.2.3.4 then all servers would advertise that which probably wouldn't make sense.

Basically I'm running a Consul server inside a VM (with Kubernetes) which is NATed from the Hypervisor
So I have Hypervisor IP -> VM IP -> Consul Pod IP.
By enabling the `exposeGossipAndRPCPorts` options I could get consul to use the Host (VM) IP, but that one is still not reachable by the clients (other physical hosts). Thus, I manually need to set the `advertise_addr` to the Hypervisor IP.

It's definitely not the most ideal environment to run consul in, but that's what we have.

t-eckert:

PR #980, opened 2021-06-07
Add ability to set log level and enable json logging for all consul-k8s commands

kschoche:
Changes proposed in this PR:
- Add ability to set json logging for consul-k8s components : `global.logJSON`
- Add ability to control log level for all consul-k8s components  `global.logLevel` 
 
NOTES:
* Envoy + consul server/client are not currently implemented.

How I've tested this PR:
unit tests, acceptance tests pass

code review, acceptance tests pass
Also have manually testing using steps in https://github.com/hashicorp/consul-k8s/pull/523

Checklist:
- [x] Bats tests added

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ orbs:
 executors:
   go:
     docker:
-      - image: docker.mirror.hashicorp.services/circleci/golang:1.14
+      - image: docker.mirror.hashicorp.services/circleci/golang:1.16
     environment:
       - TEST_RESULTS: /tmp/test-results
 
@@ -14,7 +14,14 @@ commands:
       - run:
           name: Install gotestsum, kind, kubectl, and helm
           command: |
-            go get gotest.tools/gotestsum
+            wget https://golang.org/dl/go1.16.5.linux-amd64.tar.gz
+            sudo rm -rf /usr/local/go && sudo tar -C /usr/local -xzf go1.16.5.linux-amd64.tar.gz
+            rm go1.16.5.linux-amd64.tar.gz
+            echo 'export PATH=$PATH:/usr/local/go/bin' >> $BASH_ENV
+
+            wget https://github.com/gotestyourself/gotestsum/releases/download/v1.6.4/gotestsum_1.6.4_linux_amd64.tar.gz
+            sudo tar -C /usr/local/bin -xzf gotestsum_1.6.4_linux_amd64.tar.gz
+            rm gotestsum_1.6.4_linux_amd64.tar.gz
 
             curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.11.0/kind-linux-amd64
             chmod +x ./kind
@@ -101,7 +108,7 @@ commands:
 jobs:
   unit-helm:
     docker:
-      - image: docker.mirror.hashicorp.services/hashicorpdev/consul-helm-test:0.9.1
+      - image: docker.mirror.hashicorp.services/hashicorpdev/consul-helm-test:0.10.0
 
     steps:
       - checkout
@@ -134,6 +141,7 @@ jobs:
       # check go fmt output because it does not report non-zero when there are fmt changes
       - run:
           name: check go fmt
+          working_directory: test/acceptance
           command: |
             files=$(go fmt ./...)
             if [ -n "$files" ]; then
@@ -239,7 +247,7 @@ jobs:
       - TEST_RESULTS: /tmp/test-results
     docker:
       # This image is built from test/docker/Test.dockerfile
-      - image: docker.mirror.hashicorp.services/hashicorpdev/consul-helm-test:0.9.0
+      - image: docker.mirror.hashicorp.services/hashicorpdev/consul-helm-test:0.10.0
 
     steps:
       - run:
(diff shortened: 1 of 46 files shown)


.circleci/config.yml
```diff
@@ -48,7 +48,7 @@ commands:
         type: string
       consul-k8s-image:
         type: string
-        default: "docker.mirror.hashicorp.services/hashicorpdev/consul-k8s:latest"
+        default: "kschoche/consul-k8s-dev-logging"
```
kschoche, line 51: This will be switched prior to merging

templates/connect-inject-deployment.yaml
```diff
@@ -83,6 +83,11 @@ spec:
               CONSUL_FULLNAME="{{template "consul.fullname" . }}"
 
               consul-k8s inject-connect \
+                {{- if .Values.connectInject.logLevel }}
+                -log-level={{ .Values.connectInject.logLevel }} \
+                {{- else }}
+                -log-level={{ .Values.global.logLevel }} \
+                {{- end }}
```
lkysow, line 90:
```suggestion
                -log-level={{ default .Values.global.logLevel .Values.connectInject.logLevel }} \
```
I think we can use `default` for anywhere we're doing this style

lkysow: `-log-json` isn't passed

templates/controller-deployment.yaml
```diff
@@ -59,10 +59,14 @@ spec:
...
         - |
           consul-k8s controller \
+            {{- if .Values.controller.logLevel }}
+            -log-level={{ .Values.controller.logLevel }} \
+            {{- else }}
+            -log-level={{ .Values.global.logLevel }} \
+            {{- end }}
             -webhook-tls-cert-dir=/tmp/controller-webhook/certs \
             -datacenter={{ .Values.global.datacenter }} \
             -enable-leader-election \
```
lkysow, line 69: `-log-json` isn't passed

templates/create-federation-secret-job.yaml
```diff
@@ -116,17 +116,19 @@ spec:
...
+                -gossip-key-file=/consul/gossip/gossip.key \
+                {{- end }}
+                {{- if .Values.global.acls.createReplicationToken }}
+                -export-replication-token=true \
+                {{- end }}
+                -mesh-gateway-service-name={{ .Values.meshGateway.consulServiceName }} \
+                -k8s-namespace="${NAMESPACE}" \
+                -resource-prefix="{{ template "consul.fullname" . }}" \
+                -server-ca-cert-file=/consul/tls/ca/tls.crt \
+                -server-ca-key-file=/consul/tls/server/ca/tls.key
```
lkysow, line 131: this was un-indented

templates/ingress-gateways-deployment.yaml
```diff
@@ -368,11 +370,11 @@ spec:
...
-                      /consul-bin/consul services deregister \
-                      {{- if $root.Values.global.enableConsulNamespaces }}
-                      -namespace={{ default $defaults.consulNamespace .consulNamespace }} \
-                      {{- end }}
-                      -id="${POD_NAME}"
+                    /consul-bin/consul services deregister \
+                    {{- if $root.Values.global.enableConsulNamespaces }}
+                    -namespace={{ default $defaults.consulNamespace .consulNamespace }} \
+                    {{- end }}
+                    -id="${POD_NAME}"
```
lkysow, line 377: should be indented to match style elsewhere. I think we could do with a full YAML format maybe to standardize but should be a sep PR

templates/ingress-gateways-deployment.yaml
```diff
@@ -417,13 +419,17 @@ spec:
...
           command:
-            - consul-k8s
-            - consul-sidecar
-            - -service-config=/consul/service/service.hcl
-            - -consul-binary=/consul-bin/consul
-            {{- if $root.Values.global.acls.manageSystemACLs }}
-            - -token-file=/consul/service/acl-token
-            {{- end }}
+            - "/bin/sh"
+            - "-ec"
```
lkysow, line 423: should we leave this as before? One thing with running in `/bin/sh` is you don't get passed signals when the pod stops. To get around this we could use `exec` but maybe in a sep PR we could standardize

values.yaml
```diff
@@ -1331,7 +1341,8 @@ syncCatalog:
       cpu: "50m"
 
   # Log verbosity level. One of "trace", "debug", "info", "warn", or "error".
```
lkysow, line 1343: Should we say something like "Override `global.logLevel`" for all of these overrides?

lkysow (commented):
Looks god. One question about the move to `sh -ec`

ndhanushkodi (commented):
For a command like consul-sidecar for connect-injected pods that doesn't have its logLevel/logJSON set here, do you need to add something to the consul-k8s PR to invoke that internal command with the right flag based on the global value?

templates/controller-deployment.yaml
kschoche: Ah good catch, actually this was intentional because it wasn't initially scoped in the PR, but I've since added that so it definitely needs to go in to both controller and inject-connect!

templates/ingress-gateways-deployment.yaml
kschoche: Good call, reverting!

ndhanushkodi (approved):
Nice work!!

lkysow (approved):

lkysow:
Should we have a follow-up PR that sets the log level/json for Consul clients/servers too?

kschoche:
> Should we have a follow-up PR that sets the log level/json for Consul clients/servers too?

Yup this was only for subcommands.

PR #982, opened 2021-06-08
Update type of connectTimeout and TTL

thisisnotashwin:
Update duration types from int64 to string to allow users to set the duration as `5s` for example.

How I've tested this PR: Created a resource with the following fields
```
kind: ServiceResolver
metadata:
  name: resolver
spec:
  redirect:
    service: bar
  connectTimeout: 5s
  loadBalancer:
    policy: policy
    hashPolicies:
    - cookieConfig:
        ttl: 3s
```
It gets applied successfully. The config does not sync with consul but that is because the loadBalancer config here has some error with the hashPolicy but the type are correct.

How I expect reviewers to test this PR: Reference https://github.com/hashicorp/consul-k8s/pull/529 and check changelog.

Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## UNRELEASED
+
+BUG FIXES:
+* CRDs: Update the type of connectTimeout and TTL in ServiceResolver and ServiceRouter from int64 to string.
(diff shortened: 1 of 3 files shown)


kschoche (approved):
🚀

lkysow (approved):

PR #991, opened 2021-06-15
Update acceptance test fixture with startup probe

thisisnotashwin:
- This should test that the probe overwrite with tproxy enabled works
  for the startup probe as well

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: consul
 version: 0.32.0-beta3
-appVersion: 1.10.0-beta3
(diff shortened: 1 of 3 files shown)


ndhanushkodi (approved):
great idea!

ishustava (approved):

thisisnotashwin:
This test is currently failing acceptance because consul currently fails if 2 expose paths share the same path which happens in this case between the startup probe and liveness probe and hence the reconcile fails. this bug is fixed in consul now, but this test should succeed when a new consul image is available. I will merge this PR in once we have a consul 1.10 GA image.

PR #996, opened 2021-06-18
Cleanup acceptance tests leftover resources on Azure and GCP

ishustava:
Changes proposed in this PR:
- Cleanup leftover resources on Azure and GCP:
   - On azure, since all resources are scoped to a resource group, we can just delete the resource groups that are created by tests, and azure will recursively delete any resources belonging to that group as well.
   - On GCP, we only need to delete the GKE clusters since no additional resources are created.
- Successful job on Azure: https://app.circleci.com/pipelines/github/hashicorp/consul-helm/3200/workflows/8b17310d-27fb-4069-9a5b-7800c66442b2/jobs/13607
- Successful job on GCP: https://app.circleci.com/pipelines/github/hashicorp/consul-helm/3199/workflows/b7ae9323-243c-4532-bab8-1e4af8a834de/jobs/13600

How I've tested this PR:
Ran these jobs in CI

code review

Checklist:
- [ ] Bats tests added

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -351,9 +351,9 @@ jobs:
             terraform destroy -auto-approve
           when: always
 
-      - slack/status:
-          fail_only: true
-          failure_message: "AKS acceptance tests failed. Check the logs at: ${CIRCLE_BUILD_URL}"
+#      - slack/status:
+#          fail_only: true
+#          failure_message: "AKS acceptance tests failed. Check the logs at: ${CIRCLE_BUILD_URL}"
 
   acceptance-eks-1-18:
     environment:
@@ -469,9 +469,9 @@ jobs:
           command: |
             terraform destroy -auto-approve
           when: always
-      - slack/status:
-          fail_only: true
-          failure_message: "OpenShift acceptance tests failed. Check the logs at: ${CIRCLE_BUILD_URL}"
+#      - slack/status:
+#          fail_only: true
+#          failure_message: "OpenShift acceptance tests failed. Check the logs at: ${CIRCLE_BUILD_URL}"
 
   acceptance-kind-1-21:
     environment:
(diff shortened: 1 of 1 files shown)


thisisnotashwin (approved):
NICE!!

.circleci/config.yml
```diff
@@ -648,6 +676,13 @@ workflows:
       - acceptance-eks-1-18
       - acceptance-aks-1-19
       - acceptance-kind-1-21
+      - cleanup-azure-resources:
+          requires:
+            - acceptance-aks-1-19
+            - acceptance-openshift
```
lkysow, line 682: I think you need to run the cleanup as a step as part of the actual `acceptance-aks-1-19` or `acceptance-gke-1-17` jobs because as written these jobs will only run if the tests succeed but we want the cleanup to always run. It doesn't look like there's a way to say "run this after this even if it failed" in the workflows config.

Unfortunately we would then still run into the issue of the destroy job being stopped during its run.

lkysow: https://discuss.circleci.com/t/how-to-always-run-a-sequential-final-job-e-g-slack-notification/35362

ishustava: Yeah I was thinking about that too, but unfortunately, since openshift tests are running in parallel, I can't make it a step in the job because then one parallel run could delete resources from another run. Moreover, this could happen for other test jobs if we make them parallel too. So I think this is the "best effort" type of job that will run only if tests succeed. Another option would be to make it a separate workflow, but then we'd have to make sure to time it correctly so that it doesn't overlap with existing tests running.

lkysow: What if it runs *before* the jobs?

ishustava: I think that will still cause race conditions, because you'd have two processes pulling up the same list of resources and trying to delete them simultaneously. I could try though and see if that works.

ishustava: Sorry that I've misunderstood you. Updated to run the cleanup jobs before the test jobs!

lkysow (approved):

PR #804, opened 2021-02-03
add several hostPort customizations

iameli:
I'm not sure yet which of these should be fully-customizable and which should just be customizable by hostPort, you know?

While I'm here, should I just make every port customizable across the board?

Changes proposed in this PR:
- Add the ability to customize several ports, enabling the ability to 
- Fixes https://github.com/hashicorp/consul-helm/issues/803

How I've tested this PR:

* Used it to install multiple Consuls on the same cluster

How I expect reviewers to test this PR:

* Use it to install multiple Consuls on the same cluster

Checklist:
- [ ] Bats tests added
- [ ] CHANGELOG entry added (*HashiCorp engineers only, community PRs should not add a changelog entry*)

Issue #803, opened 2021-02-03
add ability to run more than one consul release per cluster


iameli:
#### Is your feature request related to a problem? Please describe.

Installing this chart multiple times on the same cluster results in the second deployment having lots and lots of hostPort conflicts.

#### Feature Description

All of the hostPorts should be customizable, so you can run more than one Consul per Kubernetes cluster if you care to.

#### Use Case(s)

I want to run a "staging Consul" and a "production Consul" on the same nodes.

#### Contributions

Yep, I've already started — will send a PR with my preliminary work once I get something running.

Mostly filing this in case there's a way to do this that I'm not seeing.

ndhanushkodi:
Hey @iameli, this seems similar to #126 and #597. So in addition to the hostPort conflicts in consul-helm, there's also lots of automation that's done in consul-k8s that requires specific ports and doesn't support multiple installations on the same cluster. I am not sure if this is something we'd want to support in the long term, since most production clusters would ideally have just one consul installation. We do really appreciate you wanting to contribute the change, but can we see if there's more feedback that this is a use case that a lot of people need before deciding to support it in the long term?

iameli:
Got it. Let me explain a a little about our use case, perhaps there's a smarter way to accomplish what I'm trying to do. Livepeer is building a global network of high-availability RTMP servers to provide video ingest into our decentralized transcoding network. Toward that end, we've bought or leased bare-metal servers all over the world, which we've tied together in to Kubernetes clusters with k3s.

How we're currently using Consul: when livestreams come in, the RTMP server that handles the incoming TCP connection writes an entry to Consul KV. Incoming HTTPS connections hit Traefik, and Traefik uses that data to route HTTPS playback to the appropriate origin server presently handling that connection. This is how we provide load distribution and origin discovery, and it's working great.

How we'd like to use Consul: If I federate all of our currently-independent Consul servers, and muck around with consul-replicate, I can push that same KV entry into all of our datacenters. This will allow us to have a single playback URL, `cdn.livepeer.com`, that can route traffic appropriately from each of the datacenters. Neat! But, two problems:

[...]

Issue #597, opened 2020-09-08
port conflict when more than one consul deployed on a kubernetes cluster


nixilin:
I have 2 namespaces and want to install 2 consuls for each namespace. But all agents' ready state on second consul is 0/1.
I described related pod and see the following message
0/5 nodes are available: 5 node(s) didn't have free ports for the requested pod ports.

So I guess error reason is about port conflict
Because agents' 8500 port will listen on all kubernetes worknode by using daemonset.
change 8500 to other port may cause other unexpected problem?
The only way to solve this is disable agent . But I dont think is a good idea.
So is there any other way to solve this?
Thanks~

ishustava:
Hey @nixilin,

This was already reported in #126.

Yes, unfortunately, we don't support installing the consul helm chart into two Kubernetes namespaces on the same cluster.

PR #999, opened 2021-06-21
Make all Consul client ports customizable

### LINKED ISSUE #997: Allow Consul client daemonset connectivity via nodePort (instead of only hostPort)
#### Feature Description

Currently the Consul deployment on K8s relies on Kubernetes `hostPorts` to provide connectivity to the Consul clients as well as ensuring each app pod talks to its "K8s worker-local" Consul client. The design decision is well described in the Consul docs below:

* https://www.consul.io/docs/k8s/installation/install#client-agents

This causes various problems in organizations and environments where `hostPort` is not available, for example when running TKGI (previously known as PKS) < 1.11 or simply when security requirements do not allow you to use it.

Therefore we (@flavioaiello, @pree & @karras) investigated various scenarios how the Consul deployment could be adjusted to avoid the `hostPort` option. This would not only solve the problem but also improve security related requirements in certain environments.

We managed to successfully make the switch to `nodePorts` via 2-3 additional modifications:

* Adjust the K8s `--service-node-port-range` option to include the ports `8501` and `8502` (e.g. `8500-32767`)
* Activate the `ServiceTopology` feature gate (or `TopologyAwareHints` on K8s 1.21+)
* Create a `nodePort` service for the Consul clients with `topologyKeys` configured to avoid traffic loadbalancing and route everything to the local worker node (or `EndpointSlice` on K8s 1.21+)

**Note:** The adjustment of the `service-node-port-range` is only required due to the [hardcoded ports in consul-k8s](https://github.com/hashicorp/consul-k8s/blob/08cc8b62a359af12641bfa0bbcfc6d2473df6fe2/connect-inject/container_init.go#L264) and consul-helm. We'll track this in separate PRs. See also: https://github.com/hashicorp/consul-k8s/issues/544

Most likely making `nodePorts` the default deployment option would be desirable but I guess it's not possible because it would mean that every organization would need to enable the K8s feature gates. These are still in an alpha state and also disabled by default in vanilla K8s.

Additional references:

* https://kubernetes.io/docs/concepts/services-networking/service-topology/#only-node-local-endpoints
* https://kubernetes.io/docs/concepts/services-networking/topology-aware-hints/
* https://docs.pivotal.io/tkgi/1-11/release-notes.html

#### Use Case(s)

* Make it possible to deploy Consul in environments, organizations and K8s distributions where `hostPorts` are not an option
* Be closer aligned with K8s best practices

#### Contributions

We anticipate to open a PR soon with a new consul-helm flag/variable to allow this kind of deployment option.

karras:
Changes proposed in this PR:
* The background and reasoning behind this PR is related to #997 (i.e. to avoid `--service-node-port-range` customization)
* Furthermore this PR is also related to various old PRs and issues to make the Consul client ports customizable. We are aware that changing the ports (especially for HTTPS and gRPC) will brick the inject workflow due to hardcoded ports in consul-k8s. 
  - #126
  - #268 
  - #597 
  - #803
  - #804
* For the consul-k8s counterpart see https://github.com/hashicorp/consul-k8s/issues/544
* Fixes #997

How I've tested this PR:
- Wrote unit tests and executed them
- Deployed on Rancher (K8s 1.18, 1.19 and 1.21.1) with adjusted ports (32500, 32501, 32502) + Mesh GWs, then check if Mesh GWs have been successfully registered. App inject depends on separate consul-k8s PR.

- Run standard deployment using the default `values.yml`
- Adjust one of the Consul client ports, do a full deployment, check if the Mesh GWs have registered successfully

Checklist:
- [x] Bats tests added
- [n/a] CHANGELOG entry added (*HashiCorp engineers only, community PRs should not add a changelog entry*)

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -43,7 +43,7 @@ spec:
         {{- if (and .Values.global.metrics.enabled .Values.global.metrics.enableAgentMetrics) }}
         "prometheus.io/scrape": "true"
         "prometheus.io/path": "/v1/agent/metrics"
-        "prometheus.io/port": "8500"
+        "prometheus.io/port": "{{ .Values.client.ports.http.port }}"
         {{- end }}
     spec:
     {{- if .Values.client.affinity }}
@@ -177,7 +177,7 @@ spec:
             {{- end }}
             {{- if .Values.global.tls.enabled }}
             - name: CONSUL_HTTP_ADDR
-              value: https://localhost:8501
+              value: https://localhost:{{ .Values.client.ports.https.port }}
             {{- if .Values.global.tls.enableAutoEncrypt }}
             - name: CONSUL_HTTP_SSL_VERIFY
               value: "false"
@@ -185,6 +185,9 @@ spec:
             - name: CONSUL_CACERT
               value: /consul/tls/ca/tls.crt
             {{- end }}
+            {{- else }}
+            - name: CONSUL_HTTP_ADDR
+              value: http://localhost:{{ .Values.client.ports.http.port }}
             {{- end }}
             {{- include "consul.extraEnvironmentVars" .Values.client | nindent 12 }}
           command:
@@ -218,13 +221,16 @@ spec:
                 -hcl='verify_server_hostname = true' \
                 {{- end }}
                 {{- end }}
-                -hcl='ports { https = 8501 }' \
+                -hcl='ports { https = {{ .Values.client.ports.https.port }} }' \
                 {{- if .Values.global.tls.httpsOnly }}
                 -hcl='ports { http = -1 }' \
                 {{- end }}
                 {{- end }}
+                {{- if (not (and .Values.global.tls.enabled .Values.global.tls.httpsOnly)) }}
+                -hcl='ports { http = {{ .Values.client.ports.http.port }} }' \
+                {{- end }}
                 {{- if .Values.client.grpc }}
-                -hcl='ports { grpc = 8502 }' \
+                -hcl='ports { grpc = {{ .Values.client.ports.grpc.port }} }' \
                 {{- end }}
                 {{- if (and .Values.global.metrics.enabled .Values.global.metrics.enableAgentMetrics) }}
                 -hcl='telemetry { prometheus_retention_time = "{{ .Values.global.metrics.agentMetricsRetentionTime }}" }' \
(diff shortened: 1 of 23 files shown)


templates/client-daemonset.yaml
```diff
@@ -217,13 +217,13 @@ spec:
                 -hcl='verify_server_hostname = true' \
                 {{- end }}
                 {{- end }}
-                -hcl='ports { https = 8501 }' \
+                -hcl='ports { https = {{ .Values.client.ports.https.port }} }' \
                 {{- if .Values.global.tls.httpsOnly }}
                 -hcl='ports { http = -1 }' \
```
karras, line 222: Needs to be extended by an else condition and add hcl parameter to adjust http port

templates/client-daemonset.yaml
```diff
@@ -177,14 +177,17 @@ spec:
...
             - name: CONSUL_HTTP_ADDR
-              value: https://localhost:8501
+              value: https://localhost:{{ .Values.client.ports.https.port }}
             {{- if .Values.global.tls.enableAutoEncrypt }}
             - name: CONSUL_HTTP_SSL_VERIFY
               value: "false"
             {{- else }}
             - name: CONSUL_CACERT
               value: /consul/tls/ca/tls.crt
             {{- end }}
```
karras, line 187: @pree, I have extended the lines below to also customize the `http` port here in the env vars. I think we overlooked it initially. ;)

karras:
Code and unit tests added, ready for review. Acceptance tests seem to be failing due to the 1.10 license handling changes.

karras:
Rebased on master, all checks are green.

values.yaml
```diff
@@ -736,6 +736,23 @@ client:
   # This also changes the clients' advertised IP to the `hostIP` rather than `podIP`.
   exposeGossipPorts: false
 
+  ports:
+  # Customizes ports for the different client functionalities.
+    serflanTcp:
+      port: 8301
+    serflanUdp:
+      port: 8301
```
ishustava, line 744: You need to pass these ports to the client via the `ports` configuration. We're using `-hcl` flags for that (e.g. http, https and grpc ports).

values.yaml
```diff
@@ -736,6 +736,23 @@ client:
...
+    http:
+      port: 8500
+    https:
+      port: 8501
+    grpc:
+      port: 8502
+    dnsTcp:
+      port: 8600
+    dnsUdp:
+      port: 8600
```
ishustava, line 754: Same as above you need to set them in the clients configuration too.

ishustava (commented):
This looks good! Just a couple of comments.

karras:
@ishustava Thanks for the review and additional inputs, much appreciated! As discussed in the issue with David it can't be merged. Thus I'm closing this and the other related PRs and issues to prevent cluttering and prepare for the upcoming repo consolidation into consul-k8s.

PR #1003, opened 2021-06-23
Add optional Consul client nodePort service if hostPorts are not available

### LINKED ISSUE #997: Allow Consul client daemonset connectivity via nodePort (instead of only hostPort)

karras:
Changes proposed in this PR:
- The background and reasoning behind this PR is related to #997 (i.e. to make it possible to deploy Consul in K8s environments where `hostPort` is not available or simply not permitted) and https://github.com/hashicorp/consul-k8s/issues/544
- Fixes #997

How I've tested this PR:
- Wrote unit tests and executed them
- Deployed on Rancher with Kubernetes 1.18, 1.19 and 1.21 to test the topologyKeys and topology hints annotations
- Deployed in combination with the other PRs referenced in the issue

- Ideally this is tested in combination with the other PRs referenced in the issue to allow the client port customization
- If the ports can't be customized the `--service-node-port-range` K8s option must be extended to include the 85XY ports

Checklist:
- [x] Bats tests added
- [n/a] CHANGELOG entry added (*HashiCorp engineers only, community PRs should not add a changelog entry*)

### DIFF
--- a/templates/client-podsecuritypolicy.yaml
+++ b/templates/client-podsecuritypolicy.yaml
@@ -31,6 +31,7 @@ spec:
   {{- else }}
   hostNetwork: false
   {{- end }}
+  {{- if (not .Values.client.service.enabled) }}
   hostPorts:
   {{- if (not (and .Values.global.tls.enabled .Values.global.tls.httpsOnly)) }}
   # HTTP Port
@@ -50,6 +51,7 @@ spec:
   {{- if .Values.client.exposeGossipPorts }}
   - min: 8301
     max: 8301
+  {{- end }}
   {{- end }}
   hostIPC: false
   hostPID: false
--- a/templates/client-securitycontextconstraints.yaml
+++ b/templates/client-securitycontextconstraints.yaml
@@ -19,7 +19,11 @@ allowHostDirVolumePlugin: false
 allowHostIPC: false
 allowHostNetwork: {{ .Values.client.hostNetwork }}
 allowHostPID: false
+{{- if (not .Values.client.service.enabled) }}
 allowHostPorts: true
+{{- else }}
+allowHostPorts: false
+{{- end }}
 allowPrivilegeEscalation: true
 allowPrivilegedContainer: false
 allowedCapabilities: null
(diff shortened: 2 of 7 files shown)


karras:
Code and unit tests added, ready for review. Acceptance tests seem to be failing due to the 1.10 license handling changes.

karras:
Rebased on master, all green now.

ishustava (commented):
This looks good!

karras:
@ishustava Thanks for the review and additional inputs, much appreciated! As discussed in the issue with David it can't be merged. Thus I'm closing this and the other related PRs and issues to prevent cluttering and prepare for the upcoming repo consolidation into consul-k8s.

Issue #1011, opened 2021-06-28
GKE Ingress requires pathType to be ImplementationSpecific


tvories:
### Overview of the Issue

In a GKE cluster, when attempting to use GKE ingress, we are receiving an error:

`4m          Warning   Translate          ingress/consul-consul-ingress                     Translation failed: invalid ingress spec: only "ImplementationSpecific" path type is supported`

[According to Google](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress), for ingress to work properly, `pathType` needs to be set to `ImplementationSpecific`.  In the `ui-ingress.yaml` template, the value is hardcoded to `prefix`.

### Reproduction Steps

1. When running helm install with the following `values.yml`:
```
ui:
      ingress:
        annotations: | # https://cloud.google.com/kubernetes-engine/docs/how-to/load-balance-ingress#ingress_annotations
          kubernetes.io/ingress.class: gce
          kubernetes.io/ingress.global-static-ip-name: ${google_compute_global_address.consul.name}
          networking.gke.io/v1beta1.FrontendConfig: ${helm_release.consul_gcp_ingress_configs.name}
        enabled: true
        hosts:
          - host: ${trimsuffix(google_dns_record_set.consul.name, ".")}
            paths:
              - /* # Google wants /* instead of /
        tls:
          - hosts:
            - consul.example.com
            secretName: ${kubernetes_secret.consul_pki.metadata.0.name}

```
1. View error

`4m          Warning   Translate          ingress/consul-consul-ingress                     Translation failed: invalid ingress spec: only "ImplementationSpecific" path type is supported`

### Expected behavior

Values.yaml should allow someone to override default `pathType` to whichever type they need.

### Environment details

If not already included, please provide the following:
- `consul-k8s` version: 1.19.10-gke.1600
- `consul-helm` version: 0.31.1

Additionally, please provide details regarding the Kubernetes Infrastructure, if known:
- Cloud Provider (If self-hosted, the Kubernetes provider utilized): GKE

PR #1012, opened 2021-06-28
Some providers do not support ingress pathType: Prefix

TJM:
Changes proposed in this PR:
- Make PathType a parameter so it can be overridden (for example on the latest GKE 1.19.10)

How I've tested this PR:
`helm template` with a test.yaml that enables ingress and has a test host
Also tried the "bats" test? (`make`)

see above?

If you want to test overriding the value, it would be like

`helm install consul . -f test.yaml`

test.yaml:
```yaml
ui:
  ingress:
    enabled: true
    pathType: ImplementationSpecific
    hosts:
      - host: foo.bar
        paths:
          - /
```

Checklist:
- [x] Bats tests added

Fixes: #1011

### DIFF
--- a/templates/ui-ingress.yaml
+++ b/templates/ui-ingress.yaml
@@ -40,7 +40,7 @@ spec:
             port:
               number: 80
         path: {{ . }}
-        pathType: Prefix
+        pathType: {{ $.Values.ui.ingress.pathType }}
       {{- end }}
       {{- if $global.tls.enabled }}
       - backend:
@@ -49,7 +49,7 @@ spec:
             port:
               number: 443
         path: {{ . }}
-        pathType: Prefix
+        pathType: {{ $.Values.ui.ingress.pathType }}
       {{- end }}
       {{- end }}
   {{- end }}
--- a/test/unit/ui-ingress.bats
+++ b/test/unit/ui-ingress.bats
@@ -202,3 +202,29 @@ load _helpers
       yq -r '.metadata.annotations.foo' | tee /dev/stderr)
   [ "${actual}" = "bar" ]
 }
+
+#--------------------------------------------------------------------
+# pathtype
+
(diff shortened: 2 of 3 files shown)


TJM:
```console
circleci@c587be94a60b:/test$ bats test/unit/ui-ingress.bats
 ✓ ui/Ingress: disabled by default
 ✓ ui/Ingress: enable with ui.ingress.enabled
 ✓ ui/Ingress: disable with ui.ingress.enabled
 ✓ ui/Ingress: disable with ui.ingress.enabled dash string
 ✓ ui/Ingress: no hosts by default
 ✓ ui/Ingress: hosts can be set
 ✓ ui/Ingress: exposes single port 80 when global.tls.enabled=false
 ✓ ui/Ingress: exposes single port 443 when global.tls.enabled=true and global.tls.httpsOnly=true
 ✓ ui/Ingress: exposes the port 80 when global.tls.enabled=true and global.tls.httpsOnly=false
 ✓ ui/Ingress: exposes the port 443 when global.tls.enabled=true and global.tls.httpsOnly=false
 ✓ ui/Ingress: no tls by default
 ✓ ui/Ingress: tls can be set
 ✓ ui/Ingress: tls with secret name can be set
 ✓ ui/Ingress: no annotations by default
 ✓ ui/Ingress: annotations can be set
 ✓ ui/Ingress: default PathType Prefix
 ✓ ui/Ingress: set PathType ImplementationSpecific

17 tests, 0 failures
```

PS: Thanks for the Docker image :)

TJM:
I have to assume that the acceptance tests that failed are because *I* ran them, and don't have access to a protected enterprise license variable?

thisisnotashwin (approved):
Thanks so much for the contribution @TJM

thisisnotashwin:
Closes #1011

PR #1008, opened 2021-06-25
Disable private link policies on the subnet for the openshift terraform

ishustava:
As described in the tutorial to create openshift clusters on Azure:

https://docs.microsoft.com/en-us/azure/openshift/tutorial-create-cluster

How I've tested this PR:
we need to let nightlies run a few times because the deployment failure was intermittent 

How I expect reviewers to test this PR:
code review

Checklist:
- [ ] Bats tests added
- [ ] CHANGELOG entry added (*HashiCorp engineers only, community PRs should not add a changelog entry*)

PR #1014, opened 2021-06-29
Disable openshift tests and revert #1008

ishustava:
Changes proposed in this PR:
- revert #1008 (turns out the original value was correct even though confusing)
- disable openshift nightly tests until we can make them less flakey

How I've tested this PR:
:trollface: Don't need to test if you're disabling them!

code review

Checklist:
- [ ] Bats tests added


kschoche (approved):
👏🏽

ndhanushkodi (approved):

PR #1017, opened 2021-06-30
Update Helm docs generation

lkysow:
Add headings to each top-level key

So it looks like https://github.com/hashicorp/consul/pull/10498/files


thisisnotashwin (approved):
NICE! Thanks for doing this!!

ndhanushkodi (approved):
Looks great!

PR #1034, opened 2021-07-12
add hostIP as env variable to server statefulset and client daemonset

### LINKED ISSUE #657: Add support for environment variable HOST_IP

kschoche:
Changes proposed in this PR:
- Add hostIP as env variable to server statefulset
- Template and process the extra-env-args and pass them to consul as a new config file instead of using a configmap for the client daemonset and server statefulset.
- Remove existing config-map tests specific to extraConfig and move them to the respective server-statefulset/client-daemonset bats tests.

Resolves https://github.com/hashicorp/consul-helm/issues/657.

How I've tested this PR:
unit tests added and existing ones pass, manually tested using the extra-env-args telemetry specified in issue #657 : 
```
global:
  imageK8S: kschoche/consul-k8s-dev
  image: hashicorp/consul:1.10.0
  tls:
    enabled: true
  acls:
    manageSystemACLs: true
  metrics:
    enabled: true
controller:
  enabled: true
server:
  extraConfig: '{ "telemetry": [ { "disable_hostname": true, "dogstatsd_addr": "HOST_IP:8125", "dogstatsd_tags": ["env:weweq","service:consul","version:1.10.0"] } ] }'
  replicas: 1
```
And confirmed that the server and clients comes online with the changes and fails to come online with existing code.

unit tests pass, code review.

Checklist:
- [x] Bats tests added

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -31,7 +31,6 @@ the trimSuffix removes the trailing newline that occurs when there are no overri
               [ -n "${HOSTNAME}" ] && sed -Ei "s|HOSTNAME|${HOSTNAME?}|g" /consul/extra-config/extra-from-values.json
 {{- end -}}
 
-
 {{/*
 Create chart name and version as used by the chart label.
 */}}
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -16,17 +16,18 @@ as well as the global.name setting.
 {{- end -}}
 
 {{/*
-Set's the args for custom command to render the Vault configuration
-file with IP addresses to make the out of box experience easier
-for users looking to use this chart with Consul Helm.
+Sets up the extra-from-values config file passed to consul and then uses sed to do any necessary
+substitution for HOST_IP/POD_IP/HOSTNAME. Useful for dogstats telemetry. The output file
+is passed to consul as a -config-file param on command line.
 */}}
 {{- define "consul.extraargs" -}}
               mkdir -p /consul/extra-config
-              echo '{{ tpl .Values.server.extraConfig . | trimAll "\"" }}' > /consul/extra-config/extra-from-values.json
+              {{- $conf := toJson .extraConfig }}
+              echo '{{ regexReplaceAll "\\\\\"" $conf "\"" | trimAll "\"" | trimSuffix "\\n" }}' > /consul/extra-config/extra-from-values.json
               [ -n "${HOST_IP}" ] && sed -Ei "s|HOST_IP|${HOST_IP?}|g" /consul/extra-config/extra-from-values.json
               [ -n "${POD_IP}" ] && sed -Ei "s|POD_IP|${POD_IP?}|g" /consul/extra-config/extra-from-values.json
               [ -n "${HOSTNAME}" ] && sed -Ei "s|HOSTNAME|${HOSTNAME?}|g" /consul/extra-config/extra-from-values.json
-{{- end}}
+{{- end -}}
 
 
 {{/*
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -19,8 +19,10 @@ as well as the global.name setting.
 Sets up the extra-from-values config file passed to consul and then uses sed to do any necessary
 substitution for HOST_IP/POD_IP/HOSTNAME. Useful for dogstats telemetry. The output file
 is passed to consul as a -config-file param on command line.
+The regexReplaceAll removes excess escaped quotes, the trim removes excess quotes,
+the trimSuffix removes the trailing newline that occurs when there are no overrides (default).
 */}}
-{{- define "consul.extraargs" -}}
+{{- define "consul.extraconfig" -}}
               mkdir -p /consul/extra-config
               {{- $conf := toJson .extraConfig }}
               echo '{{ regexReplaceAll "\\\\\"" $conf "\"" | trimAll "\"" | trimSuffix "\\n" }}' > /consul/extra-config/extra-from-values.json
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -16,6 +16,20 @@ as well as the global.name setting.
 {{- end -}}
 
 {{/*
+Set's the args for custom command to render the Vault configuration
+file with IP addresses to make the out of box experience easier
+for users looking to use this chart with Consul Helm.
+*/}}
+{{- define "consul.extraargs" -}}
+              mkdir -p /consul/extra-config
+              echo '{{ tpl .Values.server.extraConfig . | trimAll "\"" }}' > /consul/extra-config/extra-from-values.json
+              [ -n "${HOST_IP}" ] && sed -Ei "s|HOST_IP|${HOST_IP?}|g" /consul/extra-config/extra-from-values.json
+              [ -n "${POD_IP}" ] && sed -Ei "s|POD_IP|${POD_IP?}|g" /consul/extra-config/extra-from-values.json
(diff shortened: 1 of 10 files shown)


ishustava (approved):

thisisnotashwin (approved):

lkysow:
It's fine to merge this before the full fix but we should note that it doesn't fully resolve #657.

templates/_helpers.tpl
```diff
@@ -15,6 +15,23 @@ as well as the global.name setting.
...
+{{- define "consul.extraconfig" -}}
+              mkdir -p /consul/extra-config
+              {{- $conf := toJson .extraConfig }}
+              echo '{{ regexReplaceAll "\\\\\"" $conf "\"" | trimAll "\"" | trimSuffix "\\n" }}' > /consul/extra-config/extra-from-values.json
+              [ -n "${HOST_IP}" ] && sed -Ei "s|HOST_IP|${HOST_IP?}|g" /consul/extra-config/extra-from-values.json
+              [ -n "${POD_IP}" ] && sed -Ei "s|POD_IP|${POD_IP?}|g" /consul/extra-config/extra-from-values.json
+              [ -n "${HOSTNAME}" ] && sed -Ei "s|HOSTNAME|${HOSTNAME?}|g" /consul/extra-config/extra-from-values.json
+{{- end -}}
+
+
```
ishustava, line 34:
```suggestion
```

test/unit/client-daemonset.bats
```diff
@@ -569,9 +569,31 @@ load _helpers
...
       yq -r '.spec.template.metadata.annotations."consul.hashicorp.com/config-checksum"' | tee /dev/stderr)
-  [ "${actual}" = 8496f6bcdec460eac8a5c890e7899f5757111e13e54808af533aaf205ef18bd0 ]
+  [ "${actual}" = e9ab414837a3c551f398da62e80fad73a2f1e3c0d0ca82e1067bb4ff68ba6aa9 ]
 }
 
+@test "client/DaemonSet: adds extra-from-vals by default" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/client-daemonset.yaml  \
+      --set 'global.acls.manageSystemACLs=true' \
```
ishustava, line 579: prob don't need acls enabled
```suggestion
```

test/unit/client-daemonset.bats
```diff
@@ -569,9 +569,31 @@ load _helpers
...
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[] | select(.name=="consul") | .command | join(" ") | contains("{}")' | tee /dev/stderr)
+  [ "${actual}" = "true" ]
+}
+
+@test "client/DaemonSet: extraConfig values get set correctly" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/client-daemonset.yaml  \
+      --set 'global.acls.manageSystemACLs=true' \
```
ishustava, line 589:
```suggestion
```

test/unit/client-daemonset.bats
```diff
@@ -569,9 +569,31 @@ load _helpers
...
+  [ "${actual}" = e9ab414837a3c551f398da62e80fad73a2f1e3c0d0ca82e1067bb4ff68ba6aa9 ]
 }
 
+@test "client/DaemonSet: adds extra-from-vals by default" {
+  cd `chart_dir`
+  local actual=$(helm template \
+      -s templates/client-daemonset.yaml  \
+      --set 'global.acls.manageSystemACLs=true' \
+      . | tee /dev/stderr |
+      yq '.spec.template.spec.containers[] | select(.name=="consul") | .command | join(" ") | contains("{}")' | tee /dev/stderr)
```
ishustava, line 581: I think we also need to test that the `-config-file` with the right file is passed in

templates/_helpers.tpl
```diff
@@ -15,6 +15,23 @@ as well as the global.name setting.
...
+the trimSuffix removes the trailing newline that occurs when there are no overrides (default).
+*/}}
+{{- define "consul.extraconfig" -}}
+              mkdir -p /consul/extra-config
+              {{- $conf := toJson .extraConfig }}
+              echo '{{ regexReplaceAll "\\\\\"" $conf "\"" | trimAll "\"" | trimSuffix "\\n" }}' > /consul/extra-config/extra-from-values.json
+              [ -n "${HOST_IP}" ] && sed -Ei "s|HOST_IP|${HOST_IP?}|g" /consul/extra-config/extra-from-values.json
+              [ -n "${POD_IP}" ] && sed -Ei "s|POD_IP|${POD_IP?}|g" /consul/extra-config/extra-from-values.json
+              [ -n "${HOSTNAME}" ] && sed -Ei "s|HOSTNAME|${HOSTNAME?}|g" /consul/extra-config/extra-from-values.json
+{{- end -}}
```
ishustava, line 32: I think we can test it in `helpers.bats`.

values.yaml
```diff
@@ -484,7 +484,7 @@ server:
   # This can also be set using Helm's `--set` flag using the following syntax:
   #
   # ```shell
-  # --set 'server.extraConfig="{"log_level": "DEBUG"}"'
+  # --set 'server.extraConfig='{"log_level": "DEBUG"}"'
```
ishustava, line 487: This single quote would be an extra one. I think this example needs double quotes inside the json string escaped.

ishustava (commented):
Looks good, Kyle. I left a couple comments about tests, but otherwise it looked good. Will wait for those to be resolved before approving.

values.yaml
kschoche: good catch, thx!

test/unit/server-statefulset.bats
```diff
@@ -2,6 +2,7 @@
 
 load _helpers
 
+
```
thisisnotashwin, line 5:
```suggestion
```

values.yaml
```diff
@@ -794,7 +794,7 @@ client:
   # This can also be set using Helm's `--set` flag using the following syntax:
   #
   # ```shell
-  # --set 'client.extraConfig="{"log_level": "DEBUG"}"'
+  # --set 'client.extraConfig='{"log_level": "DEBUG"}"'
```
thisisnotashwin, line 797: I think this needs to be reverted. I think it was correct previously.

values.yaml
thisisnotashwin: This does not seem resolved 🤔

thisisnotashwin (approved):
This looks good!! I agree with @ishustava 's suggestion regarding adding a test to ensure that the correct file is passed to `-config-file`

values.yaml
kschoche: I dont think the original is valid, the quotes need to be escaped.

PR #1042, opened 2021-07-16
add hostIP as env to server-statefulset and substitute env variables in extraConfig

### LINKED ISSUE #657: Add support for environment variable HOST_IP

kschoche:
Changes proposed in this PR:
- Add hostIP as env variable to server statefulset
- Copies the `extraConfig` configMap file to `/consul/extra-config/extra-from-values.json` and uses `sed` to replace `HOST_IP/POD_IP/HOSTNAME` at runtime with the environment.
- Passes `-config-file=/consul/extra-config/extra-from-values.json` to server-statefulset and client-daemonset.

Resolves #657.

How I've tested this PR:
unit tests updated and pass, manually tested using the extra-env-args telemetry specified in issue #657 :

```
global:
  imageK8S: kschoche/consul-k8s-dev
  image: hashicorp/consul:1.10.0
  tls:
    enabled: true
  acls:
    manageSystemACLs: true
  metrics:
    enabled: true
controller:
  enabled: true
server:
  extraConfig: '{ "telemetry": [ { "disable_hostname": true, "dogstatsd_addr": "HOST_IP:8125", "dogstatsd_tags": ["env:weweq","service:consul","version:1.10.0"] } ] }'
  replicas: 1
```
And confirmed that the server and clients comes online with the changes and fails to come online with existing code.

unit tests pass, code review.

Note to reviewers:  
1. Given this is not configurable and just becomes hard-coded text in `server-statefulset` and `client-daemonset` is there anything worth additionally testing in bats?
2. In the current design consul will parse the config-file twice, once because it's in the configMap file `/consul/config/extra-from-values.json` and once again in the new one. Based on https://www.consul.io/docs/agent/options#_config_file merging should be "ok". I'm not sure this is guaranteed. If we are concerned about that, I can rename the configMap entry to `extra-from-values.tmp` and Consul will ignore it when processing `/consul/config/` and only catch the one we modify. Thoughts?

Checklist:
- [ ] Bats tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 IMPROVEMENTS:
+* Substitute `HOST_IP/POD_IP/HOSTNAME` variables in `server.extraConfig` and `client.extraConfig` at runtime so they are passed in to server/client config already evaluated. [[GH-1042](https://github.com/hashicorp/consul-helm/pull/1042)]
 * Set failurePolicy to Fail for connectInject mutating webhook so that pods fail to schedule when the webhook is offline. This can be controlled via `connectInject.failurePolicy`. [[GH-1024](https://github.com/hashicorp/consul-helm/pull/1024)]
 * Allow setting global.logLevel and global.logJSON and propogate this to all consul-k8s commands. [[GH-980](https://github.com/hashicorp/consul-helm/pull/980)]
 * Allow setting `connectInject.replicas` to control number of replicas of webhook injector. [[GH-1029](https://github.com/hashicorp/consul-helm/pull/1029)]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Unreleased
 
 IMPROVEMENTS:
-* Substitute `HOST_IP/POD_IP/HOSTNAME` variables in `server.extraConfig` and `client.extraConfig` at runtime so they are passed in to server/client config already evaluated. [[GH-1042](https://github.com/hashicorp/consul-helm/pull/1042)]
+* Substitute `HOST_IP/POD_IP/HOSTNAME` variables in `server.extraConfig` and `client.extraConfig` so they are passed in to server/client config already evaluated at runtime. [[GH-1042](https://github.com/hashicorp/consul-helm/pull/1042)]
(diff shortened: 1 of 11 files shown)


thisisnotashwin (approved):
I looked at the comment regarding the testing. I think i agree with the complexity of it. I havent tested this PR, but if you have and it does solve the problem, even when the input json is complicated, this LGMT!!

Thanks for putting in the hard work on this one. It definitely was quite a wild journey for a PR!!

ishustava (approved):
Looks good!

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Unreleased
 
 IMPROVEMENTS:
+* Substitute `HOST_IP/POD_IP/HOSTNAME` variables in `server.extraConfig` and `client.extraConfig` at runtime so they are passed in to server/client config already evaluated. [[GH-1042](https://github.com/hashicorp/consul-helm/pull/1042)]
```
thisisnotashwin, line 4:
```suggestion
* Substitute `HOST_IP/POD_IP/HOSTNAME` variables in `server.extraConfig` and `client.extraConfig` so they are passed in to server/client config already evaluated at runtime. [[GH-1042](https://github.com/hashicorp/consul-helm/pull/1042)]
```

PR #1046, opened 2021-07-20
Add ability to manual specify a server-cert

barrymars:
Changes proposed in this PR:
-
- Adds the ability to manually specify a secret containing a pre-generated server-cert.
-
-This coupled with the autoEncrypt feature, removes the need for the tls-init job and so can be run in a gitops way without having to rely on helm hooks etc.

How I've tested this PR:
Used to deploy consul into both local 'kind' cluster and AWS EKS cluster.

Checklist:
- [x] Bats tests added

### DIFF
--- a/templates/tls-init-cleanup-job.yaml
+++ b/templates/tls-init-cleanup-job.yaml
@@ -1,5 +1,5 @@
 {{- if (or (and (ne (.Values.server.enabled | toString) "-") .Values.server.enabled) (and (eq (.Values.server.enabled | toString) "-") .Values.global.enabled)) }}
-{{- if .Values.global.tls.enabled }}
+{{- if (and .Values.global.tls.enabled (not .Values.server.serverCert.secretName)) }}
 # tls-init-cleanup job deletes Kubernetes secrets created by tls-init
 apiVersion: batch/v1
 kind: Job
--- a/values.yaml
+++ b/values.yaml
@@ -357,7 +357,7 @@ server:
   #     -additional-dnsname="*.{{fullname}}-server.{{namespace}}" \
   #     -additional-dnsname="*.{{fullname}}-server.{{namespace}}.svc" \
   #     -additional-dnsname="*.server.{{datacenter}}.{{domain}}" \
-  #     -additional-dnsname="server.local.consul"
+  #     -additional-dnsname="server.{{datacenter}}.{{domain}}"
   # ```
   #
   # If you have generated the
--- a/values.yaml
+++ b/values.yaml
@@ -345,6 +345,36 @@ server:
     # Manages license autoload. Required in Consul 1.10.0+, 1.9.7+ and 1.8.12+.
     enableLicenseAutoload: true
 
+  # A Kubernetes secret containing a certificate & key for the server agents to use
+  # for TLS communication within the Consul cluster. Cert needs to be carefully
+  # formed to work as if tls-init created it (reproducing what consul-k8s runs):
+  #
+  # ```bash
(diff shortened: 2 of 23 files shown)


david-yu:
Hi @barrymars out of curiosity are you potentially using Vault for minting your server certs? How do you envision rotating them? We are looking into further integrations with Vault to support retrieving server certs from Vault for TLS.

barrymars:
we're actually moving away from vault in favour of AWS secrets manager.

hadn't got as far as working out certificate rotation, still at the proof of concept stage

ishustava (changes requested):
Looking good, thanks for making this change! 

With server cert provided, I think we also need to disable tls-init and tls-init-cleanup jobs.

barrymars:
good call @ishustava 

I am using the helm chart via Tanka (which doesn't support hooks), so hadn't thought of that

I've also added a sanity check so that if setting serverCert you have to have caCert set too

values.yaml
```diff
@@ -345,6 +345,36 @@ server:
...
+  # ```bash
+  # consul tls cert create -server -days=730 -domain=consul -ca=consul-agent-ca.pem \
+  #     -key=consul-agent-ca-key.pem -dc={{datacenter}} \
+  #     -additional-dnsname="{{fullname}}-server" \
+  #     -additional-dnsname="*.{{fullname}}-server" \
+  #     -additional-dnsname="*.{{fullname}}-server.{{namespace}}" \
+  #     -additional-dnsname="*.{{fullname}}-server.{{namespace}}.svc" \
+  #     -additional-dnsname="*.server.{{datacenter}}.{{domain}}" \
+  #     -additional-dnsname="localhost" \
+  #     -additional-dnsname="127.0.0.1" \
```
ishustava, line 361: These are added by default by the consul CLI
```suggestion
```

values.yaml
```diff
@@ -345,6 +345,36 @@ server:
...
+  #     -additional-dnsname="server.local.consul"
```
ishustava, line 362: Curious, what is this DNS name for?

templates/tls-init-cleanup-job.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if (or (and (ne (.Values.server.enabled | toString) "-") .Values.server.enabled) (and (eq (.Values.server.enabled | toString) "-") .Values.global.enabled)) }}
-{{- if .Values.global.tls.enabled }}
+{{- if (and .Values.global.tls.enabled (not .Values.server.serverCert.secretName)) }}
```
ishustava, line 2: Would you mind adding bats tests for these as well? 🙏

values.yaml
```diff
@@ -345,6 +345,36 @@ server:
     # Manages license autoload. Required in Consul 1.10.0+, 1.9.7+ and 1.8.12+.
     enableLicenseAutoload: true
 
+  # A Kubernetes secret containing a certificate & key for the server agents to use
+  # for TLS communication within the Consul cluster. Cert needs to be carefully
+  # formed to work as if tls-init created it (reproducing what consul-k8s runs):
```
ishustava, line 350: nit: Rephrasing this a bit
```suggestion
  # for TLS communication within the Consul cluster. Cert needs to be provided with
  # additional DNS name SANs so that it will work within the Kubernetes cluster:
```

values.yaml
barrymars: pretty sure that should be server.{{datacenter}}.{{domain}}

ishustava (approved):
Looks good! Thanks so much for this contribution!

ndhanushkodi (approved):
Looks good to me and @sadjamz!

PR #1055, opened 2021-07-26
Use docker-entrypoint when running consul 

lkysow:
Previously we were executing consul directly via /bin/consul and we were
using the Kubernetes `command` key which overrides the default Docker
ENTRYPOINT.

This had no issues, however I propose that it's best to execute via the
entrypoint so we keep things consistent with how the Docker image was
supposed to be run. This is also consistent with Vault's Helm chart. If
you step through the entrypoint script, nothing actually changes with
how we execute consul because we don't trigger any of the `if`
statements. So there is no real effect right now.

In order to support using the entrypoint script, we need to disable part of the
script that attempts to change the ownership of the /consul directory
because the ownership is already set via Kube.

Entrypoint script is here: https://github.com/hashicorp/docker-consul/blob/master/0.X/docker-entrypoint.sh

How I've tested this PR:
* acceptance tests

* code

Checklist:

### DIFF
--- a/templates/client-daemonset.yaml
+++ b/templates/client-daemonset.yaml
@@ -164,6 +164,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: status.podIP
+            - name: CONSUL_DISABLE_PERM_MGMT
+              value: "true"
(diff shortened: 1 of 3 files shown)


lkysow:
Closing for https://github.com/hashicorp/consul-k8s/pull/590

PR #1056, opened 2021-07-27
acceptance-tests: connect tests verify that the ACL token was deleted

ishustava:
Changes proposed in this PR:
- Add an assertion that ACL token gets deleted when service is deleted to the connect acceptance tests.

How I've tested this PR:
Don't need to test the tests 🐢 🐢 🐢 

code review

Checklist:
- [ ] Bats tests added

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,7 +55,7 @@ commands:
         type: string
       consul-k8s-image:
         type: string
-        default: "hashicorpdev/consul-k8s:2dfffed"
+        default: "docker.mirror.hashicorp.services/hashicorpdev/consul-k8s:latest"
     steps:
       - when:
           condition: << parameters.failfast >>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,7 +55,7 @@ commands:
         type: string
       consul-k8s-image:
         type: string
-        default: "docker.mirror.hashicorp.services/hashicorpdev/consul-k8s:latest"
+        default: "docker.mirror.hashicorp.services/hashicorpdev/consul-k8s:28d7141"
     steps:
       - when:
           condition: << parameters.failfast >>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,7 +55,7 @@ commands:
         type: string
       consul-k8s-image:
         type: string
-        default: "docker.mirror.hashicorp.services/hashicorpdev/consul-k8s:28d7141"
+        default: "docker.mirror.hashicorp.services/hashicorpdev/consul-k8s:latest"
     steps:
       - when:
           condition: << parameters.failfast >>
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -55,7 +55,7 @@ commands:
         type: string
       consul-k8s-image:
         type: string
-        default: "docker.mirror.hashicorp.services/hashicorpdev/consul-k8s:latest"
+        default: "hashicorpdev/consul-k8s:5b0b946"
```
ishustava, line 58: reminder: remove once hashicorp/consul-k8s#571 is merged

test/acceptance/tests/connect/connect_inject_test.go
```diff
@@ -50,6 +50,25 @@ func TestConnectInject(t *testing.T) {
...
+				// because golang will execute them in reverse order i.e. the last registered
+				// cleanup function will be executed first.
+				t.Cleanup(func() {
+					retry.Run(t, func(r *retry.R) {
+						tokens, _, err := consulClient.ACL().TokenList(nil)
+						require.NoError(r, err)
+						for _, token := range tokens {
+							require.NotContains(r, token.Description, staticServerName)
+							require.NotContains(r, token.Description, staticClientName)
+						}
```
kschoche, line 67: Probably a super nit comment but I'm wondering if this should exit if `TokenList()` returns error (during shutdown or otherwise), or it may end up stuck in a retry loop?

test/acceptance/tests/connect/connect_inject_test.go
```diff
@@ -76,10 +95,8 @@ func TestConnectInject(t *testing.T) {
 					k8s.CheckStaticServerConnectionFailing(t, ctx.KubectlOptions(t), "http://localhost:1234")
 				}
 
-				consulClient := consulCluster.SetupConsulClient(t, true)
-
 				logger.Log(t, "creating intention")
-				_, _, err := consulClient.Connect().IntentionCreate(&api.Intention{
+				_, err := consulClient.Connect().IntentionUpsert(&api.Intention{
```
kschoche, line 99: Kinda curious about this one, for my own knowledge why are we doing upsert instead of create?

kschoche (approved):
Looks great! I like the cleanup as well!
I left a couple questions that are non-blocking!

test/acceptance/tests/connect/connect_inject_test.go
ishustava: It won't get stuck in retry loop since the retry function times out after 7 seconds by default. I think it's ok in this case since it won't add that much time to the test running time.

test/acceptance/tests/connect/connect_inject_test.go
ishustava: `IntentionCreate` is [deprecated](https://github.com/hashicorp/consul/blob/21e855d860865f1472b3fdcb64a916e92bc83a9f/api/connect_intention.go#L376). Sorry I should have commented on the PR.

sadjamz (approved):
These are some well written tests!

PR #1066, opened 2021-08-04
Pin to aws provider 3.51.0

lkysow:
Version 3.51.2 requires Terraform 0.14.6 and we're running 0.13.5. 3.51.2 was released a couple days ago which is why our tests started failing.

I didn't upgrade Terraform because I didn't want to make any major changes unless it was necessary.


lkysow:
Hold off on this, looks like the error was from the merge-repos branch.

PR #1067, opened 2021-08-04
Fix all outstanding acceptance test errors from golangci-lint

kschoche:
Changes proposed in this PR:
- Pre-emptive fixes to acceptance tests so that golangci-lint does not fail when monorepo is merged.

How I've tested this PR:
tests should still pass

code review, tests still pass
Note to reviewers:  To test this out locally check out the branch from the PR and run `golangci-lint run --config=/path/to/consul-k8s/.golangci.yaml` from `consul-helm/tests/acceptance` 

Checklist:
- [ ] Bats tests added

### DIFF
--- a/test/acceptance/framework/consul/consul_cluster.go
+++ b/test/acceptance/framework/consul/consul_cluster.go
@@ -41,7 +41,6 @@ type Cluster interface {
 // HelmCluster implements Cluster and uses Helm
 // to create, destroy, and upgrade consul
 type HelmCluster struct {
-	cfg                config.TestConfig
 	ctx                environment.TestContext
 	helmOptions        *helm.Options
 	releaseName        string
@@ -144,7 +143,7 @@ func (h *HelmCluster) Destroy(t *testing.T) {
 
 	// Ignore the error returned by the helm delete here so that we can
 	// always idempotently clean up resources in the cluster.
-	helm.DeleteE(t, h.helmOptions, h.releaseName, false)
+	_ = helm.DeleteE(t, h.helmOptions, h.releaseName, false)
 
 	// Force delete any pods that have h.releaseName in their name because sometimes
 	// graceful termination takes a long time and since this is an uninstall
@@ -162,7 +161,8 @@ func (h *HelmCluster) Destroy(t *testing.T) {
 	}
 
 	// Delete PVCs.
-	h.kubernetesClient.CoreV1().PersistentVolumeClaims(h.helmOptions.KubectlOptions.Namespace).DeleteCollection(context.Background(), metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: "release=" + h.releaseName})
+	err = h.kubernetesClient.CoreV1().PersistentVolumeClaims(h.helmOptions.KubectlOptions.Namespace).DeleteCollection(context.Background(), metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: "release=" + h.releaseName})
+	require.NoError(t, err)
 
 	// Delete any serviceaccounts that have h.releaseName in their name.
 	sas, err := h.kubernetesClient.CoreV1().ServiceAccounts(h.helmOptions.KubectlOptions.Namespace).List(context.Background(), metav1.ListOptions{LabelSelector: "release=" + h.releaseName})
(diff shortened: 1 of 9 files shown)


test/acceptance/framework/consul/consul_cluster.go
```diff
@@ -162,7 +161,8 @@ func (h *HelmCluster) Destroy(t *testing.T) {
 	}
 
 	// Delete PVCs.
-	h.kubernetesClient.CoreV1().PersistentVolumeClaims(h.helmOptions.KubectlOptions.Namespace).DeleteCollection(context.Background(), metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: "release=" + h.releaseName})
+	err = h.kubernetesClient.CoreV1().PersistentVolumeClaims(h.helmOptions.KubectlOptions.Namespace).DeleteCollection(context.Background(), metav1.DeleteOptions{}, metav1.ListOptions{LabelSelector: "release=" + h.releaseName})
+	require.NoError(t, err)
```
kschoche, line 165: Not totally sure on this one.

test/acceptance/framework/helpers/helpers.go
```diff
@@ -57,7 +57,7 @@ func WaitForAllPodsToBeReady(t *testing.T, client kubernetes.Interface, namespac
 // Sets up a goroutine that will wait for interrupt signals
 // and call cleanup function when it catches it.
 func SetupInterruptHandler(cleanup func()) {
-	c := make(chan os.Signal)
+	c := make(chan os.Signal, 1)
```
lkysow, line 60: why are we changing this to a buffered channel?

test/acceptance/framework/k8s/debug.go
```diff
@@ -78,20 +78,29 @@ func WritePodsDebugInfoIfFailed(t *testing.T, kubectlOptions *k8s.KubectlOptions
 
 		// Describe any stateful sets.
 		statefulSets, err := client.AppsV1().StatefulSets(kubectlOptions.Namespace).List(context.Background(), metav1.ListOptions{LabelSelector: labelSelector})
+		if err != nil {
+			logger.Log(t, "unable to get statefulsets", "err", err)
+		}
 		for _, statefulSet := range statefulSets.Items {
 			// Describe stateful set and write it to a file.
 			writeResourceInfoToFile(t, statefulSet.Name, "statefulset", testDebugDirectory, kubectlOptions)
 		}
```
lkysow, line 87:
```suggestion
		if err != nil {
			logger.Log(t, "unable to get statefulsets", "err", err)
		}
		else {
			for _, statefulSet := range statefulSets.Items {
				// Describe stateful set and write it to a file.
				writeResourceInfoToFile(t, statefulSet.Name, "statefulset", testDebugDirectory, kubectlOptions)
			}
		}
```

I think we need an `else` for all of these cuz the resource could be `nil` and then `.Items` will panic.

test/acceptance/framework/consul/consul_cluster.go
lkysow: Looks correct to me

test/acceptance/framework/helpers/helpers.go
kschoche: This was flagged by the linter.
"The os/signal package will not block when sending a signal, which means that the channel passed to signal.Notify should always be buffered." - https://github.com/golang/lint/issues/175

lkysow: 👍 thanks!

lkysow (approved):

test/acceptance/framework/consul/consul_cluster.go
```diff
@@ -41,7 +41,6 @@ type Cluster interface {
 // HelmCluster implements Cluster and uses Helm
 // to create, destroy, and upgrade consul
 type HelmCluster struct {
-	cfg                config.TestConfig
```
sadjamz, line 44: Was this attribute deleted just because it's unused?

test/acceptance/tests/connect/connect_inject_namespaces_test.go
```diff
@@ -198,7 +198,7 @@ func TestConnectInjectNamespaces(t *testing.T) {
 				}
 
 				logger.Log(t, "creating intention")
-				_, _, err := consulClient.Connect().IntentionCreate(intention, nil)
+				_, err := consulClient.Connect().IntentionUpsert(intention, nil)
```
sadjamz, line 201: I'm just curious, why did this end up changing?

sadjamz (approved):
Great job Kyle! I really like what I see on line 387 for example where outside the if block the variable definition is '_' and since we only use it inside an if case we define it there. This kind of localization is solid.

I also left 2 non-blocking questions for my own curiosity. 

Thanks!

test/acceptance/framework/consul/consul_cluster.go
kschoche: Yup!

test/acceptance/tests/connect/connect_inject_namespaces_test.go
kschoche: `IntentionCreate` has been deprecated by the api, so now it's `IntentionUpsert`!
Kinda cool that the linter checks for this and flags it! :)

test/acceptance/framework/consul/consul_cluster.go
kschoche: If you're further curious you could check out master and run golangci-lint against it to see the original. It was pretty verbose so I did not include it in PR but maybe that'd have been useful.
The github action that we're using in `consul-k8s` is kinda cool in that it flags these in the PR along with explanations.

